PR #1, opened 2021-12-20
src/trigger.py: fix --poll-period

gctucker:
Fix the --poll-period help in trigger.py and add missing import time.

Fixes: 1f49cd92b848 ("src/trigger.py: add script to detect new revisions")


gctucker (commented):
Tested OK locally

PR #2, opened 2021-12-20
src/runner.py: add Runner class and use temp directory

gctucker:
Move all the code to actually run a test to a separate Runner class
and create a separate temporary directory when running each job so the
generated files don't overwrite each other.


gctucker (commented):
Tested OK locally

PR #4, opened 2021-12-28
github: Create main.yml

JenySadadia:
Signed-off-by: Jeny Sadadia <jeny.sadadia@gmail.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,32 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  check:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.10']
+
+    steps:
+
+      - name: Check out source code
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@master
+        with:
+          python-version: ${{ matrix.python-version }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,3 +30,9 @@ jobs:
         uses: actions/setup-python@master
         with:
           python-version: ${{ matrix.python-version }}
+
+      - name: Install Python packages
+        run: |
+          python --version
+          python -m pip install --upgrade pip
+          pip install -r requirements-dev.txt
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,3 +36,7 @@ jobs:
           python --version
           python -m pip install --upgrade pip
           pip install -r requirements-dev.txt
+
+      - name: Run pycodestyle
+        run: |
+          pycodestyle src/*.py
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,1 @@
+pycodestyle==2.8.0


.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
...
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  check:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.8', '3.9', '3.10']
```
gctucker, line 22: Maybe we should reduce to only test the oldest and newest versions supported to reduce the amount of computing needed:
```suggestion
        python-version: ['3.7', '3.10']
```

.github/workflows/main.yml
```diff
@@ -23,15 +23,15 @@ jobs:
 
     steps:
 
-     -  name: Check out source code
```
gctucker: Please squash these changes with the first patch that adds `main.yml`.

.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
+  push:
+    branches: [ main, setup-github-actions ]
```
gctucker, line 9: This shouldn't be necessary now that a PR has been created for the `main` branch:
```suggestion
    branches: [ main ]
```

gctucker (commented):
A few things to tweak, LGTM otherwise.  Please also add commit messages to describe the changes.

.github/workflows/main.yml
JenySadadia: This change contains code from commits after the first patch.
Merge conflicts are there when trying to squash.

gctucker: OK, I mean the indentation fixes could be done in the `github: Create main.yml` commit to make the last commit do nothing, and then it'll disappear.

If it's causing too much git conflicts then no worries, we can merge it like this.

gctucker:
I see the checks aren't run yet, even though this is a PR for the `main` branch.  But I think it should start working once this has been merged.

.github/workflows/main.yml
JenySadadia: Yes, this commit is on top of the two commits:
https://github.com/kernelci/kernelci-pipeline/pull/4/commits/82af750d1cfd794fea7714ab41322f6bdf9169f1
https://github.com/kernelci/kernelci-pipeline/pull/4/commits/49d7a9053878cf993afa985fa38d6ed16b8ecb7f

That's why it's difficult to squash it.
I can force push those two commits with code changes and delete this commit, or we can merge it like this.

JenySadadia:
> 

Yes, it needs to be merged first.

.github/workflows/main.yml
gctucker: Yeah that's what I meant, it should be possible to move the changes from the last commit to the earlier ones.  I had a look locally, here's how I did it:
```
$ git checkout jeny/setup-github-actions -b tmp
$ cp .github/workflows/main.yml _main.yml
$ git rebase -i main --onto=main
# edit all the commits that change the .yml file
# then for each commit:
$ diff -u .github/workflows/main.yml _main.yml
# to see the differences and adjust the file accordingly, you can even copy things from _main.yml. Then:
$ git add config/core/build-configs-staging.yaml
$ git rebase --continue
# eventually:
nothing added to commit but untracked files present
Could not apply 8d1fcdb2315e... github/main.yaml: Fix indentation
$ git rebase --continue
```
And then `git diff jeny/setup-github-actions` shows no difference, so the resulting code is the same but without the last commit to fix the indent.  I've pushed my resulting branch here, feel free to pick commits from it and force-push to this PR or do the same things by hand:
https://github.com/gctucker/kernelci-pipeline/commits/jeny-setup-github-actions

I believe the indentation matters in this YAML file, I haven't tried to parse it but it's better if we can avoid having any commits with issues if we can.

JenySadadia: Updated commits.

JenySadadia: Thank you for the guide @gctucker :)

gctucker (approved):

Issue #3, opened 2021-12-24
Only push Git commit in trigger.py


gctucker:
The current `trigger.py` script provides a proof-of-concept implementation for submitting new revisions to the API.  A few things need to be modified in order to be able to use it in a real-world pipeline:

* Rather than updating a local full Git checkout, use `kernelci.build.get_branch_head()` to get the latest git commit for the build config being monitored (e.g. `mainline`)
* Send a GET query to the API to see if there is any Node object with a revision matching the git commit found.  If there is one, then stop at this point and try again later.
* If there isn't a Node with a revision with the git commit found, submit one with an empty `describe` and empty `status`.  The API will then generate a pub/sub event.

This should be all that the `trigger.py` script should do.  Another client will then be listening for events when a new revision has been added to push a source tarball and update the revision information with the git describe (requires a local checkout) and set the status to `True` if it worked.  Then other clients will be running tests that rely on the source tarball etc.

Depends on https://github.com/kernelci/kernelci-api/issues/66 to only create new Node using `POST`.

JenySadadia:
Hello @gctucker, 
For the below task: 
```
Submit a Node object without describe in the revision to the API and without status value, which will return an error if the revision already exists and ignore it or create a new one and generate a pub/sub event if it's a new one
```
I tried to POST node without 'describe'. kernelci-api is not able to process the request because of the missing property.
```
trigger     | Sending revision node to API: df0cc57e057f18e44dac8e6c18aba47ab53202f9
trigger     | node {'name': 'checkout', 'status': True, 'revision': {'tree': 'mainline', 'url': 'https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git', 'branch': 'master', 'commit': 'df0cc57e057f18e44dac8e6c18aba47ab53202f9'}}
trigger     | 422 Client Error: Unprocessable Entity for url: http://172.17.0.1:8001/node
trigger     | Traceback (most recent call last):
trigger     |   File "/home/kernelci/./pipeline/trigger.py", line 101, in <module>
trigger     |     status = opts.command(configs, opts)
trigger     |   File "/home/kernelci/./pipeline/trigger.py", line 89, in __call__
trigger     |     _run_trigger(args, build_config, db)
trigger     |   File "/home/kernelci/./pipeline/trigger.py", line 56, in _run_trigger
trigger     |     resp_obj = db.submit({'node': node})[0]
[...]

PR #6, opened 2022-01-17
Push git commit with trigger.py

JenySadadia:
Fixes https://github.com/kernelci/kernelci-pipeline/issues/3

src/trigger.py: add get_node_by_commit_id
src/trigger.py: get latest git commit and use it to get node
src/trigger.py: send revision node with empty describe
src/trigger.py: define revision dict and remove unnecessary code

Signed-off-by: Jeny Sadadia <jeny.sadadia@gmail.com>

### DIFF
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -57,12 +57,8 @@ def _run_trigger(args, build_config, db):
     print(f"Sending revision node to API: {revision['commit']}")
     sys.stdout.flush()
     node = {
-        'name': "checkout",
-        'status': True,
-        'revision': {
-            k: revision[k] for k in [
-                'tree', 'url', 'branch', 'commit', 'describe',
-            ]},
+        'name': 'checkout',
+        'revision': revision,
     }
     resp_obj = db.submit({'node': node})[0]
     node_id = resp_obj['_id']
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -31,28 +31,19 @@ def get_node_by_commit_id(commit_id):
 
 
 def _run_trigger(args, build_config, db):
-    if args.skip_pull:
-        print("Not updating repo")
-    else:
-        head_commit = kernelci.build.get_branch_head(build_config)
-        node_list = get_node_by_commit_id(head_commit)
-        if node_list:
-            print(f"Node exists with the latest git commit: {head_commit}")
-            return
-        print(f"Updating repo: {args.kdir}")
-        sys.stdout.flush()
-        kernelci.build.update_repo(build_config, args.kdir)
-
-    print("Gathering revision meta-data")
+    head_commit = kernelci.build.get_branch_head(build_config)
+    node_list = get_node_by_commit_id(head_commit)
+    if node_list:
+        print(f"Node exists with the latest git commit {head_commit}")
+        return
     sys.stdout.flush()
-    step = kernelci.build.RevisionData(args.kdir, args.output, reset=True)
-    res = step.run(opts={
+
+    revision = {
         'tree': build_config.tree.name,
         'url': build_config.tree.url,
         'branch': build_config.branch,
-    })
-    meta = kernelci.build.Metadata(args.output)
-    revision = meta.get('bmeta', 'revision')
+        'commit': head_commit
+    }
 
     print(f"Sending revision node to API: {revision['commit']}")
     sys.stdout.flush()
@@ -72,11 +63,6 @@ class cmd_run(Command):
         Args.kdir, Args.build_config, Args.output, Args.db_config,
     ]
     opt_args = [
-        {
-            'name': '--skip-pull',
-            'action': 'store_true',
-            'help': "Skip git pull",
-        },
         {
             'name': '--poll-period',
             'type': int,
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -34,6 +34,11 @@ def _run_trigger(args, build_config, db):
     if args.skip_pull:
         print("Not updating repo")
     else:
+        head_commit = kernelci.build.get_branch_head(build_config)
+        node_list = get_node_by_commit_id(head_commit)
+        if node_list:
+            print(f"Node exists with the latest git commit: {head_commit}")
+            return
         print(f"Updating repo: {args.kdir}")
         sys.stdout.flush()
         kernelci.build.update_repo(build_config, args.kdir)
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -7,7 +7,6 @@
 
 import json
 import os
-import requests
 import sys
 import time
 
@@ -16,6 +15,19 @@ import kernelci.build
 import kernelci.config
 import kernelci.data
 from kernelci.cli import Args, Command, parse_opts
+import urllib
+import requests
+
+DEFAULT_URL = 'http://172.17.0.1:8001'
+
+
+def get_node_by_commit_id(commit_id):
+    """
+    Get node object by sending commit ID to 'nodes' endpoint
+    """
+    get_node_path = '?'.join(['nodes', 'revision.commit='+commit_id])
+    url = urllib.parse.urljoin(DEFAULT_URL, get_node_path)
+    return requests.get(url).json()
 
 
 def _run_trigger(args, build_config, db):


src/trigger.py
```diff
@@ -38,6 +38,9 @@ def _run_trigger(args, build_config, db):
     else:
         node_list = get_node_by_commit_id(
             kernelci.build.get_branch_head(build_config))
+        if len(node_list) != 0:
```
gctucker, line 41: This is correct, but `if not node_list:` is a bit better for checking whether a collection is empty.  It's something `pylint` would generate a warning about.

src/trigger.py
```diff
@@ -38,6 +38,9 @@ def _run_trigger(args, build_config, db):
...
+            print(f"Node exists with the latest git commit")
+            return 0
```
gctucker, line 43: I don't think there's any expected return value from this method, so maybe just `return` is better.  That would actually return `None`.

src/trigger.py
```diff
@@ -16,6 +16,20 @@
 import kernelci.config
 import kernelci.data
 from kernelci.cli import Args, Command, parse_opts
+import urllib
+
+DEFAULT_URL = 'http://172.17.0.1:8001'
+
+
+def get_node_by_commit_id(commit_id):
```
gctucker, line 24: I think we should be adding this to `kernelci.data.kernelci_api` rather than creating requests from scratch in this file.

src/trigger.py
```diff
@@ -36,6 +36,8 @@ def _run_trigger(args, build_config, db):
     if args.skip_pull:
         print("Not updating repo")
     else:
+        node_list = get_node_by_commit_id(
+            kernelci.build.get_branch_head(build_config))
```
gctucker, line 40: The `node_list` variable is not used in this commit, so I think it makes more sense to squash it with the next one which checks if it's empty.

gctucker (commented):
LGTM, I think it should work but a few things can be tweaked to improve the code.

data/kernelci_api.py
```diff
@@ -0,0 +1,15 @@
+import urllib
```
gctucker, line 1: There is a `kernelci/data/kernelci_api.py` file in the `kernelci-core` repository.  So either we add support for getting a single object there, or we keep `get_node_by_commit_id()` in `trigger.py`.

src/trigger.py
```diff
@@ -16,12 +15,18 @@
 import kernelci.config
 import kernelci.data
 from kernelci.cli import Args, Command, parse_opts
+from kernelci.data import kernelci_api
```
gctucker, line 18: This won't import the local `data/kernelci_api.py` file, so I don't think this is going to work.  We either need to use a local function, or use `db.something()` with a new method added in `kernelci-core` as that's where the `db` object is coming from.

For an early prototype, we can use a local function to try things out.  Then when it's working, we can add support for this in `kernelci.data.Database` with a new method to retrieve database entries and use it instead.

JenySadadia: Yes, I'll move the function back to trigger.py. Once this gets merged, we'll move it to `kernelci-core`.

JenySadadia:
Hi @gctucker,
For the below task:
`If there isn't a Node with a revision with the git commit found, submit one with an empty describe and empty status. The API will then generate a pub/sub event.`
I have updated the code to send empty `describe`. In the case of `status`, we can not send empty status, otherwise will receive a validation error. We need to send bool value for it. I have removed the `status` parameter. Hence the node will have the default value `None` for it.

src/trigger.py
gctucker: Thanks, sorry if my previous review wasn't clear enough.

gctucker:
> Hi @gctucker, For the below task:
> 
> ```
> If there isn't a Node with a revision with the git commit found, submit one with an empty describe and empty status. The API will then generate a pub/sub event.
> ```
> 
> I have updated the code to send empty `describe`. In the case of `status`, we can not send empty status, otherwise will receive a validation error. We need to send bool value for it. I have removed the `status` parameter. Hence the node will have the default value `None` for it.

Sure, yes I think that's the way to create an node with an empty status.  You may also not send anything for describe, in fact it's probably a bit better than an empty string to make it clear that the field isn't being populated at all by `trigger.py`.

JenySadadia:
@gctucker We can not remove the describe from the node dictionary like we did for status.
```
curl -X 'POST'   'http://localhost:8001/node'   -H 'accept: application/json'   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IifQ.tYsu5nu2njDzrkrf_sVK1BRzCL6pUuop-YQuFnZ2xLo'   -H 'Content-Type: application/json'   -d '{
  "name":"checkout",
  "revision":{"tree":"mainline",
  "url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
  "branch":"master",
  "commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26"}
}'
{"detail":[{"loc":["body","revision","describe"],"msg":"field required","type":"value_error.missing"}]}
```

gctucker:
> @gctucker We can not remove the describe from the node dictionary like we did for status.
> 
> ```
> curl -X 'POST'   'http://localhost:8001/node'   -H 'accept: application/json'   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IifQ.tYsu5nu2njDzrkrf_sVK1BRzCL6pUuop-YQuFnZ2xLo'   -H 'Content-Type: application/json'   -d '{
>   "name":"checkout",
>   "revision":{"tree":"mainline",
>   "url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
>   "branch":"master",
>   "commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26"}
> }'
> {"detail":[{"loc":["body","revision","describe"],"msg":"field required","type":"value_error.missing"}]}
> ```

Ah, so maybe we need to make it an `Optional()` field in the model then.  The `describe` value has been used until now by kernelci-backend as a main identifier for kernel revisions with a human-readable string.  Now I think we may start relying on the git tree / branch / commit values instead, with an optional git describe string to help when it's available.

JenySadadia:
> > @gctucker We can not remove the describe from the node dictionary like we did for status.
> > ```
> > curl -X 'POST'   'http://localhost:8001/node'   -H 'accept: application/json'   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IifQ.tYsu5nu2njDzrkrf_sVK1BRzCL6pUuop-YQuFnZ2xLo'   -H 'Content-Type: application/json'   -d '{
> >   "name":"checkout",
> >   "revision":{"tree":"mainline",
> >   "url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
> >   "branch":"master",
> >   "commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26"}
> > }'
> > {"detail":[{"loc":["body","revision","describe"],"msg":"field required","type":"value_error.missing"}]}
> > ```
> 
> Ah, so maybe we need to make it an `Optional()` field in the model then. The `describe` value has been used until now by kernelci-backend as a main identifier for kernel revisions with a human-readable string. Now I think we may start relying on the git tree / branch / commit values instead, with an optional git describe string to help when it's available.

Created https://github.com/kernelci/kernelci-api/pull/98 to make it an Optional field.

gctucker:
> Created [kernelci/kernelci-api#98](https://github.com/kernelci/kernelci-api/pull/98) to make it an Optional field.

Great.  Have you confirmed that it worked now and the Node object is being accepted by the API without a `revision.describe` value?

JenySadadia:
> > Created [kernelci/kernelci-api#98](https://github.com/kernelci/kernelci-api/pull/98) to make it an Optional field.
> 
> Great. Have you confirmed that it worked now and the Node object is being accepted by the API without a `revision.describe` value?

Yes, I have confirmed it. It is working with `describe`.

src/trigger.py
```diff
@@ -16,12 +15,32 @@
...
         print("Not updating repo")
     else:
+        node_list = get_node_by_commit_id(
+            kernelci.build.get_branch_head(build_config))
+        if node_list:
+            print(f"Node exists with the latest git commit")
+            return
         print(f"Updating repo: {args.kdir}")
         sys.stdout.flush()
         kernelci.build.update_repo(build_config, args.kdir)
```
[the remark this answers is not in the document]
JenySadadia: As per the current code, we need to clone the repo once. As `cd /home/kernelci/data/linux` is used in various functions in build.py.
```
trigger     | Gathering revision meta-data
trigger     | /bin/sh: 3: cd: can't cd to /home/kernelci/data/linux
trigger     | Traceback (most recent call last):
trigger     |   File "/home/kernelci/./pipeline/trigger.py", line 112, in <module>
trigger     |     status = opts.command(configs, opts)
trigger     |   File "/home/kernelci/./pipeline/trigger.py", line 100, in __call__
trigger     |     _run_trigger(args, build_config, db)
trigger     |   File "/home/kernelci/./pipeline/trigger.py", line 51, in _run_trigger
trigger     |     res = step.run(opts={
trigger     |   File "/home/kernelci/.local/lib/python3.10/site-packages/kernelci/build.py", line 947, in run
trigger     |     revision['describe'] = git_describe(opts['tree'], self._kdir)
trigger     |   File "/home/kernelci/.local/lib/python3.10/site-packages/kernelci/build.py", line 234, in git_describe
trigger     |     describe = shell_cmd(cmd)
trigger     |   File "/home/kernelci/.local/lib/python3.10/site-packages/kernelci/__init__.py", line 29, in shell_cmd
trigger     |     return subprocess.check_output(cmd, shell=True).decode()
trigger     |   File "/usr/local/lib/python3.10/subprocess.py", line 420, in check_output
trigger     |     return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
trigger     |   File "/usr/local/lib/python3.10/subprocess.py", line 524, in run
trigger     |     raise CalledProcessError(retcode, process.args,
trigger     | subprocess.CalledProcessError: Command '
trigger     | set -e
trigger     | cd /home/kernelci/data/linux
trigger     | git describe --always
trigger     | ' returned non-zero exit status 2.
trigger exited with code 1
```

src/trigger.py
```diff
@@ -16,35 +15,48 @@
...
+    url = urllib.parse.urljoin(DEFAULT_URL, get_node_path)
+    res = requests.get(url).json()
+    print("get_node_by_commit_id response", res)
+    return res
 
 
 def _run_trigger(args, build_config, db):
     if args.skip_pull:
         print("Not updating repo")
     else:
```
gctucker, line 38: This can be dropped too, as there's no git repo to update any more.

gctucker:
I've tried it locally and made these changes to simplify the code, please feel free to pick bits from it:
```diff
diff --git a/src/trigger.py b/src/trigger.py
index 50b4d45caff2..1ad863de24dc 100755
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -27,37 +27,29 @@ def get_node_by_commit_id(commit_id):
     """
     get_node_path = '?'.join(['nodes', 'revision.commit='+commit_id])
     url = urllib.parse.urljoin(DEFAULT_URL, get_node_path)
-    res = requests.get(url).json()
-    print("get_node_by_commit_id response", res)
-    return res
+    return requests.get(url).json()
 
 
 def _run_trigger(args, build_config, db):
-    if args.skip_pull:
-        print("Not updating repo")
-    else:
-        head_commmit = kernelci.build.get_branch_head(build_config)
-        node_list = get_node_by_commit_id(head_commmit)
-        if node_list:
-            print(f"Node exists with the latest git commit")
-            return
-        sys.stdout.flush()
+    head_commit = kernelci.build.get_branch_head(build_config)
+    node_list = get_node_by_commit_id(head_commit)
+    if node_list:
+        print(f"Node exists with the latest git commit: {head_commit}")
+        return
+    sys.stdout.flush()
 
     revision = {
         'tree': build_config.tree.name,
         'url': build_config.tree.url,
         'branch': build_config.branch,
-        'commit': head_commmit
+        'commit': head_commmit,
     }
 
     print(f"Sending revision node to API: {revision['commit']}")
     sys.stdout.flush()
     node = {
         'name': 'checkout',
-        'revision': {
-            k: revision[k] for k in [
-                'tree', 'url', 'branch', 'commit',
-            ]},
+        'revision': revision,
     }
     resp_obj = db.submit({'node': node})[0]
     node_id = resp_obj['_id']
@@ -71,11 +63,6 @@ class cmd_run(Command):
         Args.kdir, Args.build_config, Args.output, Args.db_config,
     ]
     opt_args = [
-        {
-            'name': '--skip-pull',
-            'action': 'store_true',
-            'help': "Skip git pull",
-        },
         {
             'name': '--poll-period',
             'type': int,
```

gctucker:
@JenySadadia  The `runner.py` is now failing because it's expecting a `revision.describe` value, but that's expected.  I'll work on updating this part, while you could take a look at improving the `notifier.py` script in preparation for sending email reports.

JenySadadia:
> @JenySadadia The `runner.py` is now failing because it's expecting a `revision.describe` value, but that's expected. I'll work on updating this part, while you could take a look at improving the `notifier.py` script in preparation for sending email reports.

@gctucker , We have already made `revision.describe` Optional. The `runner.py` is failing because POST node request is receiving node id in the request json.
`Error: Object cannot be created with id: 61ef9728fd525cf02b33147b`

Need to fix it by using PUT request instead of POST.

gctucker (approved):

PR #9, opened 2022-01-28
src/trigger.py: drop --output and --kdir

gctucker:
Drop the now unused `--output` and `--kdir` command line arguments since there is no local git checkout any more. Update `kernelci.conf` accordingly.


gctucker (commented):
Tested OK locally

PR #10, opened 2022-01-28
Add tarball client

gctucker:
Add `tarball.py` to listen to `checkout` Node events without a status, then update a Git repo, push a source tarball and update the Node with `describe` string and status (`True` if everything passed).  Update config files accordingly.

Depends on https://github.com/kernelci/kernelci-api/pull/106


gctucker (commented):
Tested OK locally.  Merging now so we can unblock progress with other parts of the pipeline.

PR #20, opened 2022-02-11
docker-compose.yaml: add STORAGE_URL environment variable

gctucker:
Add STORAGE_URL environment variable to be able to configure the
--storage-url parameter used by the tarball.py script.  This is to
match the API configuration which might be serving files with an
alternative host or port number.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -38,6 +38,7 @@ services:
       - './pipeline/tarball.py'
       - '--settings=/home/kernelci/config/kernelci.conf'
       - 'run'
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging.

However there appears to be a problem with loading this value from `kernelci.conf`, if that worked we wouldn't need to pass it via the environment and docker-compose.  Investigating.

gctucker:
This is not needed any more now that there is #26 which allows the storage URL to be defined in kernelci.conf instead.

PR #21, opened 2022-02-14
config/pipeline.yaml: add KernelCI staging branches

gctucker:
Add build configurations for the KernelCI staging kernel branches so
the pipeline can be run on staging periodically.


gctucker (commented):
Tested OK on staging.

PR #31, opened 2022-03-01
docker/Dockerfile: bump kernelci-core with get_root_node()

gctucker:
Bump the version of kernelci-core to use get_root_node() with the
KernelCI API.


gctucker (commented):
Tested OK on staging

PR #32, opened 2022-03-08
Use event filters in clients

gctucker:
Use the new method from kernelci-core subscribe_node_channel() with filters to only receive nodes objects that match a particular set of criteria and simplify the logic in the client code accordingly.

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -81,23 +81,17 @@ class Runner:
         # ToDo: if stat != 0 then report error to API?
 
     def run(self):
-        sub_id = self._db.subscribe('node')
-        self._print("Listening for new checkout events")
+        sub_id = self._db.subscribe_node_channel(filters={
+            'op': 'updated',
+            'name': 'checkout',
+            'status': 'pass',
+        })
+        self._print("Listening for completed checkout events")
         self._print("Press Ctrl-C to stop.")
 
         try:
             while True:
-                sys.stdout.flush()
-                event = self._db.get_event(sub_id)
-                if event.data['op'] != 'updated':
-                    continue
-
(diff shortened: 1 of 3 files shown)


src/runner.py
```diff
@@ -81,23 +81,17 @@ def _cleanup_paths(self):
...
 
     def run(self):
-        sub_id = self._db.subscribe('node')
-        self._print("Listening for new checkout events")
+        sub_id = self._db.subscribe_node_channel(filters={
+            'op': 'updated',
+            'name': 'checkout',
+            'status': 'pass',
+        })
+        self._print("Listening for completed checkout events")
```
mgalka, line 89: as a follow up work we could get the `_print` method out of the class or even better switch to `logging` module.

mgalka (approved):
All good. I just made a note about a small class change, but this is a follow up.

src/runner.py
gctucker: Yes, maybe create a GitHub issue for it?

gctucker:
This was tested OK on staging this morning:
```
today at 8:10:07 AMListening for events... 
today at 8:10:07 AMPress Ctrl-C to stop.
today at 8:10:09 AMTime                        Commit        Status    Name
today at 8:10:09 AM2022-03-08 08:10:08.906433  abba0b6c99db  Pending   checkout
today at 8:11:00 AM2022-03-08 08:11:00.099377  abba0b6c99db  Pass      checkout
today at 8:11:00 AM2022-03-08 08:11:00.254930  abba0b6c99db  Pending   check-describe
today at 8:12:03 AM2022-03-08 08:12:03.275342  abba0b6c99db  Fail      check-describe
```

PR #43, opened 2022-04-21
src/runner.py: make --verbose an optional argument

gctucker:
Fix the --verbose argument to be optional rather than required.

Fixes: c712aedfe8f8 ("src/runner.py: add --verbose argument")


gctucker (commented):
Tested OK by hand and on staging.

PR #45, opened 2022-04-25
Integrate 'logging' module

JenySadadia:
Define handler, formatter, and other configuration
options for root and other loggers for client scripts.

### DIFF
--- a/src/logger.py
+++ b/src/logger.py
@@ -1,0 +1,32 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+"""Logger module"""
+
+import logging
+import logging.config
+
+
+class Logger:
+    """Logging utility class
+
+    This class accepts `name` of the logger in its contructor, by default
+    `root` will be used if not provided.
+    """
+
+    def __init__(self, name="root"):
+        self.logger = self.get_logger(name)
+
+    def get_logger(self, name):
+        """Returns logger object using configurations and logger name"""
+        logging.config.fileConfig('/home/kernelci/config/logger.conf')
+        self.logger = logging.getLogger(name)
+        return self.logger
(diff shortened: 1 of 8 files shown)


gctucker (commented):
Maybe the logger objects could be made part of the classes in each service file rather than global variables?  Then we would have `self._logger.log_message(...)` or even a shorter method name like `self._info(...)` which would internally call `self._logger.log_message(logging.INFO, ...)`.

src/logger.py
```diff
@@ -0,0 +1,32 @@
...
+    This class accepts `name` of the logger in its contructor, by default
+    `root` will be used if not provided.
+    """
+
+    def __init__(self, name="root"):
+        self.logger = self.get_logger(name)
+
+    def get_logger(self, name):
+        """Returns logger object using configurations and logger name"""
+        logging.config.fileConfig('/home/kernelci/config/logger.conf')
```
gctucker, line 26: This hard-coded path to the config file is likely to break at some point.  It means the code can only work inside the Docker image that sets up a `kernelci` user etc.  I think it would be better to have the config file name as an argument passed to `__init__()`.  Also this method could be renamed to `_get_logger()` as it's not really a public one but something used by the constructor to create the internal `self._logger` object.

gctucker (commented):
This works but I've added some comments about a few things that could be improved.

gctucker:
Merging this now to avoid conflicts with other changes in runner.py in particular.  Further changes may be done via follow-up PRs.

JenySadadia:
> Maybe the logger objects could be made part of the classes in each service file rather than global variables? Then we would have `self._logger.log_message(...)` or even a shorter method name like `self._info(...)` which would internally call `self._logger.log_message(logging.INFO, ...)`.

Yes, I will change it to `self._logger.log_message(...)`.
About `self._info(...)`, if it internally calls `self._logger.log_message(logging.INFO, ...)`, it can only be used for INFO logs. 
I have implemented a generic method that can be used to log messages with any log level.
e.g. we can log debug message with `self._logger.log_message(logging.DEBUG, ...)`.

PR #47, opened 2022-04-25
src/runner.py: load package templates

gctucker:
Load Jinja2 templates installed from the kernelci Python package.
These include the base templates for scripts and Kubernetes job
definitions.

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -62,8 +62,14 @@ class Runner:
         }
         params.update(self._plan_config.params)
         params.update(self._device_config.params)
+        config_path = self._runtime.config.config_path
+        templates_path = [
+            os.path.join(path, config_path)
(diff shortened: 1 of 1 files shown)


gctucker:
The template files are not installed in the same place with a system-wide `python3 setup.py install` and with the way kernelci-core is installed in the Dockerfile with `pip install`
```
/home/kernelci/.local/config/scripts
```
So it looks like the templates have to be installed in a sub-directory of the `kernelci` package in order to make this more reliable.  Then they should be in:
```
/home/kernelci/.local/lib/python3.10/site-packages/kernelci
```

gctucker:
OK that seems to be working locally, but some renaming will need to be done to not mix the `config` files with the `kernelci.config` package etc.

gctucker (commented):

PR #48, opened 2022-04-25
config/pipeline.yaml: add config_path to shell

gctucker:
Add the config_path attribute to the shell "lab" so that the scripts
templates installed from the kernelci Pyhon package can be found by
runner.py.

Depends on https://github.com/kernelci/kernelci-core/pull/1142


gctucker (commented):
Tested OK on staging

PR #49, opened 2022-04-26
T34238 Add kubernetes to pipeline.yaml

gctucker:
Add definitions for a Kubernetes lab and device type to run Python scripts in Kubernetes.


gctucker (commented):

PR #50, opened 2022-04-26
Add options for manual test runs rather than pub/sub

gctucker:
Add `--node-id` and `--git-commit` command line arguments to `runner.py` to specify a node or revision manually rather than listen for pub/sub events.

Depends on #45

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -24,6 +24,7 @@ from logger import Logger
 class Runner:
 
     def __init__(self, configs, args):
+        self._logger = Logger("config/logger.conf", "runner")
         self._db_config = configs['db_configs'][args.db_config]
         api_token = os.getenv('API_TOKEN')
         self._db = kernelci.db.get_db(self._db_config, api_token)
@@ -36,7 +37,8 @@ class Runner:
             os.makedirs(self._output)
         self._verbose = args.verbose
         self._job_tmp_dirs = {}
-        self._logger = Logger("config/logger.conf", "runner")
+        self._node_id = args.node_id
+        self._git_commit = args.git_commit
 
     def _create_node(self, checkout_node):
         node = {
@@ -76,6 +78,21 @@ class Runner:
         self._logger.log_message(logging.INFO, f"output_file: {output_file}")
         return output_file
 
+    def _schedule_test(self, checkout_node):
+        self._logger.log_message(logging.INFO, "Tarball: {}".format(
+            checkout_node['artifacts']['tarball']
+        ))
+
+        self._logger.log_message(logging.INFO, "Creating test node")
+        node = self._create_node(checkout_node)
+
+        tmp = tempfile.TemporaryDirectory(dir=self._output)
+        output_file = self._generate_job(node, tmp.name)
+
+        self._logger.log_message(logging.INFO, "Running test")
+        process = self._runtime.submit(output_file, get_process=True)
+        return process, tmp
+
     def _cleanup_paths(self):
         job_tmp_dirs = {
             process: tmp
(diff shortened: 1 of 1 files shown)


src/runner.py
```diff
@@ -99,34 +131,39 @@ def run(self):
...
                 self._cleanup_paths()
         except KeyboardInterrupt as e:
             self._logger.log_message(logging.INFO, "Stopping.")
         finally:
             self._db.unsubscribe(sub_id)
             self._cleanup_paths()
 
+    def run(self):
+        if self._node_id:
+            checkout_node = self._db.get_node(node_id)
```
JenySadadia, line 145:
```suggestion
            checkout_node = self._db.get_node(self._node_id)
```

src/runner.py
```diff
@@ -99,34 +131,39 @@ def run(self):
...
+            self._run_single_job(self._checkout_node)
```
JenySadadia, line 146:
```suggestion
            self._run_single_job(checkout_node)
```

Or else we can directly write,
```
self._run_single_job(self._db.get_node(self._node_id))
```
I think the first approach is better for code readability.

src/runner.py
JenySadadia, line 145: We need to check if node exists before running test to avoid errors like:
```File "/home/kernelci/./pipeline/runner.py", line 45, in _create_node
    'parent': checkout_node['_id'],
TypeError: 'NoneType' object is not subscriptable'
```

```suggestion
            checkout_node = self._db.get_node(self._node_id)
            if not checkout_node:
                self._logger.log_message(logging.ERROR,
                                         f"Node not found with id:\
{self._node_id}")
                return
            self._run_single_job(checkout_node)
```

src/runner.py
```diff
@@ -99,34 +131,39 @@ def run(self):
...
+        elif self._git_commit:
+            node = self._get_node_from_commit(self._git_commit)
```
JenySadadia, line 148:
```suggestion
            checkout_node = self._get_node_from_commit(self._git_commit)
            if not checkout_node:
                self._logger.log_message(logging.ERROR,
                                         f"Node not found with commit hash:\
{self._git_commit}")
                return
            self._run_single_job(checkout_node)
```
Also, `node` can be renamed to `checkout_node`.

src/runner.py
gctucker: I think 2 lines is better in this case. Thanks I'll fix it.

src/runner.py
gctucker: OK, however this is just one case of errors. The id could be a node that's not a checkout, or there could be an issue while communicating with the API. So I think it's better to handle it via exceptions.

src/runner.py
gctucker: Yes. But I don't like the idea of adding all the error handling code here again.

gctucker: I think I'll add exception handling in #59 instead which has a separate method for manual runs.

JenySadadia: Yes, we need to catch this too:
```
'tarball_url': node['artifacts']['tarball'],
TypeError: 'NoneType' object is not subscriptable
```

gctucker: Actually there's already some logic in place in this commit from #59 to return an error if the checkout node wasn't found using the command line arguments:
https://github.com/kernelci/kernelci-pipeline/pull/59/commits/49210ba5590cbb64012ec703a93064e50862eac5
```python
   def run(self):
        if self._node_id:
            checkout_node = self._db.get_node(self._node_id)
            self._run_single_job(checkout_node)
        elif self._git_commit:
            checkout_node = self._get_node_from_commit(self._git_commit)
            self._run_single_job(checkout_node)
        else:
            node = None

        if node is None:
            return False

        plan_config = self._plan_configs[args.plan]
        device_config = self._device_configs[args.target]
        return self._run_single_job(node, plan_config, device_config)
```
So I could add something there to print an error message when there's no node, and also another commit to handle exceptions for any unexpected issues.  If that's OK then we can merged this current PR #50 and I'll add the error handling improvements in #59.

JenySadadia: Sounds good to me.

JenySadadia (approved):

PR #42, opened 2022-04-21
config/scripts: remove base templates

gctucker:
Remove base templates in config/scripts as they are now being added to
kernelci-core and installed as part of the Python package.

Signed-off-by: Guillaume Tucker <guillaume.tucker@collabora.com>

Depends on https://github.com/kernelci/kernelci-core/pull/1141

PR #51, opened 2022-04-26
src/*.py: use kernelci.db

gctucker:
Update all the client scripts to use kernleci.db as it is being renamed from kernelci.data in kernelci-core.

### DIFF
--- a/src/notifier.py
+++ b/src/notifier.py
@@ -12,7 +12,7 @@ import sys
 
 import kernelci
 import kernelci.config
-import kernelci.data
+import kernelci.db
 from kernelci.cli import Args, Command, parse_opts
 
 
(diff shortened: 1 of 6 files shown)


JenySadadia (changes requested):
'base-python.jinja2' template should also be modified.

```
return kernelci.data.get_db(db_config, api_token)
```

gctucker:
> 'base-python.jinja2' template should also be modified.
> 
> ```
> return kernelci.data.get_db(db_config, api_token)
> ```

I left this unchanged because of #42, the base template is now in kernelci-core instead: https://github.com/kernelci/kernelci-core/pull/1151/files#diff-4907571040d6121e983188b3bad44465a3be8061724cb743dc56a733f7d9c7c5

There was another issue with #42 which is why it hasn't been merged yet, but I think it's all working now.

PR #52, opened 2022-04-26
config/kubernetes: add check-describe.jinja2

gctucker:
Add config/kubernetes/check-describe.jinja2 template which inherits
from the base-k8s-python.jinja2 template from kernelci-core.  This is
an intermediate step towards merging it with the
scripts/check-describe.jinja2 template which would inherit a different
base depending on the runtime environment (local shell or Kubernetes
etc.).


gctucker (commented):

PR #55, opened 2022-04-29
T34953 Refactor logger

JenySadadia:
src/logger.py: Refactor logger class
src/*.py: Remove global 'logger' variable

### DIFF
--- a/src/logger.py
+++ b/src/logger.py
@@ -14,19 +14,16 @@ import logging.config
 class Logger:
     """Logging utility class
 
-    This class accepts `name` of the logger in its contructor, by default
-    `root` will be used if not provided.
+    This class accepts name and path of the configuration file of the logger
+    in its contructor, by default `root` will be used as name of the logger
+    if not provided.
     """
 
-    def __init__(self, name="root"):
-        self.logger = self.get_logger(name)
-
-    def get_logger(self, name):
+    def __init__(self, config_path, name='root'):
         """Returns logger object using configurations and logger name"""
-        logging.config.fileConfig('/home/kernelci/config/logger.conf')
-        self.logger = logging.getLogger(name)
-        return self.logger
+        logging.config.fileConfig(config_path)
+        self._logger = logging.getLogger(name)
 
     def log_message(self, log_level, msg):
         """Creates a log record with provided log level and message"""
-        self.logger.log(log_level, msg)
+        self._logger.log(log_level, msg)
--- a/src/notifier.py
+++ b/src/notifier.py
@@ -19,12 +19,14 @@ from kernelci.cli import Args, Command, parse_opts
 
 from logger import Logger
 
-logger = Logger("notifier")
-
 
 class cmd_run(Command):
     help = "Listen for events and report them on stdout"
     args = [Args.db_config]
+
+    def __init__(self, sub_parser, name):
+        super().__init__(sub_parser, name)
+        self._logger = Logger("config/logger.conf", "notifier")
 
     def __call__(self, configs, args):
         log_fmt = "{time:26s}  {commit:12s}  {status:8s}  {name}"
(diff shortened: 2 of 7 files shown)


JenySadadia:
https://github.com/kernelci/kernelci-pipeline/pull/45#issuecomment-1112878302
> Maybe the logger objects could be made part of the classes in each service file rather than global variables? Then we would have self._logger.log_message(...) or even a shorter method name like self._info(...) which would internally call self._logger.log_message(logging.INFO, ...).

Yes, I have changed it to self._logger.log_message(...).
About self._info(...), if it internally calls self._logger.log_message(logging.INFO, ...), it can only be used for INFO logs.
I have implemented a generic method that can be used to log messages with any log level.
e.g. we can log debug message with self._logger.log_message(logging.DEBUG, ...).

JenySadadia:
Depends on https://github.com/kernelci/kernelci-pipeline/pull/56

src/notifier.py
```diff
@@ -19,12 +19,11 @@
 
 from logger import Logger
 
-logger = Logger("notifier")
-
 
 class cmd_run(Command):
     help = "Listen for events and report them on stdout"
     args = [Args.db_config]
+    _logger = Logger("config/logger.conf", "notifier")
```
gctucker, line 26: Oh I was thinking of creating one in the `__init__` constructor rather than making it a class attribute.  So if later on we want to have command line arguments to specify things like the logger config file path or the logging levels then we could use that when constructing the object.

gctucker: I see you're doing this in `runner.py` actually - so we should just add an `__init__()` in this class too and do the same.  In fact we may even make a derived class from `Command` with the logger support, or a mixin.  Then the classes in the pipeline scripts would inherit from it to avoid duplicating the initialisation etc. everywhere.

gctucker:
Adding `staging-skip` label for now so the other PRs can be tested on staging.

src/logger.py
```diff
@@ -18,15 +18,15 @@ class Logger:
...
+        self._logger = self._get_logger(config_path, name)
 
-    def get_logger(self, name):
+    def _get_logger(self, config_path, name):
         """Returns logger object using configurations and logger name"""
-        logging.config.fileConfig('/home/kernelci/config/logger.conf')
-        self.logger = logging.getLogger(name)
-        return self.logger
+        logging.config.fileConfig(config_path)
+        self._logger = logging.getLogger(name)
```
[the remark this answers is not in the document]
JenySadadia: Remove `get_logger` method and moved the code to the constructor.

src/notifier.py
JenySadadia: I added constructors to `notifier`, `send_kcidb`, and `trigger` scripts.

gctucker (approved):
Tested OK on staging:
```
2022-05-03T14:31:44.584276531Z 05/03/2022 02:31:44 PM UTC [INFO] Listening for events... 
2022-05-03T14:31:44.584329031Z 05/03/2022 02:31:44 PM UTC [INFO] Press Ctrl-C to stop.
2022-05-03T14:31:44.584423732Z 05/03/2022 02:31:44 PM UTC [INFO] Time                        Commit        Status    Name
2022-05-03T14:31:45.250180634Z 05/03/2022 02:31:45 PM UTC [INFO] 2022-05-03 14:31:45.196371  25a974e9bd70  Pending   checkout
2022-05-03T14:32:41.339243569Z 05/03/2022 02:32:41 PM UTC [INFO] 2022-05-03 14:32:41.292817  25a974e9bd70  Pass      checkout
2022-05-03T14:32:41.417115316Z 05/03/2022 02:32:41 PM UTC [INFO] 2022-05-03 14:32:41.381864  25a974e9bd70  Pending   check-describe
2022-05-03T14:33:25.444635866Z 05/03/2022 02:33:25 PM UTC [INFO] 2022-05-03 14:33:25.374305  25a974e9bd70  Fail      check-describe
```

PR #57, opened 2022-05-02
T35140 runner.py --lab-config

gctucker:
Add `--lab-config` option to `runner.py` to be able to specify which "lab" to use rather than the hard-coded 'shell'.

Depends on #50


JenySadadia (commented):
Looks fine to me.

PR #54, opened 2022-04-27
T34238 use k8s lab

gctucker:
Switch to Kubernetes for running the check-describe Python PoC test script.

Depends on #50 and #57

PR #58, opened 2022-05-03
T34238 drop get process

gctucker:
Depends on #54


gctucker (commented):
Tested OK on staging combined with https://github.com/kernelci/kernelci-core/pull/1162

PR #59, opened 2022-05-03
T35140 runner loop

gctucker:
Split the runner.py script into a "loop" command to listen to Pub/Sub
events and a "run" command to only run with a single checkout node.
Update docker-compose.yaml accordingly.

Depends on #58

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -28,8 +28,8 @@ class Runner:
         self._db_config = configs['db_configs'][args.db_config]
         api_token = os.getenv('API_TOKEN')
         self._db = kernelci.db.get_db(self._db_config, api_token)
-        self._plan_config = configs['test_plans'][args.plan]
-        self._device_config = configs['device_types']['shell_python']
+        self._plan_configs = configs['test_plans']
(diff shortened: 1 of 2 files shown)


gctucker:
Adding `staging-skip` already to first get the other PRs tested and merged as this rewrites some parts of it.

gctucker:
Tested manually with both shell and k8s, as well as on staging.

src/runner.py
```diff
@@ -169,8 +195,10 @@ class cmd_run(Command):
     ]
 
     def __call__(self, configs, args):
-        Runner(configs, args).run()
-        return True
+        if not args.node_id and not args.git_commit:
+            print("Either --node-id or --git-commit is required")
```
mgalka, line 199: Shouldn't this be printed on stderr?

mgalka (commented):
LGTM, just a small comment

src/runner.py
gctucker: Good point, I've fixed that now.  Also did this in #60 when printing errors from exceptions.

mgalka (approved):

PR #60, opened 2022-05-06
T35140 runner exceptions

gctucker:
As the "run" command is meant to be used interactively on the command
line rather than as a standalone service, print an error message and
exit cleanly when an exception is raised.

Depends on #59

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -198,8 +198,11 @@ class cmd_run(Command):
         if not args.node_id and not args.git_commit:
             print("Either --node-id or --git-commit is required",
                   file=sys.stderr)
+        try:
+            return Runner(configs, args).run(args)
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


JenySadadia (commented):
@gctucker How would cleanup happen in `run()`?
We haven't called `_cleanup_paths()` in the function.

gctucker:
> @gctucker How would cleanup happen in `run()`? We haven't called `_cleanup_paths()` in the function.

That's on purpose, the `run()` method doesn't leave processes running in the background so there's nothing to clean up.

gctucker:
I guess I could rework the code with 2 sub-classes, one for each command.  That way it would be clearer that the temporary job directories are being used only by the `loop` command.

JenySadadia (approved):

PR #63, opened 2022-05-31
Update email configs

alexandrasp:
Now that kernelci-core can load email configs we can update the pipeline code to use email configs coming from pipeline.yaml file. 

Depends on [#1231](https://github.com/kernelci/kernelci-core/pull/1231)

### DIFF
--- a/config/kernelci.conf
+++ b/config/kernelci.conf
@@ -1,5 +1,6 @@
 [DEFAULT]
 db_config: docker-host
+email_config: staging
 storage_url: http://172.17.0.1:8002/
 verbose: true
 
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2021 Collabora Limited
 # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
+# Author: Alexandra Pereira <alexandra.pereira@collabora.com>
 
 trees:
 
@@ -96,3 +97,11 @@ device_types:
     class: shell
     params:
       shebang: '/usr/bin/env python3'
+
+
+email_configs:
+
+  staging:
+    email_send_to: 'kernelci-results-staging@groups.io'
+    email_send_from: 'bot@kernelci.org'
+    email_subject: 'Kernel CI Test Reports'
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -103,5 +103,3 @@ email_configs:
 
   staging:
     email_send_to: 'kernelci-results-staging@groups.io'
-    email_send_from: 'bot@kernelci.org'
-    email_subject: 'Kernel CI Test Reports'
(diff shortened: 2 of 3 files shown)


mgalka (commented):
Please have a look at the comments

src/test_report.py
```diff
@@ -31,12 +31,13 @@ def __init__(self, configs, args):
...
         self._db = kernelci.db.get_db(self._db_config, api_token)
+        self._email_config = configs['email_configs'][args.email_config]
         self._logger = Logger("config/logger.conf", "test_report")
         self._email_host = args.email_host
         self._email_port = int(args.email_port)
-        self._email_send_from = 'bot@kernelci.org'
-        self._email_subject = 'Kernel CI Test Reports'
-        self._email_send_to = 'kernelci-results-staging@groups.io'
+        self._email_send_from = self._email_config.email_send_from
+        self._email_subject = self._email_config.email_subject
```
mgalka, line 39: I think these two were removed from the configuration file

alexandrasp: updated

config/pipeline.yaml
```diff
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2021 Collabora Limited
```
JenySadadia, line 3:
```suggestion
# Copyright (C) 2021, 2022 Collabora Limited
```

src/test_report.py
```diff
@@ -31,12 +31,13 @@ def __init__(self, configs, args):
         self._db_config = configs['db_configs'][args.db_config]
         api_token = os.getenv('API_TOKEN')
         self._db = kernelci.db.get_db(self._db_config, api_token)
+        self._email_config = configs['email_configs'][args.email_config]
         self._logger = Logger("config/logger.conf", "test_report")
         self._email_host = args.email_host
         self._email_port = int(args.email_port)
         self._email_send_from = 'bot@kernelci.org'
```
JenySadadia, line 38: Shouldn't you fetch the value of `self._email_send_from` from `pipeline.yaml` just like you did for `self._email_send_from`?

config/pipeline.yaml
```diff
@@ -96,3 +97,9 @@ device_types:
     class: shell
     params:
       shebang: '/usr/bin/env python3'
+
+
+email_configs:
+
+  staging:
+    email_send_to: 'kernelci-results-staging@groups.io'
```
gctucker, line 105: The issue here is that YAML configuration should be about listing the recipients for each combination of kernel tree and report type (build, test etc.).

config/kernelci.conf
```diff
@@ -1,5 +1,6 @@
 [DEFAULT]
 db_config: docker-host
+email_config: staging
```
gctucker, line 3: This entry should be about the email server to use, not the recipients.

gctucker (commented):
Maybe we should just close this PR? It hasn't been really discussed and I think it will need to be reverted if we mix up email server configuration (for `kernelci.conf`) with email reports configuration (for `email-configs.yaml`). A lot more work would be required to do this properly, and at the moment it's just as good to have the recipients hard-coded and the server options passed via the environment in docker-compose.

gctucker:
No replies after 3 weeks so closing this PR now.

PR #66, opened 2022-06-08
T35078 Add 'set_complete' script

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/163

Add a client script that marks the root node as 'complete' when all child nodes are completed.

### DIFF
--- a/config/kernelci.conf
+++ b/config/kernelci.conf
@@ -28,6 +28,8 @@ output: /home/kernelci/output
 
 [set_timeout]
 
+[set_complete]
+
 [db:docker-host]
 api: http://172.17.0.1:8001
 
--- a/src/set_complete.py
+++ b/src/set_complete.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+import logging
+import os
+import sys
+
+import kernelci
+import kernelci.config
+import kernelci.db
+from kernelci.cli import Args, Command, parse_opts
+
+from logger import Logger
+
+
+class SetComplete:
+
+    def __init__(self, configs, args):
+        self._db_config = configs['db_configs'][args.db_config]
+        api_token = os.getenv('API_TOKEN')
+        self._db = kernelci.db.get_db(self._db_config, api_token)
+        self._logger = Logger("config/logger.conf", "set_complete")
+
+    def run(self):
+        sub_id = self._db.subscribe_node_channel(filters={
+            'status': ('complete', 'timeout'),
+        })
+        self._logger.log_message(logging.INFO, "Listening for node events... ")
+        self._logger.log_message(logging.INFO, "Press Ctrl-C to stop.")
+
+        try:
+            while True:
+                node = self._db.receive_node(sub_id)
+                if node['parent'] is None:
+                    # Do nothing if root node received
+                    continue
+                sibling_nodes = self._db.get_nodes({"parent": node['parent']})
+                pending_sibling_node = False
+                for node in sibling_nodes:
+                    if node['status'] == "pending":
+                        pending_sibling_node = True
+                        break
+                if pending_sibling_node:
+                    continue
+                parent_node = self._db.get_node(node['parent'])
+                parent_node['status'] = "complete"
+                self._db.submit({'node': parent_node})
+        except KeyboardInterrupt as e:
+            self._logger.log_message(logging.INFO, "Stopping.")
+        finally:
+            self._db.unsubscribe(sub_id)
+
+
+class cmd_run(Command):
+    help = "Set parent node status to 'complete' when all child nodes are \
+completed"
+    args = [Args.db_config]
+
+    def __call__(self, configs, args):
+        set_complete = SetComplete(configs, args)
+        set_complete.run()
+
+
+if __name__ == '__main__':
+    opts = parse_opts('set_complete', globals())
+    configs = kernelci.config.load('config/pipeline.yaml')
+    status = opts.command(configs, opts)
+    sys.exit(0 if status is True else 1)
--- a/config/logger.conf
+++ b/config/logger.conf
@@ -4,7 +4,7 @@
 # Author: Jeny Sadadia <jeny.sadadia@collabora.com>
 
 [loggers]
-keys=root, notifier, runner, send_kcidb, tarball, test_report, trigger
+keys=root, notifier, runner, send_kcidb, set_complete, set_timeout, tarball, test_report, trigger
 
 [handlers]
 keys=consoleHandler
@@ -32,6 +32,12 @@ propagate=0
 level=DEBUG
 handlers=consoleHandler
 qualname=send_kcidb
+propagate=0
+
+[logger_set_complete]
+level=DEBUG
+handlers=consoleHandler
+qualname=set_complete
 propagate=0
 
 [logger_tarball]
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -103,6 +103,16 @@ services:
       - '--settings=${SETTINGS:-/home/kernelci/config/kernelci.conf}'
       - 'run'
 
+  set_complete:
+    <<: *base-service
+    container_name: 'kernelci-pipeline-set_complete'
+    command:
+      - '/usr/bin/env'
+      - 'python3'
+      - '/home/kernelci/pipeline/set_complete.py'
+      - '--settings=/home/kernelci/config/kernelci.conf'
+      - 'run'
+  
   set_timeout:
     <<: *base-service
     container_name: 'kernelci-pipeline-set_timeout'


config/kernelci.conf
```diff
@@ -25,8 +25,10 @@ lab_config: shell
 [kcidb]
 
 [test_report]
-email_host: your.host
-email_port: your.port
+email_host: mail.collabora.com
+email_port: 587
```
gctucker, line 29: Please drop these changes.  In fact I think the whole section should be removed, but that's beyond the scope of this PR.

JenySadadia: Oh, it is committed by mistake. I added it for testing only.

config/kernelci.conf
```diff
@@ -28,6 +28,8 @@ lab_config: shell
 email_host: your.host
 email_port: your.port
 
+[trigger_completed]
```
alexandrasp, line 31: @JenySadadia which kind of  `conf ` could we need to `trigger_completed`?

alexandrasp (commented):
In general the code looks good! I've used the code to automatically trigger completed events and send reports, so it is working.

config/kernelci.conf
JenySadadia: Atm, we don't have any configuration needed for the service. I have added it to the file for future purposes just like we have sections for `kcidb` and `notifier`.

gctucker: There's at least `db_config`:
```python
class cmd_run(Command):
    help = "Trigger an event if all child nodes are completed"
    args = [Args.db_config]
```

JenySadadia: Yes, I meant we don't need any additional configurations except the ones specified in `DEFAULT`.

src/trigger_completed.py
```diff
@@ -0,0 +1,66 @@
...
+        self._logger.log_message(logging.INFO, "Listening for node events... ")
+        self._logger.log_message(logging.INFO, "Press Ctrl-C to stop.")
+        sys.stdout.flush()
+
+        try:
+            while True:
+                node = self._db.receive_node(sub_id)
+                if node['parent'] is None:
+                    continue
+                resp = self._db.get_completed_event(node['parent'], seconds=60)
```
gctucker, line 42: See my other comment, we shouldn't be doing it this way. Passing a timeout in the API call and relying on the API to sleep is going to cause some big design issues. We should just be sleeping in the client, and then send a `POST` request to update nodes when the client has determined all the child nodes have completed.

gctucker (changes requested):
There's a problem with the design here, the sleep shouldn't be done in the API.

Also the commit message isn't very descriptive of what this does, it should mention that the parent is set to "completed" state when the service has detected that all child nodes have completed.

JenySadadia:
> There's a problem with the design here, the sleep shouldn't be done in the API.
> 
> Also the commit message isn't very descriptive of what this does, it should mention that the parent is set to "completed" state when the service has detected that all child nodes have completed.

My concern is what about the case when root node is in `completed` state and a new child with `pending` status is added to the API? @gctucker

JenySadadia:
Updated the client script to change root node status to 'completed' if all child nodes are completed.
Removed the logic to wait for the child node if incomplete.

gctucker:
> > There's a problem with the design here, the sleep shouldn't be done in the API.
> > Also the commit message isn't very descriptive of what this does, it should mention that the parent is set to "completed" state when the service has detected that all child nodes have completed.
> 
> My concern is what about the case when root node is in `completed` state and a new child with `pending` status is added to the API? @gctucker

I think we can leave the root node in `completed` state (or `complete`) to avoid oscillations.  It basically means, the root node has reached completion with all the child nodes at some point.  The updated time can be used to diagnose things as that would show when the root node status changed to `complete`.

With this approach, we don't need to implement anything special if a new child node is added later.  If there's some issues with this then we can review the situation later.

src/set_completed.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __init__(self, configs, args):
+        self._db_config = configs['db_configs'][args.db_config]
+        api_token = os.getenv('API_TOKEN')
+        self._db = kernelci.db.get_db(self._db_config, api_token)
+        self._logger = Logger("config/logger.conf", "set_completed")
+
+    def run(self):
+        sub_id = self._db.subscribe_node_channel(filters={
+            'op': ('created', 'updated'),
```
gctucker, line 30: If you drop this filter, then I think you'll get all the events in the same way.

src/set_completed.py
```diff
@@ -0,0 +1,78 @@
...
+            'status': ('completed', 'timeout'),
+        })
+        self._logger.log_message(logging.INFO, "Listening for node events... ")
+        self._logger.log_message(logging.INFO, "Press Ctrl-C to stop.")
+        sys.stdout.flush()
```
gctucker, line 35: This can be dropped as the logger takes care of it.

src/set_completed.py
```diff
@@ -0,0 +1,78 @@
...
+                resp = self._db.submit({'node': root_node})[0]
+                self._logger.log_message(logging.INFO, f"Set status to \
+completed of node: {resp['_id']}")
+                sys.stdout.flush()
+        except KeyboardInterrupt as e:
+            self._logger.log_message(logging.INFO, "Stopping.")
+        finally:
+            self._db.unsubscribe(sub_id)
+
+        sys.stdout.flush()
```
gctucker, line 61: Ditto

gctucker (commented):
Also I really think we should rename the state `completed` to `complete` as it's more correct.  The `timeout` state isn't a very good name either, but I can't think of a better one right now.  Maybe `expired`.

src/set_completed.py
JenySadadia: Yes, as now we have discarded the `completed` event, we can drop this.

JenySadadia:
> > > There's a problem with the design here, the sleep shouldn't be done in the API.
> > > Also the commit message isn't very descriptive of what this does, it should mention that the parent is set to "completed" state when the service has detected that all child nodes have completed.
> > 
> > 
> > My concern is what about the case when root node is in `completed` state and a new child with `pending` status is added to the API? @gctucker
> 
> I think we can leave the root node in `completed` state (or `complete`) to avoid oscillations. It basically means, the root node has reached completion with all the child nodes at some point. The updated time can be used to diagnose things as that would show when the root node status changed to `complete`.
> 
> With this approach, we don't need to implement anything special if a new child node is added later. If there's some issues with this then we can review the situation later.

Then the test reports will display pending child nodes as well.

gctucker:
Adding `skip` label to get the pending test report improvements merged.

gctucker:
> Then the test reports will display pending child nodes as well.

No, because the parent node will be set to complete when all the child nodes are complete.  If some other pending child nodes are added later then that's another issue.

src/set_completed.py
JenySadadia: The documentation states that the `StreamHandler.flush()` should be called explicitly at times. But I failed to get any other precise information about this.
https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler.flush

JenySadadia:
> Also I really think we should rename the state `completed` to `complete` as it's more correct. The `timeout` state isn't a very good name either, but I can't think of a better one right now. Maybe `expired`.

How about `lapsed` ?

gctucker:
Please rebase

JenySadadia:
Used `get_nodes` in place of `get_child_nodes_from_parent`.

src/set_complete.py
```diff
@@ -0,0 +1,73 @@
...
+                self._logger.log_message(logging.INFO, f"Set status to \
+complete of node: {resp['_id']}")
+        except KeyboardInterrupt as e:
+            self._logger.log_message(logging.INFO, "Stopping.")
+        finally:
+            self._db.unsubscribe(sub_id)
+
+
+class cmd_run(Command):
+    help = "Set root node status to 'complete' when all child nodes are \
```
gctucker, line 60: To keep using the same terminology everywhere:
```suggestion
    help = "Set parent node status to 'complete' when all child nodes are \
```

src/set_complete.py
```diff
@@ -0,0 +1,73 @@
...
+                    continue
+                nodes = self._db.get_nodes({"parent": node['parent']})
+                ret = False
+                for node in nodes:
+                    if node['status'] == "pending":
+                        ret = True
+                        break
+                if ret:
+                    continue
+                root_node = self._db.get_root_node(node['_id'])
```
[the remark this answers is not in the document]
JenySadadia: Nope, we get a complete node in event and get sibling nodes from it. Parent node will be set to complete rather than a root node. Added a few tweaks for it.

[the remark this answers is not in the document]
gctucker: Ah yes. I think the variable names are a bit ambiguous e.g. `node` and `nodes` could be anything.

Also `get_root_node()` is basically `get_node(node['parent'])`, maybe we don't need a special method for that.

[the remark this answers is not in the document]
gctucker: > I think the variable names are a bit ambiguous

Ah I see you've already updated that, thanks

JenySadadia:
Rebased.

JenySadadia:
As the node schema and pipeline flow has changed a bit, don't need to merge this PR anymore.
Hence, closing it.

PR #68, opened 2022-06-09
T35745 Fix permission issues with linux repository creation

JenySadadia:
Need to mount empty directory for linux source to have `kernelci` permission instead of `root` within tarball container.
Modify `kernelci.conf` for the same.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -48,7 +48,7 @@ services:
       - './src:/home/kernelci/pipeline'
       - './config:/home/kernelci/config'
       - './data/ssh:/home/kernelci/data/ssh'
-      - './data/linux:/home/kernelci/data/linux'
+      - './data/src:/home/kernelci/data/src'
       - './data/output:/home/kernelci/data/output'
 
   trigger:
--- a/config/kernelci.conf
+++ b/config/kernelci.conf
@@ -8,7 +8,7 @@ build_config: mainline
 poll_period: 0
 
 [tarball]
-kdir: /home/kernelci/data/linux
+kdir: /home/kernelci/data/src/linux
 output: /home/kernelci/data/output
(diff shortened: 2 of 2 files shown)


gctucker:
@JenySadadia Could you please rebase and drop the unrelated changes from this PR? I believe it can be done without depending on the other PRs that are about completed node events etc.

JenySadadia:
> @JenySadadia Could you please rebase and drop the unrelated changes from this PR? I believe it can be done without depending on the other PRs that are about completed node events etc.

Done.

docker-compose.yaml
```diff
@@ -48,7 +48,7 @@ services:
       - './src:/home/kernelci/pipeline'
       - './config:/home/kernelci/config'
       - './data/ssh:/home/kernelci/data/ssh'
-      - './data/linux:/home/kernelci/data/linux'
+      - './data/linux_src:/home/kernelci/data/linux_src'
```
gctucker, line 51:
```suggestion
      - './data/src:/home/kernelci/data/src'
```

gctucker (commented):
Having an `src` directory seems less confusing as it's not repeating `linux` in its name, and we can reuse it for other things to if needed as we may hit the same issue when doing other git clones.

JenySadadia:
> Having an `src` directory seems less confusing as it's not repeating `linux` in its name, and we can reuse it for other things to if needed as we may hit the same issue when doing other git clones.

Makes sense.

gctucker (commented):
Please also move the `.gitkeep` file to `data/src`

gctucker:
I've added a temporary patch on staging to do that (`git mv data/linux_src data/src`), the pipeline is running now.

gctucker (approved):
Tested OK on staging

PR #70, opened 2022-06-10
T35896 Refactor runner with subclasses

JenySadadia:
Create two separate subclasses of the runner to run it in a loop and a single job respectively.


gctucker:
@JenySadadia Please rebase without the dependency on #66 so we can have this merged first. Some rework is going to be needed for the node completion support.

gctucker (approved):
Tested OK on staging:
```
2022-07-20T06:11:30.563477764Z 07/20/2022 06:11:30 AM UTC [INFO] Listening for completed checkout events
2022-07-20T06:11:30.563597765Z 07/20/2022 06:11:30 AM UTC [INFO] Press Ctrl-C to stop.
2022-07-20T06:13:09.601702374Z 07/20/2022 06:13:09 AM UTC [INFO] Tarball: http://staging.kernelci.org:9080/linux-kernelci-staging-mainline-staging-mainline-20220719.0.tar.gz
2022-07-20T06:13:09.601746075Z 07/20/2022 06:13:09 AM UTC [INFO] Creating test node
2022-07-20T06:13:09.717504707Z 07/20/2022 06:13:09 AM UTC [INFO] Generating job
2022-07-20T06:13:09.717535108Z 07/20/2022 06:13:09 AM UTC [INFO] tmp: /home/kernelci/output/tmpthn4z1yb
2022-07-20T06:13:09.750853362Z 07/20/2022 06:13:09 AM UTC [INFO] output_file: /home/kernelci/output/tmpthn4z1yb/check-describe
2022-07-20T06:13:09.750891263Z 07/20/2022 06:13:09 AM UTC [INFO] Running test
2022-07-20T06:13:33.938959376Z Getting kernel source tree
2022-07-20T06:13:33.938982676Z Checking git describe
2022-07-20T06:13:33.938988376Z Result: FAIL
```

gctucker:
@JenySadadia As a follow-up, could you move the temporary file handling code to the Loop subclass? That way it won't be in the Runner one where it's not needed. I think this was the main point of splitting the commands into separate classes, so they can both have specific data fields associated with them.

JenySadadia:
> @JenySadadia As a follow-up, could you move the temporary file handling code to the Loop subclass? That way it won't be in the Runner one where it's not needed. I think this was the main point of splitting the commands into separate classes, so they can both have specific data fields associated with them.

Sure, will do that.

JenySadadia:
> @JenySadadia As a follow-up, could you move the temporary file handling code to the Loop subclass? That way it won't be in the Runner one where it's not needed. I think this was the main point of splitting the commands into separate classes, so they can both have specific data fields associated with them.

I have already moved cleanup for the temporary file from `Runner` to `RunnerLoop` in this one.
Minor changes still need to be done as we have discussed. Opened a follow-up PR https://github.com/kernelci/kernelci-pipeline/pull/80.

PR #71, opened 2022-06-21
T35078 Client script to set node status to timeout

JenySadadia:
Depends on https://github.com/kernelci/kernelci-core/pull/1265

Add a client script that periodically checks if node status is pending and changes its status to timeout if the maximum wait time is over.

### DIFF
--- a/src/set_timeout.py
+++ b/src/set_timeout.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+import logging
+import os
+import sys
+from datetime import datetime, timedelta
+from time import sleep
+
+import kernelci
+import kernelci.config
+import kernelci.db
+from kernelci.cli import Args, Command, parse_opts
+
+from logger import Logger
+
+
+class SetTimeout:
+
+    def __init__(self, configs, args):
+        self._db_config = configs['db_configs'][args.db_config]
+        api_token = os.getenv('API_TOKEN')
+        self._db = kernelci.db.get_db(self._db_config, api_token)
+        self._logger = Logger("config/logger.conf", "set_timeout")
+        self._poll_period = args.poll_period
+
+    def _update_pending_child(self, parent_id):
+        """Set child node status to timeout when parent is timed out"""
+        child_nodes = self._db.get_nodes({"status": "pending",
+                                          "parent": parent_id})
+        for child in child_nodes:
+            child['status'] = "timeout"
+            self._db.submit({'node': child})
+            self._update_pending_child(child["_id"])
+
+    def _set_timeout_status(self, node):
+        """Set Node status to timeout if maximum wait time is over"""
+        current_time = datetime.utcnow()
+        max_wait_time = datetime.fromisoformat(node['created']) + \
+            timedelta(hours=node['max_wait_time'])
+        if current_time > max_wait_time:
+            node['status'] = "timeout"
+            self._db.submit({'node': node})
+            self._update_pending_child(node["_id"])
+
+    def run(self):
+        self._logger.log_message(logging.INFO,
+                                 "Checking pending nodes...")
+        self._logger.log_message(logging.INFO, "Press Ctrl-C to stop.")
+
+        try:
+            while True:
+                nodes = self._db.get_nodes({"status": "pending"})
+                for node in nodes:
+                    self._set_timeout_status(node)
+                sleep(self._poll_period)
+        except KeyboardInterrupt as err:
+            self._logger.log_message(logging.INFO, "Stopping.")
+
+
+class cmd_run(Command):
+    help = "Set node status to timeout if maximum wait time is over"
+    args = [
+        Args.db_config,
+        {
+            'name': '--poll-period',
+            'type': int,
+            'help': "Polling period in seconds",
+            'default': 25,
+        },
+    ]
+
+    def __call__(self, configs, args):
+        set_timeout = SetTimeout(configs, args)
+        set_timeout.run()
+
+
+if __name__ == '__main__':
+    opts = parse_opts('set_timeout', globals())
+    configs = kernelci.config.load('config/pipeline.yaml')
+    status = opts.command(configs, opts)
+    sys.exit(0 if status is True else 1)
--- a/config/kernelci.conf
+++ b/config/kernelci.conf
@@ -26,6 +26,8 @@ output: /home/kernelci/output
 
 [complete_hack]
 
+[set_timeout]
+
 [db:docker-host]
 api: http://172.17.0.1:8001
 
--- a/config/logger.conf
+++ b/config/logger.conf
@@ -52,6 +52,12 @@ handlers=consoleHandler
 qualname=trigger
 propagate=0
 
+[logger_set_timeout]
+level=DEBUG
+handlers=consoleHandler
+qualname=set_timeout
+propagate=0
+
 [handler_consoleHandler]
 class=StreamHandler
 level=DEBUG
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -102,3 +102,13 @@ services:
       - '/home/kernelci/pipeline/test_report.py'
       - '--settings=${SETTINGS:-/home/kernelci/config/kernelci.conf}'
       - 'run'
+
+  set_timeout:
+    <<: *base-service
+    container_name: 'kernelci-pipeline-set_timeout'
+    command:
+      - '/usr/bin/env'
+      - 'python3'
+      - '/home/kernelci/pipeline/set_timeout.py'
+      - '--settings=${SETTINGS:-/home/kernelci/config/kernelci.conf}'
+      - 'run'


src/set_timeout.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def set_timeout_status(self, node):
+        """Set Node status to timeout if maximum wait time is over"""
+        current_time = datetime.utcnow()
+        max_wait_time = datetime.strptime(node['created'],
+                                          '%Y-%m-%dT%H:%M:%S.%f') + \
+            timedelta(hours=node['max_wait_time'])
+        if current_time > max_wait_time:
+            node['status'] = "timeout"
+            self._db.submit({'node': node})[0]
```
gctucker, line 38:
```suggestion
            self._db.submit({'node': node})
```

src/set_timeout.py
```diff
@@ -0,0 +1,76 @@
...
+        try:
+            while True:
+                nodes = self._db.get_nodes({"status": "pending"})
+                for node in nodes:
+                    if node['parent'] is None:
+                        # Do not set timeout for parent node
+                        continue
+                    if self.set_timeout_status(node):
+                        resp = self._db.submit({'node': node})[0]
+                        self._logger.log_message(logging.INFO, f"Set status \
```
gctucker, line 56: I think we can drop this, the notifier script should print a message when a node reaches timeout status.

src/set_timeout.py
```diff
@@ -0,0 +1,76 @@
...
+
+        try:
+            while True:
+                nodes = self._db.get_nodes({"status": "pending"})
+                for node in nodes:
+                    if node['parent'] is None:
+                        # Do not set timeout for parent node
+                        continue
+                    if self.set_timeout_status(node):
+                        resp = self._db.submit({'node': node})[0]
```
gctucker, line 55: The `set_timeout_status()` method already updates the node.

src/set_timeout.py
```diff
@@ -0,0 +1,76 @@
...
+        """Set Node status to timeout if maximum wait time is over"""
+        current_time = datetime.utcnow()
+        max_wait_time = datetime.strptime(node['created'],
+                                          '%Y-%m-%dT%H:%M:%S.%f') + \
+            timedelta(hours=node['max_wait_time'])
+        if current_time > max_wait_time:
+            node['status'] = "timeout"
+            self._db.submit({'node': node})[0]
+            return True
+        return False
```
gctucker, line 40: I don't think it's necessary to return this as it's not used in the main loop.

src/set_timeout.py
```diff
@@ -0,0 +1,76 @@
...
+
+class SetTimeout:
+
+    def __init__(self, configs, args):
+        self._db_config = configs['db_configs'][args.db_config]
+        api_token = os.getenv('API_TOKEN')
+        self._db = kernelci.db.get_db(self._db_config, api_token)
+        self._logger = Logger("config/logger.conf", "set_timeout")
+
+    def set_timeout_status(self, node):
```
gctucker, line 30: Maybe rename to `_set_timeout_status()` since it's a private method.

src/set_timeout.py
```diff
@@ -0,0 +1,76 @@
...
+                nodes = self._db.get_nodes({"status": "pending"})
+                for node in nodes:
+                    if node['parent'] is None:
+                        # Do not set timeout for parent node
+                        continue
+                    if self.set_timeout_status(node):
+                        resp = self._db.submit({'node': node})[0]
+                        self._logger.log_message(logging.INFO, f"Set status \
+to timeout of node: {resp['_id']}")
+                sleep(1800)  # sleep for 30 min before next execution
```
gctucker, line 58: This is a very long time to sleep. Ideally, the client should be spending most of the time listening for events with a timeout of a few seconds or maybe 1min in the method call (which would require some changes in `kernelci-core`). Then when returning from the method, if some event was received it needs to be processed by adding the node to the list of nodes to check and then it should go through all the nodes stored to check whether they have reached timeout.

JenySadadia: The script retrieves all the pending nodes in each execution rather than listening to any particular node event. Hence, there is no need to maintain any list of nodes.

gctucker: OK there are different ways to implement this, I guess we can stick to this approach for now.

src/set_timeout.py
```diff
@@ -0,0 +1,71 @@
...
+        self._db_config = configs['db_configs'][args.db_config]
+        api_token = os.getenv('API_TOKEN')
+        self._db = kernelci.db.get_db(self._db_config, api_token)
+        self._logger = Logger("config/logger.conf", "set_timeout")
+
+    def _set_timeout_status(self, node):
+        """Set Node status to timeout if maximum wait time is over"""
+        current_time = datetime.utcnow()
+        max_wait_time = datetime.strptime(node['created'],
+                                          '%Y-%m-%dT%H:%M:%S.%f') + \
```
gctucker, line 34: I thought the `created` time was already in ISO format? See how the `complete_hack.py` script uses that.

src/set_timeout.py
```diff
@@ -0,0 +1,71 @@
...
+            timedelta(hours=node['max_wait_time'])
+        if current_time > max_wait_time:
+            node['status'] = "timeout"
+            self._db.submit({'node': node})
+
+    def run(self):
+        self._logger.log_message(logging.INFO,
+                                 "Setting node status to timeout...")
```
gctucker, line 42: This message seems a bit misleading, it's not setting any particular node status to timeout at this point. So maybe something "starting loop" to know the service is ready would be more useful.

src/set_timeout.py
```diff
@@ -0,0 +1,71 @@
...
+        self._logger.log_message(logging.INFO, "Press Ctrl-C to stop.")
+
+        try:
+            while True:
+                nodes = self._db.get_nodes({"status": "pending"})
+                for node in nodes:
+                    if node['parent'] is None:
+                        # Do not set timeout for parent node
+                        continue
```
gctucker, line 51: Why is this a special case?

I think however that any pending child nodes would need to be set to timeout too if the parent timeout is reached.

src/set_timeout.py
```diff
@@ -0,0 +1,71 @@
...
+                    self._set_timeout_status(node)
+                sleep(60)  # sleep for 1 min before next execution
```
gctucker, line 53: This duration could actually be turned into a command line argument like for the `trigger` script.

src/set_timeout.py
JenySadadia: I was unsure about when to set a parent's status to timeout. It is possible that a child is created after maximum wait time of the parent. That is why I left it unchanged.
Added `_update_pending_child()` method to set pending child nodes to timeout if the parent is timed out.

gctucker: Yes I think that's fine, if a child gets added later then that's a special case. We could "lock" nodes if the state has to be consistent and not accept child nodes to be added once a parent node is complete, but I don't think it's necessarily going to be a problem.

src/set_timeout.py
```diff
@@ -0,0 +1,85 @@
...
+        self._logger = Logger("config/logger.conf", "set_timeout")
+        self._poll_period = args.poll_period
+
+    def _update_pending_child(self, parent_id):
+        """Set child node status to timeout when parent is timed out"""
+        child_nodes = self._db.get_nodes({"status": "pending",
+                                          "parent": parent_id})
+        for child in child_nodes:
+            child['status'] = "timeout"
+            self._db.submit({'node': child})
```
gctucker, line 37: This should be recursive, child nodes can have child nodes too.

src/set_timeout.py
JenySadadia: Okay. So, I think we can address that issue later if necessary.

src/set_timeout.py
JenySadadia: Yes, missed it.

gctucker: Note: It's the kind of thing we could then implement on the API side for efficiency, but for now let's just keep the logic entirely on the client side as it's easier to experiment this way.

gctucker:
OK it looks like it's all implemented now. We need to find a way to test this on staging, like pushing a node with a short timeout and then leave it to see what happens.  Have you tried something like this locally already?

JenySadadia:
> OK it looks like it's all implemented now. We need to find a way to test this on staging, like pushing a node with a short timeout and then leave it to see what happens. Have you tried something like this locally already?

Yes, I tested locally by changing created timestamp of `checkout` and `tarball` node and added another child to `checkout`. Tested OK on local.

gctucker:
Did you do that directly in the database or did you go via the API?  If you have a script maybe we could run it again on staging.  Also maybe we could start making end-to-end tests to validate this kind of feature.

JenySadadia:
> Did you do that directly in the database or did you go via the API? If you have a script maybe we could run it again on staging. Also maybe we could start making end-to-end tests to validate this kind of feature.

I tested using API by PUT node requests from the command line. I don't have any script for it but can create one for testing on staging. Another way is I can change `max_wait_time` from the node model and commit it just to test.

gctucker:
OK for now I think we'll just rely on your manual testing and check that nothing breaks as a side effect of the changes on staging.  Then we'll add some test harness for this kind of scenario as a follow-up so we don't block early development.

gctucker (approved):
Tested OK on staging

PR #76, opened 2022-07-11
T36485 Add placeholder files to data subdirectories

JenySadadia:
Add data/output and data/ssh subdirectories to source tree to avoid permission issues while creating files inside them from the docker container.


gctucker:
> add subdirectories to source

This doesn't seem very clear to me, I thought you were adding sub-directories to the source code in `src`. I think mentioning placeholder files for empty data directories would be less ambiguous.

JenySadadia:
> > add subdirectories to source
> 
> This doesn't seem very clear to me, I thought you were adding sub-directories to the source code in `src`. I think mentioning placeholder files for empty data directories would be less ambiguous.

Updated commit subject and message.

gctucker (approved):

PR #78, opened 2022-07-21
Fix up SMTP config

gctucker:
Fix the SMTP server config with sample values in comments.

### DIFF
--- a/config/kernelci.conf
+++ b/config/kernelci.conf
@@ -26,8 +26,6 @@ target: shell_python
 [kcidb]
 
 [test_report]
-email_host: your.host
-email_port: your.port
 
 [db:docker-host]
 api: http://172.17.0.1:8001
--- a/src/test_report.py
+++ b/src/test_report.py
@@ -32,8 +32,8 @@ class TestReport:
         api_token = os.getenv('API_TOKEN')
         self._db = kernelci.db.get_db(self._db_config, api_token)
         self._logger = Logger("config/logger.conf", "test_report")
-        self._email_host = args.email_host
-        self._email_port = int(args.email_port)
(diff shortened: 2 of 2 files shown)


config/kernelci.conf
```diff
@@ -25,8 +25,8 @@ lab_config: shell
 [kcidb]
 
 [test_report]
-email_host: your.host
-email_port: your.port
+# smtp_host: localhost
+# smtp_port: 25
```
JenySadadia, line 29:
```suggestion
smtp_host: localhost
smtp_port: 25
```
These lines should be uncommented.

gctucker: No they need to be in comments as sample values, then users have to explicitly un-comment them to use them.  If this is too confusing we can just drop them from the file.

JenySadadia: I don't think we should drop them. We can provide default values uncommented for which the code works.
Also, why haven't we added email sender and recipients to the configuration file? Why are they static ones?

gctucker: We can't provide them un-commented as that's something for the user to set up. There may not be an SMTP agent on the host, and the user may not (should not) be sending emails by default without knowing.

The email configuration should be in YAML, but that's a follow-up so at the moment it's hard-coded.

gctucker: OK I've dropped them to avoid confusion.

Putting sample config values in comments is very common practice for other projects but it's not something we've been doing for KernelCI config files until now, so I think we should either add a sample config files will everything in comments or nothing at all (opting for the latter here).

JenySadadia: Makes sense.

JenySadadia (approved):
Looks OK to me.

PR #82, opened 2022-07-22
Drop python suffix in device types

gctucker:
Drop the `_python` suffix in the device type names as the language being used is now defined in the test templates.


gctucker (commented):
Tested OK on staging with new pipeline and `kver` test, no issues reported so far.

PR #83, opened 2022-07-22
docker-compose.yaml: split runner-k8s as separate service

gctucker:
Split the runner-k8s service as a separate one from the basic runner.
That way, users can decide to enable one or the other or both.  Please
note that there aren't any mechanism yet for events to only get
handled by one service so all tarball nodes will result in the kver
test run both locally in a shell and in Kubernetes if both services
are running (and if Kubernetes credentials are installed).


gctucker:
This appears to be breaking the pipeline on staging, so adding the skip label for now.

gctucker:
This is now fixed with the `kunit` test plan running in Kubernetes as per the comments on #88

gctucker (commented):
Tested OK on staging.

PR #84, opened 2022-07-28
src/runner.py: add loop --plan argument

gctucker:
Add the --plan argument to the runner loop command to be able to
specify the test plan to run rather than having it hard-coded to
"kver".

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -87,9 +87,10 @@ class Runner:
 class RunnerLoop(Runner):
     """Runner subclass to execute in a loop"""
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, configs, args, **kwargs):
+        super().__init__(configs, args, **kwargs)
         self._job_tmp_dirs = {}
+        self._plan = self._plan_configs[args.plan]
 
     def _cleanup_paths(self):
         job_tmp_dirs = {
@@ -111,9 +112,6 @@ class RunnerLoop(Runner):
         self._logger.log_message(logging.INFO,
                                  "Press Ctrl-C to stop.")
 
-        # ToDo: iterate over test configs
-        plan = self._plan_configs['kver']
-
         # ToDo: iterate over device types for the current runtime
         device_type = self._runtime.config.lab_type
         device = self._device_configs.get(device_type)
@@ -126,8 +124,9 @@ class RunnerLoop(Runner):
         try:
             while True:
                 tarball_node = self._db.receive_node(sub_id)
-                job, tmp = self._schedule_test(tarball_node, plan,
-                                               device)
+                job, tmp = self._schedule_test(
+                    tarball_node, self._plan, device
+                )
                 if self._runtime.config.lab_type == 'shell':
                     self._job_tmp_dirs[job] = tmp
                 self._cleanup_paths()
@@ -179,7 +178,7 @@ class RunnerSingleJob(Runner):
 class cmd_loop(Command):
     help = "Listen to pub/sub events and run in a loop"
     args = [Args.db_config, Args.lab_config, Args.output]
-    opt_args = [Args.verbose]
+    opt_args = [Args.verbose, Args.plan]
 
     def __call__(self, configs, args):
         return RunnerLoop(configs, args).loop()


src/runner.py
```diff
@@ -87,9 +87,10 @@ def _schedule_test(self, tarball_node, plan, device):
 class RunnerLoop(Runner):
     """Runner subclass to execute in a loop"""
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, configs, args, **kwargs):
+        super().__init__(configs, args, **kwargs)
         self._job_tmp_dirs = {}
+        self._plan = self._plan_configs[args.plan]
```
JenySadadia, line 93: Getting error on this:

```
kernelci-pipeline-runner |     self._plan = self._plan_configs[args.plan]
kernelci-pipeline-runner | KeyError: 'check-describe'
```

Need to change below line in `kernelci.conf`:
```plan: check-describe```

**Review by JenySadadia — CHANGES_REQUESTED** [review]

File: src/runner.py (hunk 1 above)
Comment (JenySadadia) [inline src/runner.py:93]: `self._plan_configs[args.plan]` is used in `RunnerLoop` and `RunnerSingleJob` classes.
We can move it to `Runner` class.
```suggestion
```


The we can use it in `RunnerSingleJob.run()` too.
```
return self._run_single_job(tarball_node, self._plan,
                                    device_config)
```

JenySadadia (changes requested):

src/runner.py
JenySadadia: After I added `plan: kver` receiving errors:

```
kernelci-pipeline-runner | 07/29/2022 06:30:56 AM UTC [INFO] Tarball: http://172.17.0.1:8002/linux-mainline-master-v5.19-rc8-105-g6e2c0490769e.tar.gz
kernelci-pipeline-runner | 07/29/2022 06:30:56 AM UTC [INFO] Creating test node
kernelci-pipeline-runner | 07/29/2022 06:30:56 AM UTC [INFO] Generating job
kernelci-pipeline-runner | 07/29/2022 06:30:56 AM UTC [INFO] tmp: /home/kernelci/output/tmpxwsfewf8
kernelci-pipeline-notifier | 07/29/2022 06:30:56 AM UTC [INFO] 2022-07-29 06:30:56.859930  6e2c0490769e  Pending   NA       kver
kernelci-pipeline-runner | 07/29/2022 06:30:56 AM UTC [INFO] output_file: /home/kernelci/output/tmpxwsfewf8/kver
kernelci-pipeline-runner | 07/29/2022 06:30:56 AM UTC [INFO] Running test
kernelci-pipeline-runner | import-im6.q16: unable to open X server `' @ error/import.c/ImportImageCommand/346.
kernelci-pipeline-runner | import-im6.q16: unable to open X server `' @ error/import.c/ImportImageCommand/346.
kernelci-pipeline-runner | import-im6.q16: unable to open X server `' @ error/import.c/ImportImageCommand/346.
kernelci-pipeline-runner | import-im6.q16: unable to open X server `' @ error/import.c/ImportImageCommand/346.
kernelci-pipeline-runner | import-im6.q16: unable to open X server `' @ error/import.c/ImportImageCommand/346.
kernelci-pipeline-runner | import-im6.q16: unable to open X server `' @ error/import.c/ImportImageCommand/346.
kernelci-pipeline-runner | import-im6.q16: unable to open X server `' @ error/import.c/ImportImageCommand/346.
kernelci-pipeline-runner | import-im6.q16: unable to open X server `' @ error/import.c/ImportImageCommand/346.
kernelci-pipeline-runner | /home/kernelci/output/tmpxwsfewf8/kver: 13: DB_CONFIG_YAML: not found
kernelci-pipeline-runner | /home/kernelci/output/tmpxwsfewf8/kver: 18: NODE_ID: not found
kernelci-pipeline-runner | /home/kernelci/output/tmpxwsfewf8/kver: 19: TARBALL_URL: not found
kernelci-pipeline-runner | /home/kernelci/output/tmpxwsfewf8/kver: 20: WORKSPACE: not found
kernelci-pipeline-runner | /home/kernelci/output/tmpxwsfewf8/kver: 21: REVISION: not found
kernelci-pipeline-runner | /home/kernelci/output/tmpxwsfewf8/kver: 25: Syntax error: "(" unexpected

```

gctucker: Yes that's fixed by other PRs. The `runtime` variable needs to be passed to the template, and I've dropped the default test plan names in `kernelci.conf` as it's now defined in `docker-compose.yaml`.

gctucker: The difference is that `run` should be getting the test plan name from the command line argument, but `loop` should ideally be reading it from the YAML configuration like `kci_test`.  In fact it should get merged with `kci_test` eventually so we can run the new pipeline jobs with it too, but at the moment the list of template parameters is very different compared to LAVA jobs.  So in the meantime, there's a fixed test plan config stored in the object but it should be removed.  And for the `run` command, it shouldn't be stored in the object as it's specific to the call to the `run` method.

JenySadadia: I think the fixes for this issue should be included in this PR itself. Then we should merge this.
Or else the PR with fixes should be mentioned in the description as a dependency and the former should be merged first.

gctucker: Yes, there are just too many dependencies that I didn't list them all here.

gctucker: Actually I don't think this particular PR depends on any other changes, the `runtime` template parameter is already provided in the `main` branch. So I think the issue might be that you're testing with an older version of the `kernelci` package.

gctucker (commented):
Tested OK on staging.

PR #87, opened 2022-07-28
runner.py: add runtime_image template parameter

gctucker:
Add the runtime_image template parameter currently only used by
base/kubernetes.jinja2 to specify the Docker image to use when running
the test plan.


gctucker (commented):
Tested OK on staging.

PR #100, opened 2022-08-08
src/fstests: add README.md about PoC

### LINKED ISSUE #97: Start README.md about how to set up the environment
A `README.md` file should be added to the `fstests` directory with instructions to get started.  This should be based on the quick start guide from `kvm-xfstests` with some considerations about KernelCI.

- [ ] README.md file created with instructions to get started and run the `smoke` tests
- [ ] Instructions verified

gctucker:
Add a readme about the Proof-of-Concept project for running fstests in
KernelCI in a new sub-directory src/fstests.

Fixes #97

### DIFF
--- a/src/fstests/README.md
+++ b/src/fstests/README.md
@@ -1,0 +1,148 @@
+KernelCI fstests PoC
+====================
+
+
+## Overview
+
+This page contains information about the Proof-of-Concept (PoC) project to run
+fstests automatically using KernelCI's new API.  It is based on `kvm-xfstests`
+which starts a local VM with `kvm` natively on the host where it is running.
+This solution does not involve any Docker container as all the environment is
+included in the VM image.
+
+
+## Prerequisites
+
+A number of dependencies need to be installed in oder to run `kvm-xfstests` and
+also to build a Linux kernel natively.
+
+For running `kvm-xfstests`:
+
+    sudo apt install -y qemu-kvm
+
+For building `xfstests`: see the documentation about how to [build without a
+chroot](https://github.com/tytso/xfstests-bld/blob/master/Documentation/building-xfstests.md#without-a-build-chroot).  Typically:
+
+    sudo apt install -y \
+        autoconf autoconf2.64 automake autopoint bison build-essential \
+        ca-certificates debootstrap e2fslibs-dev ed fakechroot gettext git \
+        golang-1.13-go libblkid-dev libdbus-1-3 libgdbm-dev libicu-dev \
+        libkeyutils-dev libssl-dev libsystemd-dev libtool-bin liburcu-dev \
+        lsb-release meson pkg-config python3-setuptools rsync symlinks \
+        qemu-utils uuid-dev zlib1g-dev
+
+For building the Linux kernel:
+
+    sudo apt install -y bc bison cpio flex gcc kmod libssl-dev libelf-dev
+
+> **Note** As a follow-up improvement, the kernel builds could be done in
+> Docker containers like any regular build produced by KernelCI.  Building
+> natively is to simplify the process for the PoC.
+
+For more details, see the [quick start guide for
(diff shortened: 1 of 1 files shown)


src/fstests/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+    sudo apt install -y qemu-kvm
+
+For building `xfstests`: see the documentation about how to [build without a
+chroot](https://github.com/tytso/xfstests-bld/blob/master/Documentation/building-xfstests.md#without-a-build-chroot).  Typically:
+
+    sudo apt install -y \
+        autoconf autoconf2.64 automake autopoint bison build-essential \
+        ca-certificates debootstrap e2fslibs-dev ed fakechroot gettext git \
+        golang-1.13-go libblkid-dev libdbus-1-3 libgdbm-dev libicu-dev \
```
alexandrasp, line 29: golang version may be different..are we keeping 1.15 for any special reason? I noticed as well that the version in the xfstests script is 1.11

src/fstests/README.md
```diff
@@ -0,0 +1,148 @@
...
+    git checkout FETCH_HEAD -b linux-5.19
+
+Generate the kernel config for fstests and build the kernel image (no modules):
+
+    kvm-xfstests install-kconfig
+    make -j$(nproc)
+
+
+## Running the tests
+
```
alexandrasp, line 106: Maybe could be interesting to add here about the update in kvm.config file to set the kernel image to be used.

src/fstests/README.md
```diff
@@ -0,0 +1,148 @@
...
+Or to run just a single set of tests, which should take about 1min:
+
+    kvm-xfstests generic/001
+
+To get the XML results file , run a shell in the VM and copy the file to the
+host:
+
+    $ kvm-xfstests shell
+    # cp /results/ext4/results-4k/results.xml /vtmp/
```
alexandrasp, line 120: is the `v` in vtmp a typo?

src/fstests/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+It should start with something like this after running `generic/001`:
+
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<testsuite name="xfstests" failures="0" skipped="0" tests="1" time="4" hostname="kvm-xfstests" timestamp="2022-08-08T04:09:42">
+```
+
+or after running the full `smoke` tests:
+
```
alexandrasp, line 132: Maybe it could also be interesting to add in some next PR some details about the fails. I can do this as well..because when the test pass we won't have any special information, but when it fails we will have a diff showing why it fails.

src/fstests/README.md
gctucker: I changed it to `1.13` as that's what worked when I installed it.  It's not guaranteed to work on every Debian distro, some adjustments may need to be made.  This is just an indication of the dependencies.

src/fstests/README.md
gctucker: It's not really necessary as the script will pick the image from the default path when following the instructions here.  More advanced thins can be done as per the full documentation of `xfstests-bld` and I've put a link to that.

src/fstests/README.md
gctucker: No it's not a typo. Why? Did you not have `/vtmp`?

src/fstests/README.md
gctucker: That's about the format of the output results, I don't think we need to describe it in this README.

src/fstests/README.md
alexandrasp: alright!

src/fstests/README.md
alexandrasp: no, I didn't.. but I just put it on another path and works fine.

src/fstests/README.md
alexandrasp: ok!

alexandrasp (approved):
LGTM

PR #101, opened 2022-08-10
src/fstests: listening for new tarball events

alexandrasp:
Add a new class to handle fstests on Kernel CI. This first commit is listering for tarball events and print a message if it receives any tarball.

### DIFF
--- a/src/fstests/runner.py
+++ b/src/fstests/runner.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022 Collabora Limited
+# Author: Alexandra Pereira <alexandra.pereira@collabora.com>
+
+import os
+import sys
+
+import kernelci
+import kernelci.config
+import kernelci.db
+import kernelci.lab
+from kernelci.cli import Args, Command, parse_opts
+
+
+class FstestsRunner:
+    def __init__(self, configs, args):
+        self._db_config = configs['db_configs'][args.db_config]
+        api_token = os.getenv('API_TOKEN')
+        self._db = kernelci.db.get_db(self._db_config, api_token)
+
+    def run(self):
+        sub_id = self._db.subscribe_node_channel({
+            'op': 'created',
+            'name': 'tarball',
+        })
+        print('Listening for tarballs')
+        print('Press Ctrl-C to stop.')
+        tarball_node = None
+        try:
+            while True:
+                tarball_node = self._db.receive_node(sub_id)
+                print(f"Node tarball with id: {tarball_node['_id']}\
+                    from revision: {tarball_node['revision']['commit'][:12]}")
+        except KeyboardInterrupt as e:
+            print('Stopping.')
+        except Exception as e:
+            print('Error', e)
+        finally:
+            self._db.unsubscribe(sub_id)
+            return True
+
+
+class cmd_run(Command):
+    help = 'KVM fstests runner'
+    args = [Args.db_config]
+    opt_args = [Args.verbose]
+
+    def __call__(self, configs, args):
+        return FstestsRunner(configs, args).run()
+
+
+if __name__ == '__main__':
+    opts = parse_opts('fstests_runner', globals())
+    configs = kernelci.config.load('config/pipeline.yaml')
+    status = opts.command(configs, opts)
+    sys.exit(0 if status is True else 1)
--- a/src/fstests/runner.py
+++ b/src/fstests/runner.py
@@ -28,7 +28,6 @@ class FstestsRunner:
         })
         print('Listening for tarballs')
         print('Press Ctrl-C to stop.')
-        tarball_node = None
         try:
             while True:
                 tarball_node = self._db.receive_node(sub_id)


alexandrasp:
hey @gctucker I have some comments to do about this PR.

1. I spent some minutes trying to import/use logger file into `fstests` directory, do have any idea how to make this work? I've tried to include a `__init__.py` in `src` and import this file as a relative with `..` it does not work! Maybe it is time to turn the logger into a library than a script. What do you suggest? 

2. How I do test it: Create a node tarball and it printed the message: 
`kernelci-pipeline-fstests_runner | 08/10/2022 10:33:35 PM UTC [INFO] New Tarball :)`

3. I got some fails from `kernelci-pipeline-complete-hack`, is this expected?
```
kernelci-pipeline-complete-hack | 08/10/2022 10:33:40 PM UTC [ERROR] Traceback (most recent call last):
kernelci-pipeline-complete-hack |   File "/home/kernelci/./pipeline/complete_hack.py", line 58, in run
kernelci-pipeline-complete-hack |     self._check_pending_node()
kernelci-pipeline-complete-hack |   File "/home/kernelci/./pipeline/complete_hack.py", line 39, in _check_pending_node
kernelci-pipeline-complete-hack |     checkout_node = self._db.get_node(self._pending_node['parent'])
kernelci-pipeline-complete-hack |   File "/usr/local/lib/python3.10/site-packages/kernelci-1.1-py3.10.egg/kernelci/db/kernelci_api.py", line 97, in get_node
kernelci-pipeline-complete-hack |     resp = self._get('/'.join(['node', node_id]))
kernelci-pipeline-complete-hack | TypeError: sequence item 1: expected str instance, NoneType found
kernelci-pipeline-complete-hack | 
```

gctucker:
> hey @gctucker I have some comments to do about this PR.
> 
> 1. I spent some minutes trying to import/use logger file into `fstests` directory, do have any idea how to make this work? I've tried to include a `__init__.py` in `src` and import this file as a relative with `..` it does not work! Maybe it is time to turn the logger into a library than a script. What do you suggest?

I suggest to use `print()` for now as there aren't any other local dependencies.  It looks like we'll have to rework the whole directory structure to turn it into a Python package if we want to be able to import other modules.

> 2. How I do test it: Create a node tarball and it printed the message:
>    `kernelci-pipeline-fstests_runner | 08/10/2022 10:33:35 PM UTC [INFO] New Tarball :)`

Looks good.

> 3. I got some fails from `kernelci-pipeline-complete-hack`, is this expected?
> 
> ```
> kernelci-pipeline-complete-hack | 08/10/2022 10:33:40 PM UTC [ERROR] Traceback (most recent call last):
> kernelci-pipeline-complete-hack |   File "/home/kernelci/./pipeline/complete_hack.py", line 58, in run
> kernelci-pipeline-complete-hack |     self._check_pending_node()
> kernelci-pipeline-complete-hack |   File "/home/kernelci/./pipeline/complete_hack.py", line 39, in _check_pending_node
> kernelci-pipeline-complete-hack |     checkout_node = self._db.get_node(self._pending_node['parent'])
> kernelci-pipeline-complete-hack |   File "/usr/local/lib/python3.10/site-packages/kernelci-1.1-py3.10.egg/kernelci/db/kernelci_api.py", line 97, in get_node
> kernelci-pipeline-complete-hack |     resp = self._get('/'.join(['node', node_id]))
> kernelci-pipeline-complete-hack | TypeError: sequence item 1: expected str instance, NoneType found
> kernelci-pipeline-complete-hack | 
> ```

I think you can ignore `complete_hack` for now.

docker-compose.yaml
```diff
@@ -80,6 +80,16 @@ services:
...
+  fstests-runner:
+    <<: *base-service
+    container_name: 'kernelci-pipeline-fstests_runner'
+    command:
+      - '/usr/bin/env'
+      - 'python3'
+      - '/home/kernelci/pipeline/fstests_runner.py'
+      - '--settings=${SETTINGS:-/home/kernelci/config/kernelci.conf}'
+      - 'run'
+
```
gctucker, line 92: Please drop this, we can't run `kvm-xfstests` inside a container.  The script will be run manually directly on the host instead.

src/fstests_runner.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022 Collabora Limited
+# Author: Alexandra Pereira <alexandra.pereira@collabora.com>
+
+import logging
```
gctucker, line 8: As per my previous comment, this can be dropped for now as we need to turn the whole `src` directory into a Python package to have relative imports.  Please use `print()` for now, which is perfectly fine for the PoC as the command is run natively on the host rather than in `docker-compose`.

src/fstests_runner.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, configs, args):
+        self._db_config = configs['db_configs'][args.db_config]
+        api_token = os.getenv('API_TOKEN')
+        self._db = kernelci.db.get_db(self._db_config, api_token)
+        self._logger = Logger("config/logger.conf", "complete")
+        self._tarball_node = None
+
+    def run(self):
+        sub_id = self._db.subscribe_node_channel({
+            'name': 'tarball',
```
gctucker, line 31: We need to listen for events about new tarball nodes, not all events about tarballs.  So you can use the `'op': 'created'` filter for that.

src/fstests_runner.py
```diff
@@ -0,0 +1,65 @@
...
+        })
+        self._logger.log_message(logging.INFO, "Listening for tarballs")
+        self._logger.log_message(logging.INFO, "Press Ctrl-C to stop.")
+
+        try:
+            while True:
+                if not self._tarball_node:
+                    self._tarball_node = self._db.receive_node(sub_id)
```
gctucker, line 39: This should be a local variable rather than an object member:
```suggestion
                    tarball_node_id = self._db.receive_node(sub_id)
```

src/fstests_runner.py
```diff
@@ -0,0 +1,65 @@
...
+                    self._logger.log_message(logging.INFO, "New Tarball :)")
```
gctucker, line 40: Maybe also get the actual node data and print the git revision from it like the `notifier` script does, to track which nodes you get events from.

src/fstests_runner.py
```diff
@@ -0,0 +1,65 @@
...
+            self._logger.log_message(logging.INFO, "Stopping.")
+        except Exception as e:
+            self._logger.log_message(logging.ERROR, e)
+        finally:
+            self._db.unsubscribe(sub_id)
+            return True
+
+
+class cmd_run(Command):
+    help = "KVM-FS Tests on KernelCI"
```
gctucker, line 52:
```suggestion
    help = "KVM fstests runner"
```

src/fstests_runner.py
```diff
@@ -0,0 +1,65 @@
...
+    args = [Args.db_config]
+    opt_args = [Args.verbose]
+
+    def __call__(self, configs, args):
+        FstestsRunner(configs, args).run()
+        return True
```
gctucker, line 58:
```suggestion
        return FstestsRunner(configs, args).run()
```

gctucker (commented):
A few things to tweak but looks good otherwise.

src/fstests_runner.py
gctucker: Also the script can be moved to `fstests/runner.py` once this has been dropped.  It's good to keep it next to the README as an experimental script.

gctucker:
One way to solve the import issue is to have this in the script in `src/fstests/runner.py`:
```python
import logging
from src.logger import Logger
```

Then run the script with `python3 -m src.fstests.runner`.  That way, imports will be from the top-level source directory.

docker-compose.yaml
alexandrasp: updated

src/fstests_runner.py
alexandrasp: updated

src/fstests_runner.py
alexandrasp: updated

src/fstests_runner.py
alexandrasp: updated

src/fstests_runner.py
alexandrasp: The new output will be like this: `New tarball from revision: babf0bb978e3`

src/fstests_runner.py
alexandrasp: updated

src/fstests_runner.py
alexandrasp: updated

alexandrasp:
> One way to solve the import issue is to have this in the script in `src/fstests/runner.py`:
> 
> ```python
> import logging
> from src.logger import Logger
> ```
> 
> Then run the script with `python3 -m src.fstests.runner`. That way, imports will be from the top-level source directory.

I will include this on my next PR, for this, I just use `prinf` as recommended.

src/fstests/runner.py
```diff
@@ -0,0 +1,61 @@
...
+        print('Listening for tarballs')
+        print('Press Ctrl-C to stop.')
+        tarball_node_id = None
+        try:
+            while True:
+                if not tarball_node_id:
+                    tarball_node_id = self._db.receive_node(sub_id)
+                    print(f"New tarball from revision:\
+                         {tarball_node_id['revision']['commit'][:12]}")
+                    break
```
gctucker, line 38: Just fixing the logic here, if it's a loop it's a real loop:
```suggestion
        try:
            while True:
                tarball_node_id = self._db.receive_node(sub_id)
                print(f"New tarball from revision:\
                     {tarball_node_id['revision']['commit'][:12]}")
```

gctucker: Ah actually it't not `tarball_node_id` but `tarball_node` as it's the actual node data and not just the id.  The variable name is misleading.

gctucker: You could also print the tarball node id next to the git revision as that might be useful when debugging, in case the same git revision gets tested several times and new nodes get created for it.

src/fstests/runner.py
```diff
@@ -0,0 +1,59 @@
...
+            print('Stopping.')
+        except Exception as e:
+            print('Error', e)
+        finally:
+            self._db.unsubscribe(sub_id)
+            return True
+
+
+class cmd_run(Command):
+    help = 'KVM-FS fstests runner'
```
gctucker, line 47:
```suggestion
    help = 'KVM fstests runner'
```
`KVM` stands for kernel virtual machine.  `KVM-FS` doesn't mean anything.

gctucker:
@alexandrasp How are you testing this at the moment?

alexandrasp:
> @alexandrasp How are you testing this at the moment?

At moment, I started a bash terminal in a runner and then I can run the fstest runner..
Then with the containers running I can create a new node and get a message from the terminal where I started the script:
```
kernelci@8630f52ab319:~$ python3 pipeline/fstests/runner.py run --db-config=docker-host 
Listening for tarballs
Press Ctrl-C to stop.
Node tarball with id: 62f5551a9498b5b62ac742dd                    from revision: babf0bb978e3
```

alexandrasp:
> 

Just to clarify, I know that I need to connect to db in API. But I can update this in some next PR, because I am using subprocess and I will start running the commands from host (`kernelci2`) that I believe won´t  have the kernelci libraries.

src/fstests/runner.py
```diff
@@ -0,0 +1,59 @@
...
+        self._db = kernelci.db.get_db(self._db_config, api_token)
+
+    def run(self):
+        sub_id = self._db.subscribe_node_channel({
+            'op': 'created',
+            'name': 'tarball',
+        })
+        print('Listening for tarballs')
+        print('Press Ctrl-C to stop.')
+        tarball_node = None
```
JenySadadia, line 31: This line can be dropped.
```suggestion
```

gctucker:
> > 
> 
> Just to clarify, I know that I need to connect to db in API. But I can update this in some next PR, because I am using subprocess and I will start running the commands from host (`kernelci2`) that I believe won´t have the kernelci libraries.

The dependencies have to be installed on `kernelci2`.  You can do that with `pip3 install -r requirements.txt` in `kernelci-core`.  I guess that's something to add to the `README.md`, I'll do that.

PR #102, opened 2022-08-12
Fstests subprocess kvm xfstests

alexandrasp:
- Add a method to check if kvm-xfstests is available in the client environment.
- Include kernelci library as a requirement.

### DIFF
--- a/src/fstests/runner.py
+++ b/src/fstests/runner.py
@@ -7,6 +7,7 @@
 
 import os
 import sys
+import subprocess
 
 import kernelci
 import kernelci.config
@@ -20,6 +21,18 @@ class FstestsRunner:
         self._db_config = configs['db_configs'][args.db_config]
         api_token = os.getenv('API_TOKEN')
         self._db = kernelci.db.get_db(self._db_config, api_token)
+
+    def check_kvm_xfstest_env(self):
+        try:
+            result = subprocess.run("kvm-xfstests --help", check=True, shell=True)
+            if result.returncode == 0:
+                print(f"KVM-xftests found.")
+                return True
+            else:
+                return False
+        except Exception as e:
+            print(f"Raised exception: {e}")
+            return False
 
     def run(self):
         sub_id = self._db.subscribe_node_channel({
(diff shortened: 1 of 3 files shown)


alexandrasp:
Now, after installing the dependency (`kernelci`), we will be able to run `run` and `check` commands from the client side using:

`python3 src/fstests/runner.py --settings=config/kernelci.conf run --db-config=staging.kernelci.org`
`python3 src/fstests/runner.py --settings=config/kernelci.conf check --db-config=staging.kernelci.org`

src/fstests/runner.py
```diff
@@ -41,6 +50,13 @@ def run(self):
             self._db.unsubscribe(sub_id)
             return True
 
+class cmd_check(Command):
+    help = 'Check KVM-xfstests environment'
+    args = [Args.db_config]
+    opt_args = [Args.verbose]
+
+    def __call__(self, configs, args):
+        return FstestsRunner(configs, args).check_kvm_xfstest_env()
```
gctucker, line 59: The `check_kvm_xfstests_env()` method is not returning anything, so please make it return `True` or `False` when there was an error.

src/fstests/runner.py
```diff
@@ -21,6 +22,14 @@ def __init__(self, configs, args):
         api_token = os.getenv('API_TOKEN')
         self._db = kernelci.db.get_db(self._db_config, api_token)
 
+    def check_kvm_xfstest_env(self):
+        try:
+            result = subprocess.run(["kvm-xfstests --help"], check=True, shell=True)
```
gctucker, line 27: With `shell=True` you should be able to run this without a list:
```suggestion
            result = subprocess.run("kvm-xfstests --help", check=True, shell=True)
```

src/fstests/runner.py
alexandrasp: good point, I forgot to include this during the rebase.

src/fstests/runner.py
alexandrasp: good as well! I started this using call, but when I went to python 3.6 doc I saw the explicitly recommendation to use `run`, and probably I missed this detail about the list won't be more necessary.

src/fstests/runner.py
```diff
@@ -21,6 +22,17 @@ def __init__(self, configs, args):
...
+    def check_kvm_xfstest_env(self):
+        try:
+            result = subprocess.run("kvm-xfstests --help", check=True, shell=True)
+            if result.returncode == 0:
+                print(f"KVM-xftests found.")
+                return True
+            else:
+                return False
+        except Exception as e:
+            print(f"Raised exception: {e}")
```
gctucker, line 34: `return False` here?

gctucker (approved):

PR #104, opened 2022-08-13
src: dump call stack when catching unexpected exception

gctucker:
When an unexpected exception occurs during a loop, catch it and log
the call stack as when it's not handled rather than silencing it with
the `finally` clause.  This helps with debugging to find where errors
are coming from.

### DIFF
--- a/src/notifier.py
+++ b/src/notifier.py
@@ -11,6 +11,7 @@ import json
 import logging
 import os
 import sys
+import traceback
 
 import kernelci
 import kernelci.config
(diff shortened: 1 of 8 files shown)


src/notifier.py
```diff
@@ -74,6 +75,8 @@ def __call__(self, configs, args):
                 sys.stdout.flush()
         except KeyboardInterrupt as e:
             self._logger.log_message(logging.INFO, "Stopping.")
+        except Exception as e:
```
JenySadadia, line 78: The exception object is not being used. Hence, can be dropped.
```suggestion
        except Exception:
```

src/runner.py
```diff
@@ -132,6 +133,8 @@ def loop(self):
                 self._cleanup_paths()
         except KeyboardInterrupt as e:
             self._logger.log_message(logging.INFO, "Stopping.")
+        except Exception as e:
```
JenySadadia, line 136: The same comment as above for other services.

JenySadadia (changes requested):

src/runner.py
mgalka: The same in line 133/134

JenySadadia (approved):

PR #107, opened 2022-08-16
Add kvm xfstest conf and template

alexandrasp:
- Add a jinja2 template for fstests.
- Add a fstests test plan.
- Update the run function to run to generate a job and run it using shell lab.
- Drop the code used to run `kvm-xfstests` as a subprocess.

### DIFF
--- a/config/kernelci.conf
+++ b/config/kernelci.conf
@@ -18,6 +18,9 @@ ssh_host: 172.17.0.1
 [runner]
 output: /home/kernelci/output
 
+[fstests_runner]
+output: /home/kernelci/alex/fstests-poc/output
+
 [notifier]
 
 [kcidb]
--- a/config/runtime/fstests.jinja2
+++ b/config/runtime/fstests.jinja2
@@ -1,0 +1,32 @@
+{# -*- mode: Python -*- -#}
+{# SPDX-License-Identifier: LGPL-2.1-or-later -#}
+ 
+{%- extends 'base/python.jinja2' %}
+ 
+{%- block python_imports %}
+{{ super() }}
+import subprocess
+{%- endblock %}
+ 
+{%- block python_globals %}
+{{ super() }}
+REVISION = {{ revision }}
+{% endblock %}
+ 
+{%- block python_body %}
+def check_kvm_xfstest_env():
+   try:
+       result = subprocess.run("kvm-xfstests --help", check=True, shell=True)
+       if result.returncode == 0:
+           print(f"KVM-xftests found.")
+           return True
+       else:
+           return False
+   except Exception as e:
+       print(f"Raised exception: {e}")
+       return False
+ 
+def main(argv):
+   print("Checking if KVM-xfstests is available...")
+   check_kvm_xfstest_env()
+{% endblock %}
--- a/src/fstests/runner.py
+++ b/src/fstests/runner.py
@@ -8,6 +8,7 @@
 import os
 import sys
 import subprocess
+import tempfile
 
 import kernelci
 import kernelci.config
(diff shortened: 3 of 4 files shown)


alexandrasp:
Tested locally. 
How I tested it: 
1. Run the command below and start to listen for `tarballs`.
`python3 src/fstests/runner.py --settings config/kernelci.conf --yaml-config kernelci/pipeline.yaml run --db-config docker-host --lab-config shell --output test_output --target shell --plan fstests`
2. Using my local environment "force" a tarball, it triggers the event, and then the job is generated and run.
3. Result:
```
Listening for tarballs
Press Ctrl-C to stop.
Node tarball with id: 62fbc34f811733825c6cd86b                    from revision: babf0bb978e3
Waiting...
Usage: kvm-xfstests [<OPTIONS>] smoke|full
Usage: kvm-xfstests [<OPTIONS>] <test> ...
Usage: kvm-xfstests [<OPTIONS>] -g <group> ...
Usage: kvm-xfstests [<OPTIONS>] shell|maint
Usage: kvm-xfstests [<OPTIONS>] syz <repro>

Common options are:
	-a		- Disable auto-exclude; run all tests
	-c config	- Specify a file system configuration
	-C count	- Run the specified tests multiple times
	-I image	- Use this test appliance image
	-m mountopts	- Append mount options to fs config
	-n nr_cpus	- Specify the number of cpu's
	-numa num	- Ask KVM to create <num> NUMA nodes
	-N		- Enable networking (requires root)
	-o opts		- Extra kernel command line options
	-O opts		- Extra options for test runner
	-r ram		- Specify memory to be used in megabytes
	-x group	- Exclude group of tests from running
	-X test		- Exclude test from running
	--kernel file	- Boot the specified kernel
	--initrd initrd	- Boot with the specified initrd
	--no-log	- Don't save the log file for this run
	--no-action	- Print the command to start the VM

Common file system configurations are:
	4k 1k ext3 nojournal ext3conv metacsum dioread_nolock 
	data_journal bigalloc bigalloc_1k inline

xfstest names have the form: ext4/NNN generic/NNN shared/NNN

...done
```

gctucker:
To test this with an actual tarball, you can also just run the services locally:
```
docker-compose up trigger tarball notifier
```
while your script is running and waiting for events.  Then I think you could use the same approach as in `src/runner.py` with the `--node-id` command line argument to re-run the tests with a given tarball node id rather than waiting for an event.

src/fstests/runner.py
```diff
@@ -21,20 +21,48 @@ def __init__(self, configs, args):
...
-                print(f"KVM-xftests found.")
-                return True
-            else:
-                return False
-        except Exception as e:
-            print(f"Raised exception: {e}")
-            return False
+            tmp = tempfile.TemporaryDirectory(dir=self._output)
+            job = self._schedule_job(tarball_node, plan, device, tmp.name)
+            if self._runtime.config.lab_type == 'shell':
```
gctucker, line 56: This and other things can be simplified because it will only be running in `shell`.  Having a separate service from the regular `src/runner.py` means we can have the code tailored to what is really needed for fstests and keep it shorter.

src/fstests/runner.py
```diff
@@ -46,6 +74,9 @@ def run(self):
                 tarball_node = self._db.receive_node(sub_id)
                 print(f"Node tarball with id: {tarball_node['_id']}\
                     from revision: {tarball_node['revision']['commit'][:12]}")
+                plan = self._plan_configs[args.plan]
```
gctucker, line 77: We don't need `args.plan` as this service is only for running `fstests`.

src/fstests/runner.py
```diff
@@ -54,21 +85,17 @@ def run(self):
             self._db.unsubscribe(sub_id)
             return True
 
-class cmd_check(Command):
-    help = 'Check KVM-xfstests environment'
-    args = [Args.db_config]
-    opt_args = [Args.verbose]
-
-    def __call__(self, configs, args):
-        return FstestsRunner(configs, args).check_kvm_xfstest_env()
```
gctucker, line 63: Maybe we can actually keep this, so in the future if we want to run it again on a different host or if we're redeploying the VM we can use it as a simple sanity check to confirm the `kvm-xfstests` tool is present.

src/fstests/runner.py
```diff
@@ -54,21 +85,17 @@ def run(self):
...
 
 class cmd_run(Command):
     help = 'KVM fstests runner'
-    args = [Args.db_config]
+    args = [
+        Args.db_config, Args.lab_config, Args.output,
+        Args.plan, Args.target,
```
gctucker, line 93: We don't need `lab_config`, `plan` and `target`.  This service is specific to `fstests`.

config/kernelci.conf
```diff
@@ -18,6 +18,9 @@ ssh_host: 172.17.0.1
 [runner]
 output: /home/kernelci/output
 
+[fstests_runner]
+output: /home/kernelci/fstests/output
```
gctucker, line 22: Please note that the `/home/kernelci/output` path was based on the volumes available in containers run with docker-compose.  So in this case, maybe it should be something like `/home/kernelci/alex/fstests-poc/output` instead.  Also maybe it would be worth having a separate config file in `src/fstests/kernelci.conf` for this.

config/runtime/fstests.jinja2
```diff
@@ -0,0 +1,8 @@
+{# -*- mode: Python -*- -#}
+{# SPDX-License-Identifier: LGPL-2.1-or-later -#}
+
+{%- extends 'base/shell.jinja2' %}
+
+{%- block commands %}
+kvm-xfstests --help
+{% endblock %}
```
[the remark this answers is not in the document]
gctucker: After looking into the next steps here, I think it's really necessary to move to Python as there aren't tools available yet for submitting the results on the command line.  So this is probably fine for this PR but it will need to be removed in the next one when generating the kernel config file.

src/fstests/runner.py
alexandrasp: alright!

alexandrasp:
@gctucker I updated some details, removed the parameters, and just set them into the code, I am not sure where we could simplify more in order to have the parameters to use methods from the kernelci library (`lab.save`, `lab.define`, `lab.submit`)

alexandrasp:
> Reference

Thanks for this! It was useful and I am now using staging tarball in tests as well. :)

config/kernelci.conf
alexandrasp: updated because it makes totally sense.

config/runtime/fstests.jinja2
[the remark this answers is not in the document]
alexandrasp: Ok, I just drop the commit..then I can add a commit to drop in my next PR.

src/fstests/runner.py
alexandrasp: ok, all set with constant/hardcoded..

config/runtime/fstests.jinja2
```diff
@@ -0,0 +1,32 @@
...
+   try:
+       result = subprocess.run("kvm-xfstests --help", check=True, shell=True)
+       if result.returncode == 0:
+           print(f"KVM-xftests found.")
+           return True
+       else:
+           return False
+   except Exception as e:
+       print(f"Raised exception: {e}")
+       return False
```
gctucker, line 27: This will need to be refactored with the new `Job` class, but it can be done in the next PR.

gctucker (approved):
OK looks good.  With the next PR there should also be a node created for the `fstest` job so we'll see in the staging logs that it's running as expected (create a node before running the job, then the job will update it with the result set to `pass` if it worked, no actual test cases yet).

PR #110, opened 2022-08-18
config/runtime: update kver to use Job() class

gctucker:
Update the kver test plan template to use the new Job() class.

Depends on https://github.com/kernelci/kernelci-core/pull/1368


alexandrasp (approved):
LGTM

PR #111, opened 2022-08-19
src/runner.py: also print stack in run command

gctucker:
Also print the stack trace when catching an expected exception with
the 'run' command, not just 'loop'.  The 'run' command is mostly used
for debugging and development so having details about exceptions is
very useful.

Fixes: 75b9c16b8f26 ("src: dump call stack when catching unexpected exception")


alexandrasp (approved):
LGTM

JenySadadia (approved):

PR #112, opened 2022-08-19
Add group and path attributes

gctucker:
Add the `Node.group` and `Node.path` attributes when submitting nodes where applicable.

Depends on https://github.com/kernelci/kernelci-api/pull/167 and https://github.com/kernelci/kernelci-api/pull/168

### DIFF
--- a/config/runtime/kunit.jinja2
+++ b/config/runtime/kunit.jinja2
@@ -67,10 +67,12 @@ cd {src_path} && \
             'parent': parent['_id'],
             'status': 'complete',
             'revision': parent['revision'],
+            'group': parent['group'],
         }
         for node in child_nodes:
             base_node.update({
                 'name': node['name'],
+                'path': parent['path'] + [node['name']],
                 'result': node.get('result'),
             })
             sent_node = db.submit({'node': base_node})[0]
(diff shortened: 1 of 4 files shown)


config/runtime/kunit.jinja2
```diff
@@ -67,10 +67,12 @@ cd {src_path} && \
             'parent': parent['_id'],
             'status': 'complete',
             'revision': parent['revision'],
+            'group': parent['group'],
         }
         for node in child_nodes:
             base_node.update({
                 'name': node['name'],
+                'path': parent['path'] + [node['name']],
                 'result': node.get('result'),
```
JenySadadia, line 76: How do we skip the particular test and why?
Need to add logic to set node result to `SKIP`.

gctucker: Err, sorry?  I don't see what you mean.  This PR is about adding the `path` attribute, how is that related to the node result?

gctucker: OK I see.  The "SKIP" result from KUnit is a key in the `RESULT_MAP` dictionary so any test that KUnit has skipped will have a result set to `None` in the data structure.

gctucker:
@JenySadadia Does this look OK to you now?

JenySadadia:
> @JenySadadia Does this look OK to you now?

Approved.

JenySadadia (approved):

PR #113, opened 2022-08-19
config/runtime: set result string in kver.jinja2

gctucker:
Rather than passing True or False, pass the actual result string to
set in the node in the kver test template.

### DIFF
--- a/config/runtime/kver.jinja2
+++ b/config/runtime/kver.jinja2
@@ -64,7 +64,8 @@ class Job(BaseJob):
         with open(os.path.join(src_path, 'Makefile')) as makefile:
             kver = REVISION['version']
             res = self._check_kver(makefile, kver)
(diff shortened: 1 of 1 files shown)


alexandrasp (approved):
LGTM

config/runtime/kver.jinja2
```diff
@@ -64,7 +64,8 @@ class Job(BaseJob):
         with open(os.path.join(src_path, 'Makefile')) as makefile:
             kver = REVISION['version']
             res = self._check_kver(makefile, kver)
-            print("Result: {}".format("PASS" if res is True else "FAIL"))
+            result = 'pass' if res is True else 'fail'
+            print("Result: {}".format(result))
```
JenySadadia, line 68: I'd suggest using f-string.
```suggestion
            print(f"Result: {result}")
```

PR #117, opened 2022-08-22
Clean up email report script

gctucker:
Simplify the `email_report.py` code in preparation for adding support for results from test groups:

* Remove unnecessary exception handling
* Use `_` for all private methods
* Use object attributes when applicable
* Always print the email content on stdout

### DIFF
--- a/src/test_report.py
+++ b/src/test_report.py
@@ -55,46 +55,20 @@ class TestReport:
         self._logger.log_message(logging.INFO, email_content)
         return email_msg
 
-    def _smtp_connect(self, email_user, email_password):
+    def _smtp_connect(self):
         try:
             if self._smtp_port == 465:
-                email_server = smtplib.SMTP_SSL(self._smtp_host,
-                                                self._smtp_port)
+                smtp = smtplib.SMTP_SSL(self._smtp_host, self._smtp_port)
             else:
-                email_server = smtplib.SMTP(self._smtp_host,
-                                            self._smtp_port)
-                email_server.starttls()
-            email_server.login(email_user, email_password)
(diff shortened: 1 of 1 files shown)


src/test_report.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, configs, args):
         self._email_user = os.getenv('EMAIL_USER')
         self._email_pass = os.getenv('EMAIL_PASSWORD')
 
-    def create_email(self, email_content, send_to, send_from, email_subject):
+    def _create_email(self, send_to, send_from, email_content, email_subject):
```
mgalka, line 43: It seems a bit more intuitive to me to have `email_subject` before `email_content`. It's nothing big and I may be biased. So it's rather "for consideration" than something crucial. I wouldn't probably even put a remark here if you haven't been changing fucntion's signature.

mgalka (commented):
minor comments

src/test_report.py
gctucker: Yes, my main reason for this change was to get the send to/from together and the email subject/content together as well.  But that's a fair point.

src/test_report.py
```diff
@@ -131,19 +105,19 @@ def run(self):
...
-                                              self._email_send_to,
-                                              self._email_send_from,
-                                              email_sub)
-                email_server = self.smtp_connect(self._email_user,
-                                                 self._email_pass)
-                if email_server:
-                    self.send_mail(email_msg, email_server)
-                    email_server.quit()
+                print(content)
+                sys.stdout.flush()
```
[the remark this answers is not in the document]
gctucker: Nice, sometimes I should read the docs to learn what happened after Python 2.6.

PR #122, opened 2022-08-25
config/runtime: add a step to run tests

alexandrasp:
Add in fstests.jinja2 template a step to run smoke tests using kvm-xfstests.


alexandrasp:
Tested locally, using this we can see tests at the stdout logs directly. The script is already running on the client side, so if we create a new tarball on the API side, the client side will run fstests using kvm-xfstests and send the stdout to the log file.

gctucker:
Please mention fstests in the commit subject.  Also, like I said before it would be neater if all the fstests files were in the src/fstests directory.

gctucker:
Could you please include the summary from the log you're getting when the test is running on the server?

Ultimately this should be sent as an artifact but for now we can just rely on that to check it's looking fine.  Then when we parse results from the XML file we'll have more data to confirm it's producing valid results.

alexandrasp:
> Could you please include the summary from the log you're getting when the test is running on the server?
> 
> Ultimately this should be sent as an artifact but for now we can just rely on that to check it's looking fine. Then when we parse results from the XML file we'll have more data to confirm it's producing valid results.

OK, after rebasing here, I updated the server and run the tests.

The file with all tests output could be found `/home/kernelci/alex/fstests-poc/kernelci-pipeline/logs` directory, file `log-fstests.txt`, to test it I just used a `node-id`, but when I finished I restarted the script to keep listening for new tarballs.

```               total        used        free      shared  buff/cache   available
Mem:            1977          60        1866           0          50        1884
Swap:              0           0           0
END TEST: Ext4 4k block Thu Aug 25 20:39:50 UTC 2022
-------------------- Summary report
KERNEL:    kernel 6.0.0-rc2-xfstests #1 SMP PREEMPT_DYNAMIC Thu Aug 25 20:17:20 UTC 2022 x86_64
CMDLINE:   smoke
CPUS:      2
MEM:       1977.51

ext4/4k: 406 tests, 2 failures, 17 skipped, 1209 seconds
  Failures: generic/634 generic/635 
Totals: 389 tests, 17 skipped, 2 failures, 0 errors, 1198s
```

src/fstests/fstests.jinja2
```diff
@@ -36,6 +36,16 @@ class Job(BaseJob):
         except Exception as e:
             print('Kernel build error:', e)
 
+    def _run_smoke_tests(self):
+        try:
+            result = subprocess.run("kvm-xfstests smoke", check=True, shell=True)
```
mgalka, line 41: In general it's better to pass command and arguments as a sequence if you pass it as a string it's interpretation becomes platform dependant.
```suggestion
            result = subprocess.run(["kvm-xfstests", "smoke"], check=True, shell=True)
```

mgalka: Do we need to run the command in the shell (shell=True)?

src/fstests/fstests.jinja2
```diff
@@ -36,6 +36,16 @@ class Job(BaseJob):
...
+            if result.returncode == 0:
+                print(f"Tests run with success.")
```
mgalka, line 43: I'd rather come up with something like "Test run success.", but that's just a suggestion.

mgalka (commented):
Some remarks

src/fstests/fstests.jinja2
alexandrasp: no, we dont. We just need to build the kernel...

src/fstests/fstests.jinja2
alexandrasp: updated.

src/fstests/runner.py
```diff
@@ -54,7 +54,7 @@ def _schedule_job(self, tarball_node, device_config, tmp):
             }
             params.update(self._plan.params)
             params.update(device_config.params)
-            templates = ['config/runtime',
+            templates = ['src/fstests',
```
gctucker, line 57: I suspect we have to keep `config/runtime` and just add a line for `src/fstests`, otherwise any base template in `config/runtime` won't be found.  In practice, at the moment all the base templates are in `/etc/kernelci/runtime` so it works, but still there's no good reason for removing `config/runtime` imho.

config/kernelci.conf
```diff
@@ -19,7 +19,7 @@ ssh_host: 172.17.0.1
 output: /home/kernelci/output
 
 [fstests_runner]
-output: /home/kernelci/alex/fstests-poc/output
+output: /home/alexandra/fstests-poc/output
```
gctucker: This gets undone in the next commit, so please just drop it.  It probably was added here by mistake from a local change on your laptop.

config/runtime/fstests.jinja2
```diff
@@ -36,6 +36,16 @@ class Job(BaseJob):
         except Exception as e:
             print('Kernel build error:', e)
 
+    def _run_smoke_tests(self):
+        try:
+            result = subprocess.run(["kvm-xfstests", "smoke"], check=True)
```
gctucker, line 41: As a follow-up, the other `subprocess.run()` calls should be cleaned up in the same way.

gctucker (commented):
Some minor things to clean up

config/kernelci.conf
gctucker: Still not resolved...  It's a trivial thing and it's really adding unnecessary noise in the git diff, which can lead to unnecessary conflicts when rebasing etc.

src/fstests/fstests.jinja2
```diff
@@ -13,7 +13,7 @@ class Job(BaseJob):
     def _kernel_config(self, kdir):
         try:
             os.chdir(kdir)
-            result = subprocess.run("kvm-xfstests install-kconfig", check=True, shell=True)
+            result = subprocess.run(["kvm-xfstests", "install-kconfig"], check=True, shell=True)
```
gctucker, line 16: I suggested to do this as a _follow-up_ which means not in this commit or this PR as it's not related to running the tests...  Adding other random changes makes the code harder to maintain and reviews harder to do.

gctucker (changes requested):

src/fstests/runner.py
```diff
@@ -55,7 +55,8 @@ def _schedule_job(self, tarball_node, device_config, tmp):
             params.update(self._plan.params)
             params.update(device_config.params)
             templates = ['config/runtime',
-                         '/etc/kernelci/runtime']
+                         '/etc/kernelci/runtime',
+                         'src/fstests']
```
gctucker, line 59: This is meant to go in the 2nd commit as in the last iteration...

Please, make sure you double check your commits individually before pushing and also be more careful during the rebase process.

gctucker (changes requested):
Close but no cigar.

src/fstests/runner.py
alexandrasp: Ok, Sorry for this mistake. I checked this carefully (double check)  and verified after pushing to github again that each commit represents what is said in the message.

Issue #119, opened 2022-08-24
Add step to run smoke tests


gctucker:
Once the step to build a kernel image is complete, the next part is to run `kvm-xfstests` with the `smoke` set of tests.  This is a relatively small set which can take around 15min to complete on a decent machine.  It's a good compromise to have some useful coverage without running the full set of tests which would take many hours.

During development, it's also possible to hard-code a couple of test names to get results much quicker rather than the whole `smoke` set.

- [x] Step added to add another call `kvm-xfstests` to run the `smoke` tests
- [x] Verified manually in the log output that the tests were run as expected
- [x] Overall status reported to the API, still with a single `fstests` node

alexandrasp:
- I am investigating how to access the VM and get the result.xml during the run time. 
- Manually I am already able to see the tests running, so when the tests finish I can join the VM and see the results file.

gctucker:
I don't think it's necessary to get the XML file out of the VM, checking the output is enough for now to know the tests were run as expected.  We'll see the details of the logs when parsing the XML results.

PR #123, opened 2022-08-31
src/fstests: parse fstests results

alexandrasp:
- add a step to get fstests results.xml file and copy it to the kernel source directory.
- parse fstests XML result file into a JSON containing the node structure to be submitted.
- small design improviments

### DIFF
--- a/src/fstests/fstests.jinja2
+++ b/src/fstests/fstests.jinja2
@@ -126,7 +126,11 @@ class Job(BaseJob):
             return 'fail'
         if not self._get_xml_from_vm(src_path):
             return 'fail'
-        with open(os.path.join(src_path, 'results.json'), 'w') as result_file:
-            result_file.write(json.dumps(self._parse_results(src_path)))
+        try:
+            with open(os.path.join(src_path, 'results.json'), 'w') as result_file:
+                result_file.write(json.dumps(self._parse_results(src_path)))
+        except Exception as e:
+            print(f"Exception raised while parsing results: {e}")
+            return 'fail'
         return 'pass'
 {% endblock %}
--- a/src/fstests/fstests.jinja2
+++ b/src/fstests/fstests.jinja2
@@ -41,6 +41,16 @@ class Job(BaseJob):
         except Exception as e:
             print('Kernel build error:', e)
 
+    def _get_xml_from_vm(self, kdir):
+        try:
+            result = subprocess.run([f"debugfs -R \"dump ext4/results-4k/results.xml {kdir}/results.xml\" {xfstests_bld_path}/run-fstests/disks/vdg"], check=True, shell=True)
+            if result.returncode == 0:
+                print(f"XML file in the output directory.")
+                return 'pass'
+        except Exception as e:
+            print('XML not found.', e)
+            return 'fail'
+
     def _run_smoke_tests(self):
         try:
             result = subprocess.run(["kvm-xfstests", "smoke"], check=True)
@@ -68,7 +78,7 @@ class Job(BaseJob):
         if self._check_kvm_xfstest_env():
             if self._kernel_config(src_path) == 'pass':
                 if self._kernel_build(src_path) == 'pass':
-                    return self._run_smoke_tests()
-        else:
-            return 'fail'
+                    if self._run_smoke_tests() == 'pass':
+                        return self._get_xml_from_vm(src_path)
+        return 'fail'
 {% endblock %}
--- a/src/fstests/fstests.jinja2
+++ b/src/fstests/fstests.jinja2
@@ -10,7 +10,9 @@ xfstests_bld_path = '{{ xfstests_bld_path }}'
 
 {%- block python_imports %}
 {{ super() }}
+import json
 import subprocess
+import xml.etree.ElementTree as ET
 {%- endblock %}
 
 {% block python_job -%}
@@ -51,6 +53,40 @@ class Job(BaseJob):
             print('XML not found.', e)
             return 'fail'
 
+    def _parse_to_json(self, node, tree, name=None):
+        if 'result' in tree:
+            node['name'] = name
+            node['result'] = tree['result']
+        else:
+            if (not name == None): node['name'] = name
+            child_nodes = []
+            for name in tree:
+                child_nodes.append(self._parse_to_json({}, tree[name], name))
+            node['child_nodes'] = child_nodes
+        return node
+
+    def _parse_results(self, kdir):
+        xml_file = ET.parse(f"{kdir}/results.xml")
+        root = xml_file.getroot()
+        tree = {}
+        for test_case in root.iter('testcase'):
+            if test_case.find('skipped') is not None:
+                result = None
+            elif test_case.find('failure') is not None:
+                result = 'fail'
+            else:
+                result = 'pass'
+            name = test_case.attrib['name'].split('/')
+            if not name[0] == 'ext4':
+                name.insert(0, 'ext4')
+            current = tree
+            for item in name:
+                if not item in current:
+                    current[item] = {}
+                current = current[item]
+            current['result'] = result
+        return self._parse_to_json({}, tree, 'fstests')
+
     def _run_smoke_tests(self):
         try:
             result = subprocess.run(["kvm-xfstests", "smoke"], check=True)
@@ -79,6 +115,8 @@ class Job(BaseJob):
             if self._kernel_config(src_path) == 'pass':
                 if self._kernel_build(src_path) == 'pass':
                     if self._run_smoke_tests() == 'pass':
-                        return self._get_xml_from_vm(src_path)
+                        if self._get_xml_from_vm(src_path) == 'pass':
+                            with open(os.path.join(src_path, 'results.json'), 'w') as result_file:
+                                result_file.write(json.dumps(self._parse_results(src_path)))
         return 'fail'
 {% endblock %}
--- a/src/fstests/fstests.jinja2
+++ b/src/fstests/fstests.jinja2
@@ -53,20 +53,8 @@ class Job(BaseJob):
             print('XML not found.', e)
             return False
 
-    def _parse_to_json(self, node, tree, name=None):
-        if 'result' in tree:
-            node['name'] = name
-            node['result'] = tree['result']
-        else:
-            if (not name == None): node['name'] = name
-            child_nodes = []
-            for name in tree:
-                child_nodes.append(self._parse_to_json({}, tree[name], name))
-            node['child_nodes'] = child_nodes
-        return node
-
-    def _parse_results(self, kdir):
-        xml_file = ET.parse(f"{kdir}/results.xml")
+    def _parse_xml_results(self, kdir):
+        xml_file = ET.parse(kdir)
         root = xml_file.getroot()
         tree = {}
         for test_case in root.iter('testcase'):
@@ -85,7 +73,7 @@ class Job(BaseJob):
                     current[item] = {}
                 current = current[item]
             current['result'] = result
-        return self._parse_to_json({}, tree, 'fstests')
+        return tree
 
     def _run_smoke_tests(self):
         try:
@@ -96,6 +84,26 @@ class Job(BaseJob):
         except Exception as e:
             print('Problem found during tests:', e)
             return False
+
+    def _convert_results_for_api(self, tree, name):
+        if 'result' in tree:
+            return {
+                'name': name,
+                'result': tree['result'],
+            }
+        else:
+            child_nodes = []
+            for child_name in tree:
+                child_nodes.append(self._convert_results_for_api(tree[child_name], child_name))
+            return {
+                'name': name,
+                'child_nodes': child_nodes,
+            }
+
+    def _parse_results(self, kdir):
+        tree_data = self._parse_xml_results(os.path.join(kdir, 'results.xml'))
+        api_data = self._convert_results_for_api(tree_data, 'fstests')
+        return api_data
 
     def _check_kvm_xfstest_env(self):
         try:
--- a/src/fstests/fstests.jinja2
+++ b/src/fstests/fstests.jinja2
@@ -23,13 +23,13 @@ class Job(BaseJob):
             result = subprocess.run("kvm-xfstests install-kconfig", check=True, shell=True)
             if result.returncode == 0:
                 print(f"Kernel config done.")
-                return 'pass'
+                return True
             else:
                 print((f"Missing kernel .config file."))
-                return 'fail'
+                return False
         except Exception as e:
             print('Kernel config error:', e)
-            return 'fail'
+            return False
 
     def _kernel_build(self, kdir):
         try:
@@ -37,9 +37,9 @@ class Job(BaseJob):
             result = subprocess.run("make -j$(nproc)", check=True, shell=True)
             if result.returncode == 0:
                 print(f"Kernel Build DONE.")
-                return 'pass'
+                return True
             else:
-                return 'fail'
+                return False
         except Exception as e:
             print('Kernel build error:', e)
 
@@ -48,10 +48,10 @@ class Job(BaseJob):
             result = subprocess.run([f"debugfs -R \"dump ext4/results-4k/results.xml {kdir}/results.xml\" {xfstests_bld_path}/run-fstests/disks/vdg"], check=True, shell=True)
             if result.returncode == 0:
                 print(f"XML file in the output directory.")
-                return 'pass'
+                return True
         except Exception as e:
             print('XML not found.', e)
-            return 'fail'
+            return False
 
     def _parse_to_json(self, node, tree, name=None):
         if 'result' in tree:
@@ -92,10 +92,10 @@ class Job(BaseJob):
             result = subprocess.run(["kvm-xfstests", "smoke"], check=True)
             if result.returncode == 0:
                 print(f"Test run success.")
-                return 'pass'
+                return True
         except Exception as e:
             print('Problem found during tests:', e)
-            return 'fail'
+            return False
 
     def _check_kvm_xfstest_env(self):
         try:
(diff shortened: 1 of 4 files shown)


alexandrasp:
depending on #122. I opened a PR and put an instance to run on `kernelci2`. I know there are some hacks here, but we need to check better how we could handle directories and other details, or if we keep it like this for now(POC).

src/fstests/fstests.jinja2
```diff
@@ -63,7 +103,32 @@ class Job(BaseJob):
...
+            base_node.update({
+                'name': node['name'],
+                'result': node['result'],
+            })
+            sent_node = db.submit({'node': base_node})[0]
+
+    def _submit(self, results, node_id, db):
+        node = super()._submit(results[0]['result'], node_id, db)
+        self._submit_fstests(node, results, db)
+        return node
```
gctucker, line 133: I think you can drop that and just focus on parsing the results in the current PR.  Otherwise, the changes to submit the results are going to delay when we can merge the code to parse the results.  That's why there are 2 issues in the backlog, first to parse the results and then to submit them to the API.

src/fstests/fstests.jinja2
```diff
@@ -5,9 +5,17 @@
...
 {{ super() }}
+import json
 import subprocess
+import xml.etree.ElementTree as ET
 {%- endblock %}
 
+{%- block python_globals %}
+{{ super() }}
+FSTESTS_ROOT = '/home/kernelci/alex/fstests-poc/kernelci-pipeline/src/fstests'
+VM_ROOT = '/home/kernelci/alex/fstests-poc/fstests/'
```
gctucker, line 16: I don't think we need these absolute paths (and we can't use it anyway as it's only valid on the `kernelci2` VM).  How about storing the JSON file in the kernel source tree like the KUnit template is doing?

src/fstests/fstests.jinja2
```diff
@@ -46,6 +54,38 @@ class Job(BaseJob):
...
+            name = child.attrib['name']
+            testcase = {}
+            testcase['name'] = name.replace('/', '.')
+            if child.find('skipped') != None:
+                testcase['result'] = None
+            elif child.find('failure') != None:
+                testcase['result'] = 'fail'
+            else:
+                testcase['result'] = 'pass'
+            results.append(testcase)
```
gctucker, line 83: There are a few issues here, it's not creating sub-groups and the test names shouldn't have a `.` in them.  Instead, the name should be split to have the path elements.

We can discuss this in more detail outside of GitHub.

gctucker (changes requested):

src/fstests/fstests.jinja2
alexandrasp: Sounds like a good idea

src/fstests/fstests.jinja2
alexandrasp: ok, updated accordingly. It makes totally sense, 1 commit = 1 significant change.

gctucker:
Please rebase to resolve the conflicts now that #119 has been merged.

src/fstests/fstests.jinja2
```diff
@@ -36,6 +36,16 @@ class Job(BaseJob):
         except Exception as e:
             print('Kernel build error:', e)
 
+    def _get_xml_from_vm(self, kdir):
+        try:
+            result = subprocess.run([f"debugfs -R \"dump ext4/results-4k/results.xml {kdir}/results.xml\" /home/kernelci/alex/fstests-poc/fstests/disks/vdg"], check=True, shell=True)
```
alexandrasp, line 43: @gctucker I got what you meant about kernel source tree for results, but in the case of VM files, do you have any other suggestion?

alexandrasp:
I hope this finally is according to the expected, the new JSON file output will be like this:  

```{
    "name": "FSTest Parsing Results",
    "child_nodes": [
        {
            "name": "ext4",
            "child_nodes": [
                {
                    "name": "001",
                    "result": "pass"
                },
                {
                    "name": "002",
                    "result": null
                },
                {
                    "name": "003",
                    "result": "pass"
                },
                {
                    "name": "005",
                    "result": "pass"
                },
                {
                    "name": "generic",
                    "child_nodes": [
                        {
                            "name": "001",
                            "result": "pass"
                        },
                        {
                            "name": "002",
                            "result": "pass"
                        },
                        {
                            "name": "003",
                            "result": "pass"
                        },
                        {
                            "name": "004",
                            "result": "pass"
                        },
                        {
                            "name": "005",
                            "result": "pass"
                        }
                    ]
                },
                {
                    "name": "shared",
                    "child_nodes": [
                        {
                            "name": "002",
                            "result": null
                        },
                        {
                            "name": "032",
                            "result": null
                        }
                    ]
                }
            ]
        }
    ]
}```

File: src/fstests/fstests.jinja2 [hunk 5]
```diff
@@ -63,7 +73,7 @@ class Job(BaseJob):
         if self._check_kvm_xfstest_env():
             if self._kernel_config(src_path) == 'pass':
                 if self._kernel_build(src_path) == 'pass':
-                    return self._run_smoke_tests()
-        else:
-            return 'fail'
+                    if self._run_smoke_tests() == 'pass':
+                        return self._get_xml_from_vm(src_path)
+        return 'fail'
```
gctucker, line 116: We'll have to stop nesting these `if` statements at some point or it'll become very hard to read and maintain.

src/fstests/fstests.jinja2
gctucker: `/home/kernelci/alex/fstests-poc/fstests/disks/vdg`

Surely we can use a relative path for that?  Also, I don't see what you mean.  Can you not copy the xml file to `kdir` which is the kernel source tree?

gctucker: As discussed, one option is to add a command line argument e.g. `--xfstests-bld-path` and store it in the Job class, then use it in the string here e.g. `f"{self._xfstests_bld_path}/run-fstests/disks/vdg"`.

gctucker: Or maybe a template variable and then it will be defined in the script globals, like the version in `kver.jinja2`.

src/fstests/fstests.jinja2
```diff
@@ -36,6 +38,50 @@ class Job(BaseJob):
...
+            name = test_case.attrib['name'].split('/')
+            if not name[0] == 'ext4':
+                name.insert(0, 'ext4')
+            current = tree
+            for item in name:
+                if not item in current:
+                    current[item] = {}
+                current = current[item]
+            current['result'] = result
+        return self._parse_to_json({}, tree, 'FSTest Parsing Results')
```
gctucker, line 83: This should be `fstests` to match what is in the top-level node which was already sent by the `runner.py` script.
```suggestion
        return self._parse_to_json({}, tree, 'fstests')
```

src/fstests/fstests.jinja2
```diff
@@ -63,7 +109,9 @@ class Job(BaseJob):
         if self._check_kvm_xfstest_env():
             if self._kernel_config(src_path) == 'pass':
                 if self._kernel_build(src_path) == 'pass':
-                    return self._run_smoke_tests()
-        else:
-            return 'fail'
+                    if self._run_smoke_tests() == 'pass':
```
gctucker, line 112: We could have some code like `def _run_fstests(self, src_path):` which does:
```python
if not self._kernel_config(src_path):
    return False
if not self._kernel_build(src_path):
    return False
# etc.
```
Then return `'pass' if res is True else 'fail'` from `_run()`.

src/fstests/fstests.jinja2
```diff
@@ -63,7 +109,9 @@ class Job(BaseJob):
...
+                        if self._get_xml_from_vm(src_path) == 'pass':
+                            with open(os.path.join(src_path, 'results.json'), 'w') as result_file:
+                                result_file.write(json.dumps(self._parse_results(src_path)))
```
gctucker, line 115: We could add results for the kernel config and kernel build steps, and then also add the actual fstests results.  We could have `fstests.kernel.config`, `fstests.kernel.build` etc. for the tests run directly in this template and then add the results parsed from the XML file.

Then we should also catch any exception while parsing the results and sent the other results we already have, for example `fstests.smoke.run`, `fstests.smoke.xml`, `fstests.smoke.parse` for each method called earlier.

gctucker: To separate the results added by KernelCI from the actual fstests, we could have `fstests.kernelci.kernel.config`, `fstests.kernelci.smoke.run` etc. then `fstests.ext4.123` for the actual fstest results.  That way all the things added by KernelCI are in a separate sub-group called `kernelci` to make it clearer that they're not from the test suite.

gctucker:
Here's a sample `results.json` file for KUnit by the way, as an example:
[results.json.txt](https://github.com/kernelci/kernelci-pipeline/files/9478385/results.json.txt)

alexandrasp:
Example from a results.xml [results.xml.txt](https://github.com/kernelci/kernelci-pipeline/files/9478417/results.xml.txt)

src/fstests/fstests.jinja2
```diff
@@ -36,6 +38,50 @@ class Job(BaseJob):
...
+        for test_case in root.iter('testcase'):
+            if test_case.find('skipped') != None:
+                result = None
+            elif test_case.find('failure') != None:
+                result = 'fail'
+            else:
+                result = 'pass'
+            name = test_case.attrib['name'].split('/')
+            if not name[0] == 'ext4':
+                name.insert(0, 'ext4')
```
gctucker, line 76: I think the logic should be to add `ext4` if the name is `generic` or `shared` as these are the special names.  We could be running the whole test suite with another file system type e.g. `btrfs`, and then we would have `btrfs.generic`.

gctucker: We probably need to clarify how this will work if we want to run the tests with several file systems in one job.

gctucker: If we can only run the tests with one file system type, then there will always only be one file system type in the XML results and then the logic would be to find the name of the file system type first, then add any results that isn't the file system type as a sub-group.  Basically like the logic in the current version of the code but with a variable rather than `ext4`.

The produced format should be the same in this case, but there could be problems if we were using a different file system type.

gctucker: It seems like the file system type is provided at the top of the XML file:
```xml
        <properties>
                <property name="TESTCFG" value="ext4/4k"/>
        </properties>
```

src/fstests/fstests.jinja2
```diff
@@ -36,6 +38,50 @@ class Job(BaseJob):
...
+            current = tree
+            for item in name:
+                if not item in current:
+                    current[item] = {}
+                current = current[item]
+            current['result'] = result
```
gctucker, line 82: OK this is about creating a tree structure just using the names from the flat XML list of results.  Then the `_parse_to_json()` method creates another tree with the format used by the API (minus the extra fields populated in the base `python.jinja2` template such as revision, path, status etc. as well as _id and parent for the root node).

This seems like a reasonable approach as producing the final data structure directly with `node` and `child_nodes` would be more complicated, it would need to look up the child node names all the time.

However, `parse_to_json()` is a misleading method name as it's not parsing the input data and it's not serialising the output to JSON.  I think it would be clearer if this method was renamed to `_parse_xml()` and returned the data structure it creates (so `tree`), then another method would be called to do the conversion to the API data structure e.g. `_convert_for_api()` and return the final data structure that can be sent as JSON to the API (via the base template code that adds the extra fields).

Does this make sense?

gctucker: Essentially:
```python
def _parse_results(self, kdir):
    tree_data = self._parse_xml_results(os.path.join(kdir, 'results.xml'))
    api_data = self._convert_results_for_api(tree_data)
    return api_data
```
with `_parse_xml_results()` doing the conversion from XML to the tree structure, and `_convert_results_for_api()` to produce the data structure stored in `results.json`.

src/fstests/fstests.jinja2
alexandrasp: Updated to use `--xfstests-bld-path`

src/fstests/fstests.jinja2
alexandrasp: Updated, now we are returning False/True as we discussed.

src/fstests/fstests.jinja2
alexandrasp: updated, now the root tree is fstests.

src/fstests/fstests.jinja2
alexandrasp: Updated accordingly

src/fstests/fstests.jinja2
alexandrasp: Ok, I will include these changes ahead, the submit parse seems to be a good PR for this refactor.

src/fstests/fstests.jinja2
alexandrasp: Ok, as we discussed, for now, I will keep `ext4` because this will be the only case used for PoC. But I agree this is important and it will make the tree flexible to handle other fs types.

src/fstests/fstests.jinja2
alexandrasp: Updated, and also handling the exceptions that could be raised while parsing results.

src/fstests/fstests.jinja2
gctucker: I don't see this, the path is still hard-coded to `/home/kernelci/alex` in this commit.

src/fstests/fstests.jinja2
```diff
@@ -46,6 +48,40 @@ class Job(BaseJob):
             print('XML not found.', e)
             return 'fail'
 
+    def _parse_to_json(self, node, tree, name=None):
+        if 'result' in tree:
+            node['name'] = name
+            node['result'] = tree['result']
+        else:
+            if (not name == None): node['name'] = name
```
gctucker:
```python
if name is not None:
    node['name'] = name
```
But also I think there are a few issues with this part of the code.

Input data structures should not be modified in the method, that's a Python convention.  Then I'm not sure why you're passing a `node` object to the method when it's basically creating one with `name`, `result` and `child_nodes`.  And I don't quite get the logic about passing `name` which could be `None` - is this just for changing the root node name to `fstests`?

src/fstests/fstests.jinja2
```diff
@@ -46,6 +48,40 @@ class Job(BaseJob):
...
+                child_nodes.append(self._parse_to_json({}, tree[name], name))
+            node['child_nodes'] = child_nodes
+        return node
+
+    def _parse_results(self, kdir):
+        xml_file = ET.parse(f"{kdir}/results.xml")
+        root = xml_file.getroot()
+        tree = {}
+        for test_case in root.iter('testcase'):
+            if test_case.find('skipped') != None:
```
gctucker, line 61:
```
if test_case.find('skipped') is not None:
```
`None` is a singleton object in Python so the standard way to compare it with another object is with `is` / `is not`.

src/fstests/fstests.jinja2
```diff
@@ -40,7 +45,7 @@ class Job(BaseJob):
 
     def _get_xml_from_vm(self, kdir):
         try:
-            result = subprocess.run([f"debugfs -R \"dump ext4/results-4k/results.xml {kdir}/results.xml\" /home/kernelci/alex/fstests-poc/fstests/disks/vdg"], check=True, shell=True)
+            result = subprocess.run([f"debugfs -R \"dump ext4/results-4k/results.xml {kdir}/results.xml\" {xfstests_bld_path}/run-fstests/disks/vdg"], check=True, shell=True)
```
gctucker, line 48: Alright, this should be part of the first commit as there's no value in first adding the hard-coded path and then removing it - it's not part of a logic series of development steps.

src/fstests/fstests.kernelci.conf
```diff
@@ -0,0 +1,3 @@
+[fstests_runner]
```
gctucker, line 1: I think you might need to add other settings such as the API URL etc.  Basically everything except optional things like `--verbose` and credentials such as the API token.

src/fstests/fstests.jinja2
```diff
@@ -110,13 +110,15 @@ class Job(BaseJob):
...
-                        if self._get_xml_from_vm(src_path) == 'pass':
-                            with open(os.path.join(src_path, 'results.json'), 'w') as result_file:
-                                result_file.write(json.dumps(self._parse_results(src_path)))
-        return 'fail'
+        if not self._kernel_config(src_path):
+            return 'fail'
+        if not self._kernel_build(src_path):
+            return 'fail'
+        if not self._run_smoke_tests():
+           return 'fail'
```
gctucker, line 126: Inconsistent indentation.

src/fstests/fstests.jinja2
alexandrasp: @gctucker my last question about this PR is if I should include these changes in this commit or should  I keep them in the commit refactoring this part of the code ahead?

src/fstests/fstests.jinja2
```diff
@@ -3,9 +3,16 @@
 
 {%- extends 'base/python.jinja2' %}
 
+{%- block python_globals %}
+{{ super() }}
+xfstests_bld_path = '{{ xfstests_bld_path }}'
```
gctucker, line 8: This could be passed to the constructor instead of a global variable with `{% block python_job_constr %}`.

src/fstests/fstests.jinja2
```diff
@@ -16,35 +23,87 @@ class Job(BaseJob):
...
+            elif test_case.find('failure') is not None:
+                result = 'fail'
+            else:
+                result = 'pass'
+            name = test_case.attrib['name'].split('/')
+            if not name[0] == 'ext4':
+                name.insert(0, 'ext4')
+            current = tree
+            for item in name:
+                if not item in current:
```
gctucker, line 72: This is a bit easier to read:
```python
if item not in current:
```

gctucker: But also I think `current.setdefault(item, {})` would work too.

src/fstests/fstests.jinja2
```diff
@@ -16,35 +23,87 @@ class Job(BaseJob):
...
+    def _convert_results_for_api(self, tree, name):
+        if 'result' in tree:
+            return {
+                'name': name,
+                'result': tree['result'],
+            }
+        else:
+            child_nodes = []
+            for child_name in tree:
+                child_nodes.append(self._convert_results_for_api(tree[child_name], child_name))
```
gctucker, line 97: This is a very long line too.  It could easily be wrapped, or an intermediate variable with the results could then be appended to the list.

PR #124, opened 2022-08-31
src/*.py: get nodes from 'items' key

JenySadadia:
The `/nodes` endpoint response has now been changed
due to pagination implementation.
Need to fetch value of 'items' from json response
dictionary to get nodes.

Depends on API Pagination PR: https://github.com/kernelci/kernelci-api/pull/164

### DIFF
--- a/src/set_timeout.py
+++ b/src/set_timeout.py
@@ -31,8 +31,8 @@ class SetTimeout:
 
     def _update_pending_child(self, parent_id):
         """Set child node status to timeout when parent is timed out"""
-        child_nodes = self._db.get_nodes({"status": "pending",
-                                          "parent": parent_id})
+        child_nodes = (self._db.get_nodes({"status": "pending",
+                                          "parent": parent_id})).get('items')
         for child in child_nodes:
             child['status'] = "timeout"
             self._db.submit({'node': child})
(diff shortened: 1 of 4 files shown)


src/set_timeout.py
```diff
@@ -31,8 +31,8 @@ def __init__(self, configs, args):
 
     def _update_pending_child(self, parent_id):
         """Set child node status to timeout when parent is timed out"""
-        child_nodes = self._db.get_nodes({"status": "pending",
-                                          "parent": parent_id})
+        child_nodes = (self._db.get_nodes({"status": "pending",
+                                          "parent": parent_id})).get('items')
```
gctucker, line 35: Maybe the `.get('items')` part should be done in `Database.get_node()` in kernelci-core instead?

JenySadadia: Yes, we can do that. However, the client will not receive other information from the response. The typical response from pagination is a dictionary with items (node objects), total (number of nodes received), limit, and offset values.

gctucker: In that case we may add something to the Database class methods to handle that, but I think exposing the raw data returned from the API is a bit fragile.  Say, if we change the API implementation then the code using the Database class will have to be updated too.  Using it as an abstraction layer helps with keeping the client code compatible.

Maybe an initial step would be to keep the `Database.get_nodes()` method return the same kind of data as before, using `.items()`.  Then we can look into how to properly handle pagination in this class.

JenySadadia:
Another way was to update `KernelCI_API.get_nodes` to return only `items`.
Closing this one as https://github.com/kernelci/kernelci-core/pull/1411 has been merged.

PR #127, opened 2022-09-02
tarball.py: update state names and put tarball URL in tarball node

gctucker:
Use the new Node state names and only set the tarball artifact URL in the tarball node, not the checkout node.

Depends on #112

### DIFF
--- a/src/tarball.py
+++ b/src/tarball.py
@@ -98,18 +98,23 @@ scp \
         os.unlink(tarball_path)
         return tarball
 
-    def _create_tarball_node(self, checkout_node, tarball):
-        node = {
+    def _create_tarball_node(self, checkout_node):
+        tarball_node = {
             'parent': checkout_node['_id'],
             'name': 'tarball',
             'path': checkout_node['path'] + ['tarball'],
+            'revision': checkout_node['revision'],
+        }
+        return self._db.submit({'node': tarball_node})[0]
+
+    def _make_tarball_node_available(self, tarball_node, tarball):
+        tarball_node.update({
+            'state': 'available',
             'artifacts': {
                 'tarball': urllib.parse.urljoin(self._storage_url, tarball),
             },
-            'revision': checkout_node['revision'],
-            'state': 'available',
-        }
-        self._db.submit({'node': node})
+        })
+        return self._db.submit({'node': tarball_node})
 
     def _update_checkout_node(self, node, describe, version):
         node['revision'].update({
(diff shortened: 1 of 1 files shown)


src/tarball.py
```diff
@@ -98,25 +98,24 @@ def _push_tarball(self, config, describe):
...
             'name': 'tarball',
-            'artifacts': checkout_node['artifacts'],
+            'path': checkout_node['path'] + ['tarball'],
+            'artifacts': {
+                'tarball': urllib.parse.urljoin(self._storage_url, tarball),
+            },
             'revision': checkout_node['revision'],
-            'status': status,
-            'result': result,
+            'status': 'available',
```
JenySadadia, line 110: I think the tarball node should be created with `running` state first. After the execution of `_push_tarball` successfully, the state of the node should be changed to `available`.

```suggestion
            'status': 'running',
```

src/tarball.py
```diff
@@ -98,25 +98,24 @@ def _push_tarball(self, config, describe):
...
 
-    def _update_checkout_node(self, node, describe, version, tarball):
+    def _update_checkout_node(self, node, describe, version):
         node['revision'].update({
             'describe': describe,
             'version': version,
         })
-        node['artifacts'] = {
-            'tarball': urllib.parse.urljoin(self._storage_url, tarball),
-        }
```
JenySadadia, line 119: Why is this removed?

gctucker: The tarball artifact belongs in the tarball node, not the checkout node.

src/tarball.py
gctucker: Yes, that's right.  Will update the code accordingly.

Eventually this probably should become a template to run in a local shell anyway, based on `python.jinja2`.

gctucker: OK I've added an extra commit as it's more than just changing the state names.

src/tarball.py
```diff
@@ -98,25 +98,30 @@ def _push_tarball(self, config, describe):
...
+    def _create_tarball_node(self, checkout_node):
+        tarball_node = {
             'parent': checkout_node['_id'],
             'name': 'tarball',
-            'artifacts': checkout_node['artifacts'],
+            'path': checkout_node['path'] + ['tarball'],
             'revision': checkout_node['revision'],
-            'status': status,
-            'result': result,
+            'state': 'running',
```
JenySadadia, line 107: This line can be omitted as the default node state is `running`.

src/tarball.py
```diff
@@ -98,27 +98,30 @@ def _push_tarball(self, config, describe):
...
+        return self._db.submit({'node': tarball_node})[0]
+
+    def _make_tarball_node_available(self, tarball_node, tarball):
+        tarball_node.update({
+            'state': 'available',
+            'artifacts': {
+                'tarball': urllib.parse.urljoin(self._storage_url, tarball),
+            },
+        })
+        return self._db.submit({'node': tarball_node})
```
JenySadadia, line 117: The return value is not being used.
```suggestion
        self._db.submit({'node': tarball_node})
```

PR #129, opened 2022-09-05
src/trigger: update checkout node state

JenySadadia:
Create checkout node with "available" state,
as the node will be ready to have child nodes
as soon as it is created.


gctucker:
I think we can drop this now that checkout and tarball nodes are being merged into one with #132.

JenySadadia:
Yes, closing this one.

PR #130, opened 2022-09-05
src/test_report: listen to tarball node with "done" state

JenySadadia:
Update node state to listen to completed
tarball nodes.


gctucker:
I think we can drop this now that we have #132

JenySadadia:
Yes.

PR #131, opened 2022-09-07
T37566 src/timeout: update node state to 'done' when timed out

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/178

Update node state and all its child nodes to
'done' when the parent node is timed out.

### DIFF
--- a/src/timeout.py
+++ b/src/timeout.py
@@ -36,6 +36,8 @@ class Timeout:
         the child nodes is incomplete"""
         holdoff_time = datetime.fromisoformat(node['holdoff'])
         if current_time > holdoff_time:
+            self._logger.log_message(logging.INFO,
+                                     f"Holdoff reached for {node['_id']}")
             total_child_nodes = self._db.count_nodes({
                 'parent': node['_id'],
             })
@@ -46,8 +48,12 @@ class Timeout:
             })
 
             if total_child_nodes > completed_child_nodes:
+                self._logger.log_message(logging.INFO,
+                                         f"Closing the node {node['_id']}")
                 node['state'] = 'closing'
             else:
+                self._logger.log_message(logging.INFO,
+                                         f"Completing the node {node['_id']}")
                 node['state'] = 'done'
             self._db.submit({'node': node})
 
@@ -67,6 +73,8 @@ class Timeout:
             expires = datetime.fromisoformat(node['created']) + \
                 timedelta(hours=node['timeout'])
             if current_time > expires:
+                self._logger.log_message(logging.INFO,
+                                         f"Node timed-out {node['_id']}")
                 node['state'] = 'done'
                 self._db.submit({'node': node})
                 self._update_child_nodes(node['_id'])


JenySadadia:
> This service is not running on staging due to this error:
> 
> ```
> The following arguments or settings are required: --db-config
> ```
> 
> That's a bit strange because the `db_config` option is set in the `[DEFAULT]` section in the staging settings file. So I'm not sure what's going on there.

I'm not quite sure about this as the script is working fine locally.

JenySadadia:
> This service is not running on staging due to this error:
> 
> ```
> The following arguments or settings are required: --db-config
> ```
> 
> That's a bit strange because the `db_config` option is set in the `[DEFAULT]` section in the staging settings file. So I'm not sure what's going on there.

The script is working fine even with `staging.kernelci.org` branch locally.

JenySadadia:
Hi @gctucker , I think we need to rename this script as `set_timeout` doesn't justify its working anymore.
How about we rename it to `verify_timeout` or `check_timeout`?

JenySadadia:
> What is this service expected to be doing? Should it be looking for nodes that reach either `holdoff` or `timeout`? I think it's fine to rename it to `timeout.py`.

It is looking for timed out nodes with `closing` state. Okay, thanks for the input. I'll rename it to `timeout.py`.

gctucker:
Also I think it's a generic concept that people will understand, OK the timeout service is for nodes that time out (and holdoff is a kind of timeout, just to move to a different state).

JenySadadia:
> Also I think it's a generic concept that people will understand, OK the timeout service is for nodes that time out (and holdoff is a kind of timeout, just to move to a different state).

Yes, I am working on a client script for holdoff.

gctucker:
> > Also I think it's a generic concept that people will understand, OK the timeout service is for nodes that time out (and holdoff is a kind of timeout, just to move to a different state).
> 
> Yes, I am working on a client script for holdoff.

Oh you mean a different one?  I thought maybe the same script could be handling both as it's effectively the same thing I believe, just a different timestamp and a different state name.

JenySadadia:
> > > Also I think it's a generic concept that people will understand, OK the timeout service is for nodes that time out (and holdoff is a kind of timeout, just to move to a different state).
> > 
> > 
> > Yes, I am working on a client script for holdoff.
> 
> Oh you mean a different one? I thought maybe the same script could be handling both as it's effectively the same thing I believe, just a different timestamp and a different state name.

No, I am going to extend the same `timeout.py` script for holdoff too. I just have a separate script for testing atm.

src/timeout.py
```diff
@@ -20,44 +20,44 @@
...
-        """Set Node status to timeout if maximum wait time is over"""
+    def _update_closing_node(self, node):
+        """Set Node state to done if maximum wait time is over"""
         current_time = datetime.utcnow()
-        max_wait_time = datetime.fromisoformat(node['created']) + \
-            timedelta(hours=node['max_wait_time'])
-        if current_time > max_wait_time:
-            node['status'] = "timeout"
+        expires = datetime.fromisoformat(node['created']) + \
+            timedelta(hours=node['timeout'])
```
gctucker, line 45: I thought the `timeout` field was supposed to be an absolute timestamp rather than a time delta now?  Maybe other parts of the pipeline still need to be updated for that and a follow-up PR will be needed here.

JenySadadia: This PR is about node state changes. I will create another one for the `timeout` field.

mgalka (approved):

JenySadadia:
Closing this PR as the commits from this one are already merged with https://github.com/kernelci/kernelci-pipeline/pull/150

PR #132, opened 2022-09-08
Merge tarball node into checkout node

gctucker:
Drop the concept of a `tarball` node and add the source tarball URL as an artifact of the `checkout` node.  Update the trigger, tarball, runner, test_report and complete_hack scripts accordingly.


gctucker:
Tested OK on staging.

PR #133, opened 2022-09-08
src/runner.py: pass templates_paths argument to .generate()

gctucker:
The keywoard argument templates_path has been renamed to templates_paths as it can now be either a string with a single path or a list of paths.


gctucker:
Tested OK on staging.

PR #134, opened 2022-09-09
src/test_report.py: drop redundant report log message

gctucker:
Drop a redundant log message showing the email contents in _create_email() as it's already done in the main loop.


gctucker:
Tested OK on staging.

PR #135, opened 2022-09-09
src/fstests: update pub/sub configuration

alexandrasp:
Now the tarball and checkout nodes have been merged into one, there's no tarball node but only checkout node and it contains the tarball URL when it's in available state.

### DIFF
--- a/src/fstests/runner.py
+++ b/src/fstests/runner.py
@@ -83,11 +83,11 @@ class FstestsRunner:
     def run(self, args):
         sub_id = None
         if not args.node_id:
-            sub_id = self._db.subscribe_node_channel({
-                'op': 'created',
-                'name': 'tarball',
+            sub_id = self._db.subscribe_node_channel(filters={
+                'name': 'checkout',
+                'state': 'available',
             })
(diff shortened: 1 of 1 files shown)


alexandrasp:
@gctucker just to clarify, as we discussed, I decided to push these changes because I was running tests and it is totally unrelated to the jinja2 file. And in order to listen and let the script listen for `checkout`  availability we need this.

gctucker:
There's a typo in the subject: `listering`

I think pub/sub would be a better term actually.

gctucker:
> @gctucker just to clarify, as we discussed, I decided to push these changes because I was running tests and it is totally unrelated to the jinja2 file. And in order to listen and let the script listen for `checkout` availability we need this.

Yes, this change is needed purely because the rest of the pipeline is evolving and we're dropping the `tarball` node.

alexandrasp:
> There's a typo in the subject: `listering`
> 
> I think pub/sub would be a better term actually.

thanks for noticing this.

gctucker (approved):

PR #136, opened 2022-09-09
src/test_report.py: add command for generating report by hand

gctucker:
Add a "single" command for generating a single report from a given checkout node id.  The report can be dumped to stdout and/or sent over SMTP.

Update the implementation to reuse the common code between these 2 commands accordingly.

### DIFF
--- a/src/test_report.py
+++ b/src/test_report.py
@@ -125,7 +125,24 @@ class TestReport:
         )
         return content, subject
 
-    def run(self):
+    def _send_email(self, content, subject):
+        smtp = self._smtp_connect()
+        if smtp:
+            email_msg = self._create_email(
+                self._email_send_to, self._email_send_from,
+                subject, content
+            )
+            smtp.send_message(email_msg)
+            smtp.quit()
+
+    def _make_report(self, root_node, dump=True, send=False):
+        content, subject = self._create_test_report(root_node)
+        if dump:
+            print(content, flush=True)
+        if send:
+            self._send_email(content, subject)
+
+    def run_loop(self):
         sub_id = self._db.subscribe_node_channel(filters={
             'name': 'checkout',
             'state': 'done',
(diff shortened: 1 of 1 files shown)


gctucker:
Sample command, from a Docker container:
```
./pipeline/test_report.py --settings=config/kernelci.conf single --node-id=631b387ecab77b89d9c6eac8 --dump --smtp-host=dummy --db-config=staging.kernelci.org
```
Note: The SMTP arguments are still required even though they shouldn't when `--send` is not used.  That's something to clean-up in another PR I guess.

src/test_report.py
```diff
@@ -125,7 +125,24 @@ def _create_test_report(self, root_node):
...
-    def run(self):
+    def _send_email(self, content, subject):
+        smtp = self._smtp_connect()
+        if smtp:
+            email_msg = self._create_email(
+                self._email_send_to, self._email_send_from,
+                subject, content
+            )
+            smtp.send_message(email_msg)
+            smtp.quit()
```
mgalka, line 136: What is the reason to keep this method as an internal API?
I don't see the point why it can't be public. Having said that, I think we should consider separating e-mail related logic and the report. I think we should have 2 separate classes for it.

mgalka (commented):
I think we can treat my comment as a follow up action for this PR.

src/test_report.py
gctucker: This service doesn't provide any API, it's only a command line tool.  It wasn't designed to be imported and reused.  This PR is a refactor and the code to send an email is being moved to a separate method, but that's all implementation detail.

The service needs to have an event loop so that every time a kernel revision has been tested a report is generated and sent.  So that needs to be done within the service's code.  The class here is only to provide the command line interface like in the `kci_*` tools.

We could rework this further by separating the logic that creates the email report (which probably could go into kernelci-core anyway) and the logic that sends emails, but I think that's beyond the scope of this PR.

mgalka (approved):
Let's refactor the code when it's functionality will be moved to kernelci-core

PR #138, opened 2022-09-09
tarball.py: update checkout node only once

gctucker:
Send an update to the checkout node only once with the version
    information and the tarball URL and state set to "available" rather
    than doing this in two steps.  This was useful previously when there
    was a separate tarball node.

Depends on #132


gctucker:
Tested OK on staging.

PR #140, opened 2022-09-16
Trigger all build configs

gctucker:
Rather than specifying a single build config, update the trigger service to iterate over all the build configs defined in the YAML configuration.  Adjust log messages accordingly and fix exception handling as we can now run this in a loop with the `--poll` argument and not the `--force` argument.

### DIFF
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -24,47 +24,10 @@ import requests
 from logger import Logger
 
 
-def _run_trigger(args, build_config, db, logger):
-    head_commit = kernelci.build.get_branch_head(build_config)
-    node_list = db.get_nodes({
-        "revision.commit": head_commit,
-    })
-    if node_list:
-        if args.force:
-            logger.log_message(
-                logging.INFO,
-                f"Creating new node with existing commit {head_commit}"
-            )
-        else:
-            logger.log_message(
-                logging.INFO,
-                f"Node alread exists with commit {head_commit}"
-            )
-            return
-    else:
-        logger.log_message(
-            logging.INFO,
-            f"Found new revision for {build_config.name}: {head_commit}"
-        )
-
-    revision = {
-        'tree': build_config.tree.name,
-        'url': build_config.tree.url,
-        'branch': build_config.branch,
-        'commit': head_commit,
-    }
-    node = {
-        'name': 'checkout',
-        'path': ['checkout'],
-        'revision': revision,
-    }
-    db.submit({'node': node})[0]
-
-
 class cmd_run(Command):
     help = "Submit a new revision to the API based on local git repo"
     args = [
-        Args.build_config, Args.db_config,
+        Args.db_config,
     ]
     opt_args = [
         {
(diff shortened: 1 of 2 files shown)


gctucker:
This makes me think, "build configs" is a rather inaccurate description.  Something like "tree configs" would be better, as it's really about the kernel trees than builds.

gctucker:
I've also updated the staging config file to drop `--force` and set a polling period of 1h to detect mainline updates.

gctucker:
@JenySadadia @mgalka Could you please review this so we can have it merged by the end of the week?

src/trigger.py
```diff
@@ -82,25 +45,80 @@ class cmd_run(Command):
...
+            f"{message:32s} {build_config.name:32s} {head_commit}"
+        )
+
+    def _run_trigger(self, build_config, force):
+        head_commit = kernelci.build.get_branch_head(build_config)
+        node_list = self._db.get_nodes({
+            "revision.commit": head_commit,
+        })
+
+        config_revision = f"{build_config.name:32s} {head_commit}"
```
JenySadadia, line 63: `config_revision` is not used anywhere.

src/trigger.py
```diff
@@ -82,25 +45,78 @@ class cmd_run(Command):
...
+
+    def _log_revision(self, message, build_config, head_commit):
+        self._logger.log_message(
+            logging.INFO,
+            f"{message:32s} {build_config.name:32s} {head_commit}"
+        )
+
+    def _run_trigger(self, build_config, force):
+        head_commit = kernelci.build.get_branch_head(build_config)
+        node_list = self._db.get_nodes({
```
JenySadadia, line 59: As we just need to check if the revision node exists or not, this can be done by getting the count instead of the actual node.
```suggestion
        node_list = self._db.count_nodes({
```

gctucker: I think that can be fixed as a follow-up PR.  The count API entry point wasn't available when this PR was made.

src/trigger.py
```diff
@@ -82,25 +45,78 @@ class cmd_run(Command):
...
+            "revision.commit": head_commit,
+        })
+
+        if node_list:
+            if force:
```
JenySadadia, line 64: We can define `_force` in `cmd_run` class to avoid passing the variable in `_iterate_build_configs`, and `_run_trigger`.

`self._force = args.force`

gctucker: We could, but I don't think that's a good idea and that's why I'm passing it as an argument.  That way is about the run, whether it should force using a new revision or not.  The same object could be called several times.

JenySadadia:
Tested OK on local:
```
kernelci-pipeline-trigger | 09/26/2022 06:11:50 AM UTC [INFO] New revision                     kernelci_staging-mainline        59e0140d56f10c74066759349f3f596bd5fbdfcd
kernelci-pipeline-notifier | 09/26/2022 06:11:50 AM UTC [INFO] 2022-09-26 06:11:50.546185  59e0140d56f1  633142a6bd96136574e0fcc7  Running    -         checkout
kernelci-pipeline-tarball | 09/26/2022 06:11:50 AM UTC [INFO] Updating repo for kernelci_staging-mainline
kernelci-pipeline-trigger | 09/26/2022 06:11:52 AM UTC [INFO] New revision                     kernelci_staging-next            6a65e9f199a67c9e266e9b5f1790c8de3f77737c
kernelci-pipeline-notifier | 09/26/2022 06:11:52 AM UTC [INFO] 2022-09-26 06:11:52.180036  6a65e9f199a6  633142a8bd96136574e0fcc8  Running    -         checkout
kernelci-pipeline-trigger | 09/26/2022 06:11:53 AM UTC [INFO] New revision                     kernelci_staging-stable          19b5bc44ea621efe664d68dd15fa722af5b001ef
kernelci-pipeline-notifier | 09/26/2022 06:11:53 AM UTC [INFO] 2022-09-26 06:11:53.641684  19b5bc44ea62  633142a9bd96136574e0fcc9  Running    -         checkout
kernelci-pipeline-trigger | 09/26/2022 06:11:54 AM UTC [INFO] New revision                     mainline                         f76349cf41451c5c42a99f18a9163377e4b364ff
kernelci-pipeline-trigger | 09/26/2022 06:11:54 AM UTC [INFO] Not polling.
```

src/trigger.py
```diff
@@ -82,25 +45,78 @@ class cmd_run(Command):
...
+            'url': build_config.tree.url,
+            'branch': build_config.branch,
+            'commit': head_commit,
+        }
+        node = {
+            'name': 'checkout',
+            'path': ['checkout'],
+            'revision': revision,
+        }
+        self._db.submit({'node': node})[0]
```
JenySadadia, line 89: No need to fetch anything further from the response as it is not being returned.
```suggestion
        self._db.submit({'node': node})
```

JenySadadia: I will add this change in a follow-up PR.

JenySadadia:
Tested OK on staging:
```
yesterday at 16:52:1509/26/2022 11:22:15 AM UTC [INFO] Existing revision                kernelci_staging-mainline        d4b335e6bad0032e299e5012189a03566f2a5fb4
yesterday at 16:52:1609/26/2022 11:22:16 AM UTC [INFO] New revision                     kernelci_staging-next            a38b947d0bcd147c748fc937880fd809f2148a67
yesterday at 16:52:1709/26/2022 11:22:17 AM UTC [INFO] Existing revision                kernelci_staging-stable          10b0ab5c002ef594deff871dc1eaf2234bd1598d
yesterday at 16:52:1809/26/2022 11:22:18 AM UTC [INFO] Existing revision                mainline                         f76349cf41451c5c42a99f18a9163377e4b364ff
yesterday at 16:52:1809/26/2022 11:22:18 AM UTC [INFO] Sleeping for 3600s
```

JenySadadia (approved):

PR #146, opened 2022-09-20
Refactor test report script

JenySadadia:
- Update the `test_report` script with below refactoring:
    * rename `_make_report` function with `_get_report`
    * modify signature of `_send_email` method
    * remove logic to create email from `_send_email`
    * call function to create email in `_get_report`
    * maintain the alphabetical order for method definitions
    * use email related methods from `email_sender` module
- src/email_sender: module to create and send emails


gctucker (commented):
One slightly confusing thing here is that `email_sender.py` is not a service but just part of the implementation of `test_report.py` which is a service run in `docker-compose`.  Is it really worth splitting this into 2 files?  We could also have separate classes in the same file.

JenySadadia:
> 

I wanted to treat `email_sender` as a separate utility module just for emails. That's why I separated it from `test_report`. The client script `test_report` should only have implementation related to API test reports. Do you suggest otherwise?

src/test_report.py
```diff
@@ -108,39 +92,18 @@ def _get_results_data(self, root_node):
...
-        if smtp:
-            email_msg = self._create_email(
-                self._email_send_to, self._email_send_from,
-                subject, content
-            )
-            smtp.send_message(email_msg)
-            smtp.quit()
-
-    def _make_report(self, root_node, dump=True, send=False):
+    def _get_report(self, root_node, dump=True, send=False):
```
mgalka, line 95: I find the method name confusing  `_get_report` suggests that it actually will return something while it doesn't.
It actually does several things: it creates report content (and subject), it sends it over e-mail or dumps it to stdout. I'd go with refactoring the code so we can have separate logic that creates the content and separate which sends it.

src/test_report.py
```diff
@@ -108,39 +92,18 @@ def _get_results_data(self, root_node):
...
-            self._send_email(content, subject)
+            try:
+                email_msg = self._email_sender.create_email(
+                    subject, content
+                )
+                self._email_sender.send_email(email_msg)
+            except Exception as err:
+                self._logger.log_message(logging.ERROR, err)
 
     def run_loop(self):
```
mgalka, line 108: I'm not entirely convinced that `run_loop` logic belongs to the TestReport. For me it belongs rather to the `class cmd_run`. Is there a reason why it's here?

src/test_report.py
```diff
@@ -164,7 +127,7 @@ def run_loop(self):
 
     def run_from_id(self, node_id, dump, send):
```
mgalka, line 128: The same comment as for `run_loop` method applies here, but instead of `cmd_run` I'd put it in the `cmd_single`

mgalka (commented):
some design related comments

src/test_report.py
JenySadadia: Thanks for pointing that out.
I have created two separate methods to dump and send reports.
Creation of email will directly take place from `email_sender` now.

src/test_report.py
JenySadadia: The reason being is to align the class with the pattern of other client scripts.
Also, I created two different classes to run the service in a loop and just for a single node by the `loop` and `run` commands respectively.

src/test_report.py
```diff
@@ -133,49 +103,54 @@ def _create_test_report(self, root_node):
...
+        self._logger.log_message(logging.INFO, "Listening for completed nodes")
         self._logger.log_message(logging.INFO, "Press Ctrl-C to stop.")
 
         try:
             while True:
                 root_node = self._db.receive_node(sub_id)
-                self._make_report(root_node, dump=True, send=True)
+                content, subject = self._get_report(root_node)
+                self._dump_report(content)
+                self._send_report(subject, content)
```
mgalka, line 132:
```python
self._send_report(subject, content)
```
never raises an exception is it done on purpose?

src/test_report.py
mgalka: @JenySadadia 
I get your point, so if it is a pattern in other scripts I guess we should follow it. Creating 2 classes just to run single and loop versions seems a bit artificial.
I guess we should get back to the previous implementation with `run_loop()` and `run()`, though the design doesn't seem optimal. Can we do it?

mgalka (commented):
some things to  possibly change

src/test_report.py
JenySadadia: It is done to catch exceptions from `email_sender.create_and_send_email`. If I don't do that, traceback from `email_sender` will not be logged on the terminal.

src/test_report.py
JenySadadia: Can you please elaborate on the comment?
In my opinion, It is easier to maintain two different classes (having two different commands to serve) when the code grows gradually. I thought we did the same thing for `runner` as well.

mgalka: Ok, I get your point, however I think that implementation is not quite right. `TestReportLoop` and `TestReportSingle` classes have different purpose than `TestReport`. They are a kind of "runners" of the logic provided by `TestReport`, hence they shouldn't inherit from `TestReport`, they should rather take an instance of `TestReport` while they're constructed, store it as an attribute and use it in the `.run()` method.

Maybe later we can also create a separate `Runner` class that could be use to run code in the other pipeline components, but that's a follow up.

JenySadadia: If I don't inherit the `TestReport` class and use its instance instead, I need to use protected members like `_db` and `_logger` from `TestReportLoop` and `TestReportSingle` classes. That will not be a good coding practice.
Rather, I chose to go with the old implementation of two separate methods but renamed them to `loop` and `run`.

mgalka (approved):

mgalka:
@JenySadadia Can you test it on staging and check if everything works as expected before we merge it?

JenySadadia:
Looks fine on staging:

```
today at 05:40:4010/21/2022 12:10:40 AM UTC [INFO] Listening for completed nodes
today at 05:40:4010/21/2022 12:10:40 AM UTC [INFO] Press Ctrl-C to stop.
today at 05:47:49kernelci/staging-next staging-next-20221021.0: 2 runs 0 failures
today at 05:47:49
today at 05:47:49Summary
today at 05:47:49=======
today at 05:47:49
today at 05:47:49Tree:     kernelci
today at 05:47:49Branch:   staging-next
today at 05:47:49Describe: staging-next-20221021.0
today at 05:47:49URL:      https://github.com/kernelci/linux.git
today at 05:47:49SHA1:     b0e05d51b8a3df4daf4e3da8eab06ac4abd80ec6
today at 05:47:49
today at 05:47:49Name            | Result   | Total    | Failures
today at 05:47:49----------------+----------+----------+---------
today at 05:47:49kver            | pass     |        0 |        0
today at 05:47:49kunit           | pass     |      249 |        0
today at 05:47:49
today at 05:47:49
today at 05:47:49Failing tests
today at 05:47:49=============
```

PR #161, opened 2022-10-06
T38154 runner: Handle exceptions when creating jobs

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/191

POST node request can result in HTTPException if the node's parent is not in `available` state while creating a child.
Need to handle the exception in methods for creating nodes and scheduling jobs in the runner script.

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -10,6 +10,8 @@ import logging
 import os
 import sys
 import tempfile
+import json
+import requests
 
 import kernelci
 import kernelci.config
@@ -42,7 +44,12 @@ class Runner(Service):
             'artifacts': checkout_node['artifacts'],
             'revision': checkout_node['revision'],
         }
-        return self._db.submit({'node': node})[0]
+        try:
+            return self._db.submit({'node': node})[0], \
+                "Node created successfully"
+        except requests.exceptions.HTTPError as err:
+            err_msg = json.loads(err.response.content).get("detail", [])
+            return None, err_msg
 
     def _generate_job(self, node, plan_config, device_config, tmp):
         self.log.info("Generating job")
@@ -74,7 +81,9 @@ class Runner(Service):
         ))
 
         self.log.info("Creating test node")
-        node = self._create_node(checkout_node, plan)
+        node, msg = self._create_node(checkout_node, plan)
+        if not node:
+            return None, msg
 
         tmp = tempfile.TemporaryDirectory(dir=self._output)
         output_file = self._generate_job(node, plan, device, tmp.name)
@@ -128,9 +137,13 @@ class RunnerLoop(Runner):
             job, tmp = self._schedule_test(
                 checkout_node, self._plan, device
             )
+            if not job:
+                self.log.error(f"Failed to schedule job for \
+{self._plan.name}. Error: {tmp}")
+                continue
+
             if self._runtime.config.lab_type == 'shell':
                 self._job_tmp_dirs[job] = tmp
-            self._cleanup_paths()
 
         return True
 
@@ -163,6 +176,11 @@ class RunnerSingleJob(Runner):
     def _run(self, ctx):
         node, plan, device = (ctx[key] for key in ('node', 'plan', 'device'))
         job, tmp = self._schedule_test(node, plan, device)
+        if not job:
+            self.log.error(
+                f"Failed to schedule job for {plan.name}. Error: {tmp}"
+            )
+            return False
         if self._runtime.config.lab_type == 'shell':
             self.log.info("Waiting...")
             job.wait()
--- a/src/fstests/runner.py
+++ b/src/fstests/runner.py
@@ -8,6 +8,8 @@
 import os
 import sys
 import tempfile
+import json
+import requests
 
 import kernelci
 import kernelci.config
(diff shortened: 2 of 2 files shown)


src/runner.py
```diff
@@ -38,15 +38,18 @@ def __init__(self, configs, args):
...
+        try:
+            node = {
+                'parent': checkout_node['_id'],
+                'name': plan_config.name,
+                'path': checkout_node['path'] + [plan_config.name],
+                'group': plan_config.name,
+                'artifacts': checkout_node['artifacts'],
+                'revision': checkout_node['revision'],
+            }
+            return self._db.submit({'node': node})[0]
```
hardboprobot, line 50: This fails when passing a custom `--node-id`, as a new node is set to state=running by default, and the changes in https://github.com/kernelci/kernelci-api/pull/191 restrict the creation of child nodes for "available" nodes only.

JenySadadia: Yes, this should fail if the parent is in `running` state.

hardboprobot (approved):

src/runner.py
JenySadadia: I believe the exception was handled properly on failure and an error message got displayed. Right? @hardboprobot

hardboprobot: Yes, I got this:

```
400 Client Error: Bad Request for url: http://172.17.0.1:8001/node
Failed to schedule job for fstests
Waiting...
```

Having the specific error cause would be a nice a nice addition that could save us a few headaches, though.

src/runner.py
```diff
@@ -130,6 +135,12 @@ def loop(self):
                 job, tmp = self._schedule_test(
                     checkout_node, self._plan, device
                 )
+
+                if not job and not tmp:
+                    self._logger.log_message(logging.ERROR, f"Failed to \
+schedule job for {self._plan.name}")
```
hardboprobot, line 141: What I meant with having more info about the error was to be able to tell the cause quickly, if possible.

This is only an example, probably a very rare error case, but it can happen in many other API interactions: When something goes wrong while submitting a new job and you get a "Failed to schedule job" message, what do you do next and how do you find out what caused the failure?

The changes in https://github.com/kernelci/kernelci-api/pull/191 make the request return a concise error cause, that could be propagated to [the callers](https://github.com/kernelci/kernelci-core/blob/d1a99dc4c2d8a516d3dda9ebb0424914403d4061/kernelci/db/kernelci_api.py#L184) , right?

I think it's worth it to implement proper error propagation as a separate task instead of conflating a whole family of different errors into one category, we may lose a lot of information that way.

JenySadadia: Absolutely right. 
The same thing crossed my mind when I was working on this one but I needed to change the responses of a few methods to propagate errors e.g. I will need to return a string even when a node is created successfully.
Let me take a look at this again.

JenySadadia: Updated the code. Please check @hardboprobot.

src/runner.py
```diff
@@ -149,6 +163,10 @@ class RunnerSingleJob(Runner):
     def _run_single_job(self, checkout_node, plan, device):
         try:
             job, tmp = self._schedule_test(checkout_node, plan, device)
+            if not job and not tmp:
```
mgalka, line 166: Isn't it enough to have just `job` or `tmp` having a boolean value of `False` to mark job as failed?

```suggestion
            if not job or not tmp:
```

or maybe it's just enough to check for `job` being evaluated to False?

```suggestion
            if not job:
```

mgalka: btw is `tmp`  used anywhere in this method? If not it should rather by renamed to `_`

```python
    job, _ = self._schedule_test(checkout_node, plan, device)
```

mgalka (commented):
Some things to consider

src/runner.py
JenySadadia: `tmp` is useful to display the error message.
Updated the code accordingly.

gctucker: Why not have the whole logic in to the main try-except block rather than adding code especially for handling this?

JenySadadia: `RunnerLoop.loop` already has try-except block implemented which is causing the runner to be terminated after catching exceptions.
API can raise exceptions while creating nodes. Hence, added a special block in `create_node`.
Seems like I would need to look into this again as service base class implementation just got merged.

gctucker: I guess we need to be careful if we want to keep the runner running even when the API is throwing exceptions.  It's probably safer to let it terminate for now, and see if we need to add some "robustness" later on without compromising anything.

JenySadadia: For POST node requests, there is a check to validate the parent (if the parent is in an available state) before creating child nodes in the API. This can result in an exception if the parent is invalid.
I have added exception handling specifically for `create_node` to handle this case.
I don't think the `runner` should be terminated if somehow the test node field for a specific revision is invalid.
As this can cause prevention of running tests (maybe with valid data) for other revisions.

JenySadadia:
Rebased this.

src/runner.py
mgalka: The call to the API is returning `HTTPException` with `400` status if the parent node is not in the available state.  I guess the runner shouldn't crash in a situation when this happens, but it'd be better to narrow down the error handling to `HTTPError` instead of silencing all possible errors with the generic `except Exception` clause.

JenySadadia: I agree with this.
Fixed the issue.

JenySadadia:
Tested OK locally by manually failing parent validation in the API:

```
kernelci-pipeline-runner | 01/02/2023 12:25:22 PM UTC [INFO] Tarball: http://172.17.0.1:8002/linux-kernelci-staging-mainline-staging-mainline-20230102.0.tar.gz
kernelci-pipeline-runner | 01/02/2023 12:25:22 PM UTC [INFO] Creating test node
kernelci-pipeline-runner | 400 Client Error: Bad Request for url: http://172.17.0.1:8001/node
kernelci-pipeline-runner | 01/02/2023 12:25:22 PM UTC [ERROR] Failed to schedule job for kver. Error: The node is unavailable to create child node: 63b2cd11ee6c506689c88359

```

mgalka (approved):
LGTM

PR #165, opened 2022-10-07
src/fstests: update fstests root node 

alexandrasp:
Now the template will be able to set the root node result as fail if any sub-test case failed.


gctucker:
No update since October so closing this now.  Also it's conflicting as the code has been rewritten since then, so I'm not sure this actually needs to be rebased or if the changes aren't applicable any more.

PR #170, opened 2022-11-07
GCE VM sharding

hardboprobot:
This changeset implements automatic sharding for test runs deployed to GCE, as well as a few other enhancements and fixes:

- new command-line parameters: `-g` to specify the test group(s) to run and `-c` to specify the fs configs per test
- extend the collection of results to properly report the results for all combinations of test configs
- fixing and improvement of artifacts cleanup
- automatic VM sharding, including querying of GCE project quotas. By default, the code will try to split the test configurations in as many different VM instances as possible. The `--max-shards` lets the user specify the maximum number of shards to use

### DIFF
--- a/src/fstests/fstests.jinja2
+++ b/src/fstests/fstests.jinja2
@@ -8,7 +8,9 @@
 XFSTESTS_BLD_PATH = '{{xfstests_bld_path}}'
 # TBD: For production, set the appropriate njobs number by default
 NJOBS             = '{{njobs if njobs is not none else 1}}'
-TESTCASE          = '{{testcase if testcase is not none else "smoke"}}'
+TESTCASE          = '{{testcase if testcase is not none else ""}}'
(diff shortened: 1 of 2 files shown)


gctucker (commented):
BTW It would be good to drop this `runner.py` file at some point in the near future as this was only meant to be a temporary solution to bootstrap things with `fstests`.  Adding support for running the tests via a runtime that works with GCE would be much better, like we have support for Kubernetes and shell tests currently (and soon LAVA like we already have with the old Jenkins stuff).

gctucker:
Note: Running KVM in containers isn't a great idea, and it's one of the main reasons why the PoC was running `kvm-xfstests` directly on the host (local shell).

Other than that, yes it seems like we can have templates for both use-cases or one template with parameters and then generate job definitions for either a shell or Kubernetes.

PR #172, opened 2022-11-11
Refactor class-based implementation of all the services

gctucker:
Refactor all the services to align their implementation using a class with common members and similar patterns for the main loop etc.  This should help consolidate things as a follow-up with a base class to group the common part and simplify each implementation.

### DIFF
--- a/src/timeout.py
+++ b/src/timeout.py
@@ -23,10 +23,10 @@ from logger import Logger
 class Timeout:
 
     def __init__(self, configs, args):
-        self._db_config = configs['db_configs'][args.db_config]
-        api_token = os.getenv('API_TOKEN')
-        self._db = kernelci.db.get_db(self._db_config, api_token)
         self._logger = Logger("config/logger.conf", "timeout")
+        db_config = configs['db_configs'][args.db_config]
+        api_token = os.getenv('API_TOKEN')
+        self._db = kernelci.db.get_db(db_config, api_token)
         self._poll_period = args.poll_period
 
     def _set_node_result(self, node):
@@ -104,6 +104,7 @@ class Timeout:
         self._logger.log_message(logging.INFO,
                                  "Checking timed-out nodes...")
         self._logger.log_message(logging.INFO, "Press Ctrl-C to stop.")
+        status = True
 
         try:
             while True:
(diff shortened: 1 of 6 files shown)


src/send_kcidb.py
```diff
@@ -24,19 +24,21 @@
...
-    def __init__(self, sub_parser, name):
-        super().__init__(sub_parser, name)
+    def __init__(self, configs, args):
         self._logger = Logger("config/logger.conf", "send_kcidb")
-
-    def __call__(self, configs, args):
         db_config = configs['db_configs'][args.db_config]
         api_token = os.getenv('API_TOKEN')
-        db = kernelci.db.get_db(db_config, api_token)
+        self._db = kernelci.db.get_db(db_config, api_token)
```
mgalka, line 33: Further `db` object calls in the script need to be changed to `self._db` i.e. lines: 66, 77, 78, 119 (64, 73, 74, 114)

gctucker: Ah yes, I think I'll just drop the refactoring of `send_kcidb.py` from this PR as I can't really test it.

src/timeout.py
```diff
@@ -130,8 +134,7 @@ class cmd_run(Command):
     ]
 
     def __call__(self, configs, args):
-        timeout = Timeout(configs, args)
-        timeout.run()
+        Timeout(configs, args).run()
```
JenySadadia, line 137:
```suggestion
        return Timeout(configs, args).run()
```

gctucker: That was actually fixed in #175 but I've fixed it here now anyway.

mgalka (approved):
LGTM

JenySadadia (changes requested):
A suggestion for `RunnerSingleJob._run_single_job`:
```
except Exception:
            self._logger.log_message(logging.ERROR, traceback.format_exc())
            return False
```

gctucker:
> A suggestion for `RunnerSingleJob._run_single_job`:
> 
> ```
> except Exception:
>             self._logger.log_message(logging.ERROR, traceback.format_exc())
>             return False
> ```

Ah yes, this is fixed already in #175 but I'll fix it here as well as it's more correct.

JenySadadia (approved):

PR #175, opened 2022-11-15
Add base Service class

gctucker:
Add a base `Service` class and start refactoring services to use it.  This is to consolidate the code and avoid duplicating the same pattern in each service.

Depends on #172

### DIFF
--- a/src/base.py
+++ b/src/base.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022 Collabora Limited
+# Author: Guillaume Tucker <guillaume.tucker@collabora.com>
+
+import logging
+import os
+import traceback
+
+import kernelci
+import kernelci.db
+
+from logger import Logger
+
+
+class Service:
+    """Common class for pipeline services"""
+
+    def __init__(self, configs, args, name):
+        self._name = name
+        self._logger = Logger("config/logger.conf", name)
+        self._db_config = configs['db_configs'][args.db_config]
+        api_token = os.getenv('API_TOKEN')
+        self._db = kernelci.db.get_db(self._db_config, api_token)
+
+    def _setup(self, args):
+        """Set up the service
+
+        Set up the service and return an arbitrary context object.  This will
+        then be passed again in `_run()` and `_stop()`.
+        """
+        return None
+
+    def _stop(self, context):
+        """Stop the service
+
+        This is where any resource allocated during _setup() should be
+        released.  Please note that if _setup() failed, _stop() is still called
+        so checks should be made to determine whether resources were actually
+        allocated.
+        """
+        pass
+
+    def _run(self, context):
+        """Implementation method to run the service"""
+        raise NotImplementedError("_run() needs to be implemented")
+
+    def run(self, args=None):
+        """Interface method to run the service
+
+        Run the service by first calling _setup() to set things up so that
+        _stop() can then be called upon termination to release any resources.
+        The _run() method is called in between and exceptions are handled
+        accordingly.  Any exception except KeyboardInterrupt will result in a
+        returned status of False.  The `args` optional arguments are for
+        providing extra options, typically parsed from the command line.
+        """
+        status = True
+        context = None
+
+        try:
+            context = self._setup(args)
+            status = self._run(context)
+        except KeyboardInterrupt:
+            self._logger.log_message(logging.INFO, "Stopping.")
+        except Exception:
+            self._logger.log_message(logging.ERROR, traceback.format_exc())
+            status = False
+        finally:
+            self._stop(context)
+
+        return status
--- a/src/notifier.py
+++ b/src/notifier.py
@@ -9,27 +9,29 @@
 import datetime
 import json
 import logging
-import os
 import sys
-import traceback
 
 import kernelci
 import kernelci.config
-import kernelci.db
 from kernelci.cli import Args, Command, parse_opts
 
-from logger import Logger
+from base import Service
 
 
-class Notifier:
+class Notifier(Service):
 
     def __init__(self, configs, args):
-        self._logger = Logger("config/logger.conf", "notifier")
-        db_config = configs['db_configs'][args.db_config]
-        api_token = os.getenv('API_TOKEN')
-        self._db = kernelci.db.get_db(db_config, api_token)
+        super().__init__(configs, args, 'notifier')
 
-    def run(self):
+    def _setup(self, args):
+        return self._db.subscribe('node')
+
+    def _stop(self, sub_id):
+        if sub_id:
+            self._db.unsubscribe(sub_id)
+        sys.stdout.flush()
+
+    def _run(self, sub_id):
         log_fmt = \
             "{time:26s}  {commit:12s}  {id:24}  " \
             "{state:9s}  {result:8s}  {name}"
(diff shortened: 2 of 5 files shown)


src/notifier.py
```diff
@@ -9,27 +9,31 @@
...
-class cmd_run(Command):
-    help = "Listen for events and report them on stdout"
-    args = [Args.db_config]
+class Notifier(Service):
 
-    def __init__(self, sub_parser, name):
-        super().__init__(sub_parser, name)
-        self._logger = Logger("config/logger.conf", "notifier")
+    def __init__(self, configs, args):
+        super().__init__(configs, args, 'nofitier')
```
mgalka, line 24: typo
```suggestion
        super().__init__(configs, args, 'notifier')
```

src/trigger.py
```diff
@@ -21,32 +19,16 @@
...
-    def __init__(self, sub_parser, name):
-        super().__init__(sub_parser, name)
-        self._logger = Logger("config/logger.conf", "trigger")
-        self._build_configs = None
-        self._db = None
+    def __init__(self, configs, args):
+        super().__init__(configs, args, 'trigger')
+        self._build_configs = configs['build_configs']
+        self._poll_period = int(args.poll_period)
+        self._force = args.force
```
mgalka, line 31: Wouldn't it be better to store these in the context?

src/notifier.py
```diff
@@ -9,27 +9,31 @@
...
 
-    def __call__(self, configs, args):
+    def _setup(self, ctx, args):
+        ctx.sub_id = self._db.subscribe('node')
+        return True
+
+    def _stop(self, ctx):
+        sub_id = getattr(ctx, 'sub_id', None)
+        if sub_id:
+            self._db.unsubscribe(ctx.sub_id)
```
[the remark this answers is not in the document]
gctucker: I hesitated between doing this and a dictionary.  I think I'll use a dictionary for the context instead.  In the end it's exactly the same, just a different syntax.  Also I was probably anticipating some code to be added to `Context` but there's no clear need for that at the moment.

As explained in the `_stop()` docstring, it's entirely possible that something in `_setup()` went wrong and things were not set up (for example, if subscribing to the API failed).  So the `_stop()` method needs to check whether the resources that might need to be freed have actually been allocated before freeing them.  Kind of like `if (ptr != NULL) free(ptr);`

src/trigger.py
gctucker: Yeah probably, it's a bit of a grey area.

src/notifier.py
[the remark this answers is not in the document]
mgalka: I personally like the idea of Context being a class not a plain dictionary. E.g. attribute access  seems a lot more clear than using indexes of the map. Maybe it's worth looking into implementing custom `__getattribute__()` in context that e.g. could return None or maybe call some function responsible for handling this situation if attribute is not present?  That could probably make the implementation of `_stop()` more simple and idiomatic.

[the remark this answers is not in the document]
gctucker: Well eventually I added a `data` dictionary to the `Context` class, see the latest version with 2 extra commits to rework this part.  What do you think?

src/trigger.py
gctucker: This was actually done in #172, this PR is not changing this part of the code.  The arguments used to be passed to the run method before the #172 refactoring so I guess it probably should be put back like it was before.  I'll do that as a follow-up PR as it's not really related to this one.

gctucker: Fixed in #176

gctucker:
@mgalka Hello, are you OK with this now?

src/base.py
```diff
@@ -0,0 +1,88 @@
...
+            self._data = data or {}
+            self._sub_id = sub_id
+
+        @property
+        def data(self):
+            return self._data
+
+        @property
+        def sub_id(self):
+            return self._sub_id
```
mgalka, line 33: How about adding `__gettattr__` to access attributes from `self._data` dictionary with a dotted notation e.g. 
```
context = Context(data={'key': 'value'})
context.key

```


```suggestion
    class Context:
        """Generic object to handle arbitrary context data"""
        def __init__(self, data=None, sub_id=None):
            self._data = data or {}
            self._sub_id = sub_id

        @property
        def data(self):
            return self._data

        @property
        def sub_id(self):
            return self._sub_id

        def __getattr__(self, name):
            return self._data.get(name)
```

mgalka (commented):
one small thing to consider

mgalka:
> @mgalka Hello, are you OK with this now?

@gctucker I made one suggestion if you find it fitting your idea we can add it, if not we can merge it as is.

src/base.py
gctucker: What's the point in doing that?

It means setting a value using `ctx.data[key] = value` and getting it either the same way or with `key.value`.  So I think it's a bit more complicated than `Context(object)` and I don't see the advantage.  Having just a dictionary attribute makes it entirely explicit and also avoids clashes with other attributes e.g. `.sub_id`.  So I know there are many ways to solve this, but what's the reason for choosing one over another?

mgalka: I guess `__setattr__()` implementation can always be added to avoid `ctx.data[key] = value` ;)

The `Context` object is some kind of an environment data container for a `Service` class.  I don't exactly understand why there is a need of explicitly having some of it being accessed differently than `sub_id`. Having said that I understand that `sub_id` is something which will be present in the majority of services and we want to have it added explicitly to the `Context`. So I thought `__gettatr__` could unify a way to access `sub_id` and other data from context.

Apart from that you can add logic to the `__getatr__` that can reduce the usage of 
```
 key = ctx.data 'key', None)
 if key:
      do_something(key)
```
pattern in the `_stop` method of `Service` derived classes.

It also easier to refactor the code which has

```python
ctx.key = value
```

then 

```python
ctx.data[key] = value

# or

if ctx.data.get(key):
   do_something()
```
as the dotted notation is less likely to confuse IDEs "Find usages..." or "Redactor..." features etc.

The `__gettatr__` is called only when the attribute wasn't found by `__getattribute__`, so if `sub_id` is present in the class it won't be looked for in the `Context.data`.

Having said all of that, as you have mentioned, there are many ways to solve this. I find the dotted notation less complex than retrieving data form the `ctx.data` dictionary, because of the reasons above. Maybe I'm also a bit inspired by the `Meta` internal class of the Django ORM models.

src/runner.py
```diff
@@ -105,11 +101,21 @@ def _cleanup_paths(self):
...
+        ctx = Service.Context()
         sub_id = self._db.subscribe_node_channel(filters={
             'name': 'checkout',
             'state': 'available',
         })
+        return Service.Context(sub_id=sub_id)
+
+    def _stop(self, ctx):
+        sub_id = getattr(ctx, 'sub_id', None)
+        if sub_id:
```
mgalka, line 114: I think that with current implementation it's guaranteed that `sub_id` is present and have a default value of None

```suggestion
        if ctx.sub_id:
```

mgalka (commented):
answered review comments

gctucker:
OK I'm completely confused by this whole thing, I'll just deal with it differently as it's taking way too much time for something entirely trivial.

mgalka:
> OK I'm completely confused by this whole thing, I'll just deal with it differently as it's taking way too much time for something entirely trivial.

It's hard to discuss with that. Feel free to merge the PR

mgalka (approved):

gctucker:
> > OK I'm completely confused by this whole thing, I'll just deal with it differently as it's taking way too much time for something entirely trivial.
> 
> It's hard to discuss with that. Feel free to merge the PR

:shrug: I just couldn't see any clear outcome from the discussions on this minor thing so I defaulted to the simplest solution.

PR #177, opened 2022-11-17
Add --startup-delay and --build-configs to trigger.py

gctucker:
Add `--startup-delay` and `--build-configs` command line arguments to `trigger.py`.  These are useful in practice respectully to wait for a short amount of time (heuristics) for all the services to be up and running, and to limit the build configurations to monitor.

Also add a default startup delay of 3 seconds for the trigger script (see commit description for more details).

Depends on #179

### DIFF
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -68,22 +68,27 @@ class Trigger(Service):
         }
         self._db.submit({'node': node})
 
-    def _iterate_build_configs(self, force):
+    def _iterate_build_configs(self, force, build_configs_list):
         for name, config in self._build_configs.items():
-            self._run_trigger(config, force)
+            if not build_configs_list or name in build_configs_list:
+                self._run_trigger(config, force)
 
     def _setup(self, args):
         return {
             'poll_period': int(args.poll_period),
             'force': args.force,
+            'build_configs_list': args.build_configs.split() or [],
         }
 
     def _run(self, ctx):
-        poll_period, force = (
-            ctx[key] for key in ('poll_period', 'force')
+        poll_period, force, build_configs_list = (
+            ctx[key] for key in (
+                'poll_period', 'force', 'build_configs_list'
+            )
         )
+
         while True:
-            self._iterate_build_configs(force)
+            self._iterate_build_configs(force, build_configs_list)
             if poll_period:
                 self._logger.log_message(
                     logging.INFO,
@@ -93,6 +98,7 @@ class Trigger(Service):
             else:
                 self._logger.log_message(logging.INFO, "Not polling.")
                 break
+
         return True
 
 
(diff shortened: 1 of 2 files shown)


src/trigger.py
```diff
@@ -75,28 +68,44 @@ def _run_trigger(self, build_config, force):
...
+        poll_period, force, build_configs_list, startup_delay = (
+            ctx.data[key] for key in (
+                'poll_period', 'force', 'build_configs_list', 'startup_delay'
+        ))
+
+        if startup_delay:
+            self._logger.log_message(
+                logging.INFO, f"Delay: {startup_delay}s"
+            )
+            time.sleep(startup_delay)
```
JenySadadia, line 96: How about considering docker container health check instead of using a random startup delay?
https://docs.docker.com/compose/compose-file/#healthcheck

gctucker: How would that solve the issue?

gctucker: To be clear, the delay here is to avoid timing issues between services.  It's a bit like SCSI drives starting to spin with a different arbitrary delay each to avoid a current surge when a server is booting... It's not to avoid an issue with the API not being available, which would cause the service to crash.

A more deterministic way of doing this would be for the trigger service to somehow wait for the tarball service to have started listening for checkout events before running its main loop.  However, this adds explicit dependencies between services which is a lot more problematic than an arbitrary delay at startup.

JenySadadia: We can declare dependencies for `trigger` service.
Health checks of other services can be defined and `service_healthy` tag can be used in `trigger` to make sure that other services are ready before starting the `trigger`.
The documentation I came across from above link:

"service_healthy: specifies that a dependency is expected to be “healthy” (as indicated by [healthcheck](https://docs.docker.com/compose/compose-file/#healthcheck)) before starting a dependent service."

"Compose implementations MUST wait for healthchecks to pass on dependencies marked with service_healthy. In the following example, db is expected to be “healthy” before web is created."

JenySadadia: This is just a suggestion. I haven't tried using this before.

JenySadadia (changes requested):
Please fix `pycodestyle` error.

src/trigger.py
gctucker: But we really don't want to do that, we should be able to run the trigger service without the tarball service.  Also it means adding dependency from the event producer on the consumers, which is very bad design.  The services need to remain decoupled, the pub/sub mechanism is not designed for point-to-point communication but generating events without any particular recipient.

gctucker: Healthchecks and dependencies would be nice to add to the API docker-compose though, like the example with the database as the API service depends on it.

JenySadadia: Yes, I am considering it for API.

gctucker:
@mgalka @JenySadadia Any further thoughts on this one?

mgalka (approved):

gctucker:
Note: #179 needs to be reviewed and merged before #177 and #178

PR #178, opened 2022-11-22
Add Service.log_* helpers and align test_report and timeout

gctucker:
Add `Service.log_*()` helper methods to simplify the syntax for log messages.  Update all services using the `Service` class accordingly.

Update `test_report.py` and `timeout.py` to use the `Service` class and align them with the other services.

Depends on #177

### DIFF
--- a/src/base.py
+++ b/src/base.py
@@ -7,7 +7,6 @@
 
 import logging
 import os
-import traceback
 
 import kernelci
 import kernelci.db
@@ -24,6 +23,10 @@ class Service:
         self._db_config = configs['db_configs'][args.db_config]
         api_token = os.getenv('API_TOKEN')
         self._db = kernelci.db.get_db(self._db_config, api_token)
+
+    @property
+    def log(self):
+        return self._logger
 
     def _setup(self, args):
         """Set up the service
@@ -64,9 +67,9 @@ class Service:
             context = self._setup(args)
             status = self._run(context)
         except KeyboardInterrupt:
-            self._logger.log_message(logging.INFO, "Stopping.")
+            self.log.info("Stopping.")
         except Exception:
-            self._logger.log_message(logging.ERROR, traceback.format_exc())
+            self.log.traceback()
             status = False
         finally:
             self._stop(context)
--- a/src/logger.py
+++ b/src/logger.py
@@ -9,6 +9,7 @@
 
 import logging
 import logging.config
+import traceback
 
 
 class Logger:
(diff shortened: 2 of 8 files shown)


src/base.py
```diff
@@ -0,0 +1,106 @@
...
+        self._logger.log_message(logging.WARNING, msg)
+
+    def _log_error(self, msg):
+        self._logger.log_message(logging.ERROR, msg)
+
+    def _log_critical(self, msg):
+        self._logger.log_message(logging.CRITICAL, msg)
+
+    def _log_traceback(self):
+        self._log_error(traceback.format_exc())
```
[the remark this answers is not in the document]
gctucker: Well not quite because then we would need to do `self.logger.log_info()` instead of `self.log_info()`.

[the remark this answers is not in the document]
JenySadadia: Well, what is wrong with `self.logger.log_info()`?
I think these methods are kind of abstractions specific to the logger. Hence, should belong to the `logger`.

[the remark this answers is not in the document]
mgalka: How about adding methods like `.warning()`, `.error()` to the `Logger` class and instantiating it as `Service._log` and then having calls like

`self._log.warning('Message')` ?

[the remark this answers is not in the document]
gctucker: Or even `self.log`. Yes I guess that would be equivalent in terms of ease of use and if that helps organising the code then fine.  OK I'll send an update with basically `sed s/log_/log./g`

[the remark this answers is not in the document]
gctucker: OK I've done that now.

src/logger.py
```diff
@@ -27,3 +28,21 @@ def __init__(self, config_path, name='root'):
     def log_message(self, log_level, msg):
```
JenySadadia, line 28: Since we have separate methods for all log levels now and the `log_message` method is only being accessed within the class, how about declaring it `protected`?
```suggestion
    def _log_message(self, log_level, msg):
```

src/test_report.py
```diff
@@ -112,47 +109,50 @@ def _get_report(self, root_node):
...
-            self._email_sender.create_and_send_email(
-                    subject, content
-                )
-        except Exception as err:
-            self._logger.log_message(logging.ERROR, err)
-
-    def loop(self):
-        """Method to execute in a loop"""
-        sub_id = self._db.subscribe_node_channel(filters={
+        self._email_sender.create_and_send_email(subject, content)
```
JenySadadia, line 112: `try-except` block shouldn't be dropped from here. As this can result in potential exceptions from SMTP lib.
Or else, it should be handled in `email_sender` module and a resultant status can be received here.
```suggestion
        try:
            self._email_sender.create_and_send_email(
                    subject, content
                )
        except Exception as err:
            self.log.error(err)
```

JenySadadia: I see this is already been handled in `Service.run` method.

gctucker: Exactly, which is why it can be dropped here.

src/logger.py
gctucker: That's not necessary, we need to get changed merged quicker and focus on the real issues.  Feel free to send a PR with follow-up improvements if you want, I think we could just merge Logger into Service.

JenySadadia: Well, In my opinion, this change is more relevant to this PR as logging methods specific to log levels have been introduced here.
Hence, it is more suitable if we can have the suggested change along with this PR itself.
Otherwise, feel free to merge the PR.

JenySadadia (approved):

src/logger.py
gctucker: It's still perfectly fine to do `Logger.log_message()` even if we have other methods that make the syntax simpler.

PR #179, opened 2022-11-24
src/trigger.py: drop Context object

gctucker:
The Context object was in a previous implementation which got reverted.  Fix this by returning a plain dictionary instead.

Fixes: ae0529889d08 ("src/trigger.py: use args in context rather than in constructor")


mgalka (approved):

gctucker:
Fixed the commit message, the `Fixes:` trailer format was a bit broken.

PR #181, opened 2022-11-25
src/trigger.py: fix issue with empty --build-configs

gctucker:
Fix the case where no --build-configs optional argument is provided using an empty string instead of None which can't be split.

Fixes: 81387ab4786d ("src/trigger.py: add --build-configs optional argument")


gctucker:
@mgalka Feel free to merge it, unless you're waiting for someone else's approval?

PR #182, opened 2022-11-25
HACK tarball holdoff set to 1min

gctucker:
This is to cover the scenario where the tarball node enters Closing state while some tests are still running.


gctucker:
Dropping this now that the timeout service rework has been merged.  Ideally there should be proper end-to-end tests which spin up a local instance to test different scenarios etc.

PR #187, opened 2022-11-26
src/test_report.py: make the SMTP args optional

gctucker:
Make the --smtp-host (and --smtp-port) arguments optional so the service can be run without them to just print the email report text on stdout.  This is particularly useful during development to have local reports without spamming any email.


mgalka:
I think it's ok to merge it. We may also consider adding a `--dry-run` parameter or something similar to explicitly mark that e-mail won't be sent.

mgalka (approved):

PR #188, opened 2022-11-28
T39195 Refactor kcidb bridge service

JenySadadia:
Refactor KCIDB bridge service for the below improvements:

- use `Service` as a base class
- code refactoring by defining methods
- define configurations options for KCIDB project id and topic

### DIFF
--- a/config/kernelci.conf
+++ b/config/kernelci.conf
@@ -20,7 +20,10 @@ output: /home/kernelci/output
 
 [notifier]
 
-[kcidb]
+[send_kcidb]
+kcidb_topic_name: playground_kcidb_new
+kcidb_project_id: kernelci-production
+origin: kernelci
 
 [test_report]
 
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -37,6 +37,7 @@ class KCIDBBridge(Service):
                 'name': 'checkout',
                 'state': 'done',
             }),
+            'origin': args.origin,
         }
 
     def _stop(self, context):
@@ -69,8 +70,8 @@ class KCIDBBridge(Service):
                 'builds': [],
                 'checkouts': [
                     {
-                        'id': f"kernelci:{node['_id']}",
-                        'origin': 'kernelci',
+                        'id': f"{context['origin']}:{node['_id']}",
+                        'origin': context['origin'],
                         'tree_name': node['revision']['tree'],
                         'git_repository_url': node['revision']['url'],
                         'git_commit_hash': node['revision']['commit'],
@@ -103,6 +104,10 @@ class cmd_run(Command):
         {
             'name': '--kcidb-project-id',
             'help': "KCIDB project ID",
+        },
+        {
+            'name': '--origin',
+            'help': "CI system identifier",
         },
     ]
 
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -8,126 +8,110 @@
 # Copyright (C) 2022 Collabora Limited
 # Author: Jeny Sadadia <jeny.sadadia@collabora.com>
 
+"""KCIDB bridge service"""
+
 import datetime
-import logging
-import os
 import sys
-import traceback
 
 import kernelci
 import kernelci.db
-from kernelci.config import load
+import kernelci.config
 from kernelci.cli import Args, Command, parse_opts
 from kcidb import Client
 import kcidb
 
-from logger import Logger
+from base import Service
+
+
+class KCIDBBridge(Service):
+    def __init__(self, configs, args, name):
+        super().__init__(configs, args, name)
+
+    def _setup(self, args):
+        return {
+            'client': Client(
+                project_id=args.kcidb_project_id,
+                topic_name=args.kcidb_topic_name
+            ),
+            'sub_id': self._db.subscribe_node_channel(filters={
+                'name': 'checkout',
+                'state': 'done',
+            }),
+        }
+
+    def _stop(self, context):
+        if context['sub_id']:
+            self._db.unsubscribe(context['sub_id'])
+
+    def _send_revision(self, client, revision):
+        if kcidb.io.SCHEMA.is_valid(revision):
+            return client.submit(revision)
+        self.log.error("Aborting, invalid data")
+
+    @staticmethod
+    def _set_timezone(created_timestamp):
+        created_time = datetime.datetime.fromisoformat(created_timestamp)
+        if not created_time.tzinfo:
+            tz_utc = datetime.timezone(datetime.timedelta(hours=0))
+            created_time = datetime.datetime.fromtimestamp(
+                created_time.timestamp(), tz=tz_utc)
+        return created_time.isoformat()
+
+    def _run(self, context):
+        self.log.info("Listening for events... ")
+        self.log.info("Press Ctrl-C to stop.")
+
+        while True:
+            node = self._db.receive_node(context['sub_id'])
+            self.log.info(f"Submitting node to KCIDB: {node['_id']}")
+
+            revision = {
+                'builds': [],
+                'checkouts': [
+                    {
+                        'id': f"kernelci:{node['_id']}",
+                        'origin': 'kernelci',
+                        'tree_name': node['revision']['tree'],
+                        'git_repository_url': node['revision']['url'],
+                        'git_commit_hash': node['revision']['commit'],
+                        'git_repository_branch': node['revision']['branch'],
+                        'start_time': self._set_timezone(node['created']),
+                        'patchset_hash': '',
+                        'misc': {
+                            'submitted_by': 'kernelci-pipeline'
+                        },
+                    }
+                ],
+                'tests': [],
+                'version': {
+                    'major': 4,
+                    'minor': 0
+                }
+            }
+            self._send_revision(context['client'], revision)
+        return True
 
 
 class cmd_run(Command):
     help = "Listen for events and send them to KCDIB"
-    args = [Args.db_config]
-
-    def __init__(self, sub_parser, name):
-        super().__init__(sub_parser, name)
-        self._logger = Logger("config/logger.conf", "send_kcidb")
+    args = [
+        Args.db_config,
+        {
+            'name': '--kcidb-topic-name',
+            'help': "KCIDB topic name",
+        },
+        {
+            'name': '--kcidb-project-id',
+            'help': "KCIDB project ID",
+        },
+    ]
 
     def __call__(self, configs, args):
-        db_config = configs['db_configs'][args.db_config]
-        api_token = os.getenv('API_TOKEN')
-        db = kernelci.db.get_db(db_config, api_token)
-
-        if not os.getenv('GOOGLE_APPLICATION_CREDENTIALS'):
-            self._logger.log_message(logging.ERROR,
-                                     "No GOOGLE_APPLICATION_CREDENTIALS \
-environment variable")
-            return False
-
-        topic_name = os.getenv('KCIDB_TOPIC_NAME')
-        if not topic_name:
-            self._logger.log_message(logging.ERROR, "No KCIDB_TOPIC_NAME \
-environment variable")
-            return False
-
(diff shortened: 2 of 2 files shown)


JenySadadia:
@mgalka  Can you please confirm that I have used the correct values for KCIDB configurations:
https://github.com/kernelci/kernelci-pipeline/pull/188/files#diff-f1c9e56ef349bb697df7f6aab238aee7e28615282824ad4c8862335adbaa69b7

src/send_kcidb.py
```diff
@@ -8,126 +8,122 @@
...
+
+    def _run(self, sub_id):
+        self.log.info("Listening for events... ")
+        self.log.info("Press Ctrl-C to stop.")
         sys.stdout.flush()
 
+        client = self._kcidb_connect()
+        if not client:
+            self.log.error("Failed to create client connection to KCIDB")
+            return False
```
gctucker, line 78: I think it would be better to do this in `self._setup()`.  Then any disconnection can be handled when an exception happens during the main loop, or the loop could be restarted without redoing the KCIDB connection etc.

config/kernelci.conf
```diff
@@ -20,7 +20,9 @@ output: /home/kernelci/output
 
 [notifier]
 
-[kcidb]
+[send_kcidb]
+kcidb_topic_name: playground_kcidb_new
+kcidb_project_id: kernelci-production
```
gctucker, line 25: Not sure this is the right project id, first of all the pipeline is actually running on staging.  Maybe `kernelci-api` would be better at this stage to differentiate it from the data sent by the current actual production / staging setup which uses `kernelci_backend`.

JenySadadia: Updated the project ID.

src/send_kcidb.py
JenySadadia: Yes, makes sense.

src/send_kcidb.py
```diff
@@ -8,126 +8,116 @@
...
-                    }],
-                    "tests": [],
-                    "version": {
-                        "major": 4,
-                        "minor": 0
-                    }
-                }
-                self.send_revision(client, revision)
+    def _stop(self, sub_id):
+        if sub_id:
```
gctucker, line 46: I guess the KCIDB Client object doesn't have any means of disconnecting?

If it's held in the context object then it will go out of scope when the loop exits and any clean-up will be done automatically then.

src/send_kcidb.py
```diff
@@ -8,126 +8,116 @@
...
-            self._logger.log_message(logging.INFO, "Stopping.")
-        except Exception:
-            self._logger.log_message(logging.ERROR, traceback.format_exc())
-        finally:
-            db.unsubscribe(sub_id)
+    def _send_revision(self, client, revision):
+        if self._validate_revision(revision):
+            return client.submit(revision)
+        self.log.error("Aborting, invalid data")
+        sys.stdout.flush()
```
gctucker, line 58: I think we can drop all the `.flush()` calls now.

src/send_kcidb.py
JenySadadia: I checked `kcidb.Client` implementation. It doesn't have any external method to disconnect.

src/send_kcidb.py
```diff
@@ -8,126 +8,116 @@
...
+        self._kcidb_topic_name = args.kcidb_topic_name
+        self._kcidb_project_id = args.kcidb_project_id
+        self._client = None
 
-    def __init__(self, sub_parser, name):
-        super().__init__(sub_parser, name)
-        self._logger = Logger("config/logger.conf", "send_kcidb")
+    def _kcidb_connect(self):
+        return Client(project_id=self._kcidb_project_id,
+                      topic_name=self._kcidb_topic_name)
```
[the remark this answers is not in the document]
gctucker: I mean, the `Client` object could be created directly in `_setup()`.  Also the KCIDB topic and project ID could be read from the `args` there rather than storing them as instance variables in the object.  That's a bit more of a grey area but I think it's better to set everything up in `_setup()` as much as possible.

src/send_kcidb.py
```diff
@@ -8,126 +8,113 @@
...
 
-        except KeyboardInterrupt:
-            self._logger.log_message(logging.INFO, "Stopping.")
-        except Exception:
-            self._logger.log_message(logging.ERROR, traceback.format_exc())
-        finally:
-            db.unsubscribe(sub_id)
+    @staticmethod
+    def _validate_revision(revision):
+        return kcidb.io.SCHEMA.is_valid(revision)
```
gctucker, line 52: I think this method can be dropped, we can just call `.is_valid()` directly.

src/send_kcidb.py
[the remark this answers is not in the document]
JenySadadia: Having separate functions increases the code's reusability.
That's why I created the `kcidb_connect` method and also verified if the connection is successful in the method.
Since the method just has a single line of code now, it won't make much of a difference and the method can be dropped.

gctucker:
OK let's just wait for some staging results now to check it's working as expected.

mgalka:
> @mgalka Can you please confirm that I have used the correct values for KCIDB configurations? https://github.com/kernelci/kernelci-pipeline/pull/188/files#diff-f1c9e56ef349bb697df7f6aab238aee7e28615282824ad4c8862335adbaa69b7

These look good

gctucker:
There's a problem though:
```
google.api_core.exceptions.NotFound: 404 Requested project not found or user does not have access to it (project=kernelci-api). Make sure to specify the unique project identifier and not the Google Cloud Console display name.
```

JenySadadia:
> There's a problem though:
> 
> ```
> google.api_core.exceptions.NotFound: 404 Requested project not found or user does not have access to it (project=kernelci-api). Make sure to specify the unique project identifier and not the Google Cloud Console display name.
> ```

I have modified project id and added `misc` field to identify submitter.

JenySadadia:
As per our discussion with @spbnick, I have changed `origin` to `kernelci-api` to segregate staging results sent from the new API.

gctucker:
I've merged the kernelci-deploy PR so let's wait for the next staging run to confirm it's sending the data as expected.

JenySadadia:
Seems like an issue with submission:
```
today at 06:48:1812/02/2022 01:18:18 AM UTC [INFO] Submitting node to KCIDB: 638944495511e7db63a0f99f
today at 06:48:1812/02/2022 01:18:18 AM UTC [ERROR] Aborting, invalid data
```
Added a debug print to verify the revision data.

gctucker:
OK great, at least it's running and talking to KCIDB.  I guess the data being sent doesn't match the required schema.

src/send_kcidb.py
```diff
@@ -8,126 +8,109 @@
...
+
+        while True:
+            node = self._db.receive_node(context['sub_id'])
+            self.log.info(f"Submitting node to KCIDB: {node['_id']}")
+
+            revision = {
+                'builds': [],
+                'checkouts': [{
+                    'id': f"kernelci:{node['_id']}",
+                    'origin': 'kernelci-api',
```
gctucker, line 72: Note: I think it would be good to have a `--origin` command line option, as a follow-up.

JenySadadia: Let me do it in this PR itself as `origin` has been added in this one.

gctucker: The `origin` field was already there as `kernelci`, but yes this PR is changing it.  I guess we can leave `kernelci` by default but set `kernelci-api` in the staging configuration file in `kernelci-deploy`.

gctucker: Ah I see you've already added it.  It might have been better as a separate commit as that's more than just refactoring the code to use the Service base class.

JenySadadia: Yes, I'll add a separate commit.

JenySadadia: Also, corrected `checkouts.id` field to use `origin` in the value to fix invalid data error.

JenySadadia:
Working on staging:
```
today at 06:34:4212/05/2022 01:04:42 AM UTC [INFO] Submitting node to KCIDB: 638d359a5511e7db63a0f9ae
today at 06:34:4212/05/2022 01:04:42 AM UTC [DEBUG] Submitting revision: {'builds': [], 'checkouts': [{'id': 'kernelci:638d359a5511e7db63a0f9ae', 'origin': 'kernelci', 'tree_name': 'kernelci', 'git_repository_url': 'https://github.com/kernelci/linux.git', 'git_commit_hash': '4009eec93bfa88b8da68864e8c3c65808d437eec', 'git_repository_branch': 'staging-mainline', 'start_time': '2022-12-05T00:04:42.061000+00:00', 'patchset_hash': '', 'misc': {'submitted_by': 'kernelci-pipeline'}}], 'tests': [], 'version': {'major': 4, 'minor': 0}}
```

The only issue is the value of `origin`. Its value from settings file will be overridden by the `default` value of the command line option.
Fixed it for `origin` to be `kernelci-api`.

gctucker:
> The only issue is the value of `origin`. Its value from settings file will be overridden by the `default` value of the command line option.
> Fixed it for `origin` to be `kernelci-api`.

If we can't set the value in the settings file due to the default value in the command line argument definition, then we just need to remove the default value from the command line argument definition.  I guess it makes sense that it takes precedence as the code treats the default command line argument value as a value passed by the user, which is meant to override the values in the settings file.

We can deal with a default value in the application code outside of the argument definition, to still set it to `kernelci` when `args.origin` is empty.

JenySadadia:
> > The only issue is the value of `origin`. Its value from settings file will be overridden by the `default` value of the command line option.
> > Fixed it for `origin` to be `kernelci-api`.
> 
> If we can't set the value in the settings file due to the default value in the command line argument definition, then we just need to remove the default value from the command line argument definition. I guess it makes sense that it takes precedence as the code treats the default command line argument value as a value passed by the user, which is meant to override the values in the settings file.
> 
> We can deal with a default value in the application code outside of the argument definition, to still set it to `kernelci` when `args.origin` is empty.

Or else we can keep the default value in the command line argument definition for staging and we can drop it later. We anyway have to use `origin:kernelci` when it gets deployed to production.

gctucker:
How about setting `kernelci_api` in the `staging.kernelci.org.conf` file and no default value?  That way we won't have to change the code for production or if anyone else wants to set up a pipeline and send data to KCIDB.

JenySadadia:
> How about setting `kernelci_api` in the `staging.kernelci.org.conf` file and no default value? That way we won't have to change the code for production or if anyone else wants to set up a pipeline and send data to KCIDB.

Seems a better idea. I will keep the origin `kernelci` in settings file for main branch and `kernelci_api` for staging.

JenySadadia:
Created https://github.com/kernelci/kernelci-deploy/pull/41

JenySadadia:
Tested OK on staging with `origin:kernelci_api`:

KCIDB bridge:
```
today at 06:44:1312/07/2022 01:14:13 AM UTC [INFO] Submitting node to KCIDB: 638fdad44f2e91321f7c31ff
today at 06:44:1312/07/2022 01:14:13 AM UTC [DEBUG] Submitting revision: {'builds': [], 'checkouts': [{'id': 'kernelci_api:638fdad44f2e91321f7c31ff', 'origin': 'kernelci_api', 'tree_name': 'kernelci', 'git_repository_url': 'https://github.com/kernelci/linux.git', 'git_commit_hash': 'e3ab23ecb5c8f2b0bcd91e48f152d7b3f9ac6744', 'git_repository_branch': 'staging-stable', 'start_time': '2022-12-07T00:14:12.429000+00:00', 'patchset_hash': '', 'misc': {'submitted_by': 'kernelci-pipeline'}}], 'tests': [], 'version': {'major': 4, 'minor': 0}}
```

Link to [KCIDB dashboard](https://kcidb.kernelci.org/d/checkout/checkout?orgId=1&var-dataset=playground_kcidb_01&var-origin=kernelci_api&var-build_architecture=All&var-build_config_name=All&var-id=kernelci_api:638fdad44f2e91321f7c31ff)

gctucker (approved):

PR #190, opened 2022-12-02
xfstests pipeline integration

hardboprobot:
This supersedes https://github.com/kernelci/kernelci-pipeline/pull/170 and https://github.com/kernelci/kernelci-pipeline/pull/171
and depends on https://github.com/kernelci/kernelci-core/pull/1565


config/pipeline.yaml
```diff
@@ -79,8 +79,25 @@ labs:
 
 test_plans:
 
-  fstests:
-    pattern: 'fstests.jinja2'
+  xfstests-base: &xfstests-base
```
gctucker, line 82: The test suite really is called `fstests` now.

config/pipeline.yaml
```diff
@@ -79,8 +79,25 @@ labs:
...
+    pattern: 'xfstests.jinja2'
+    params: &xfstests-base-params
+      njobs: 3
+      gce: True
```
gctucker, line 86: I think this should ideally be handled at the template / runtime level like kunit can be run either in a local shell or in k8s. We should ideally be able to run fstests in a local KVM or GCE with the same test plan name. Maybe it's not practical, but that's the idea with the separation between test plan definitions and runtime environments.

config/pipeline.yaml
hardboprobot: ooi, why is that? The original tests have always been called `xfstests` and this implementation is tightly coupled to that, it doesn't run generic filesystem tests, but the actual xfstests.

config/pipeline.yaml
hardboprobot: That's fine by me, but then what do you think it's the best place for the user to select whether to run this locally or on GCE? a command line option?
Can you give me some details about how the shell/k8s runtime is selected for kunit?

gctucker: That's currently done with `runner.py --lab-config`, except the term "lab" is being abused so the plan is now to replace that with "runtime".  Then a "runtime" can be anything that can generate the template and run it, e.g. a shell, local Docker, k8s, GCE, LAVA etc.  Some test plans might not be compatible with some runtimes, like you can't boot a Raspberry Pi in Kubernetes and you shouldn't run a kernel build on a Raspberry Pi, in which case some missing template layers might make it fail or it won't run when trying to submit it.  But at least the tools would provide a generic mechanism that's agnostic of the type of job and where it's meant to run.  You can already do any combination of kver / kunit tests with shell / k8s runtimes.

gctucker:
@hardboprobot Looks like this needs to be rebased?

config/pipeline.yaml
hardboprobot: Strictly speaking, since this uses the xfstests-bld scripts underneath for both KVM and GCE runtime options, this always uses `shell` as a runtime.
If you're suggesting that we implement the GCE variant as an independent and test-agnostic runtime (which I agree it's something that will eventually be needed), then... well, I'd have to rewrite the whole thing from scratch and re-implement the xfstests-gce logic.

Writing a GCE runtime is something that should be properly thought out and defined somewhere else anyway, but for now just let me know how you want to proceed with this PR

gctucker: No that's not what I'm suggesting.  So I don't know how to implement things in this particular case, maybe with a particular runtime for fstests e.g. fstests-gce and when using shell it uses kvm?

hardboprobot: Ok, so the way it works now is that there's a specific xfstests (or fstests if we want to call it that way) "runtime". By "runtime" here I mean a template in kernelci-pipeline/config/runtime, btw.

This python template has a bunch of common code for handling xfstests-bld and then some parts that are specific to kvm-xfsests and others for gce-xfstests. For instance, running a test using gce-xfstests require a few specific extra steps such as:

  - Uploading the kernel image to a Google Storage bucket
  - Check the GCE project quotas
  - Poll the VM status for completion
  - Fetch back the results

But some other things like configuring and building the kernel, and parsing the results are common between the two variants.

Now, both variants are architecturally the same from a code-generation and running point of view: they're python code, generated from a template, output to a file and run in the local host where it will then call the appropriate xfstests-bld scripts (kvm-xfstests and gce-xfstests) among other things. So while I can understand your reasoning about wanting the flexibility to run any test plan on any available "lab", I can't find a sound reason to split this into two technically equivalent runtimes.

Given the current status of things, I don't think specifying the "run mode" for this tests in the test plan config is a terrible solution. Not ideal but not terrible. We lose the ability to select kvm or gce in the command line, but adding a testcase-specific command line flag for that is a worse option IMO. The ideal scenario of having GCE as a selectable `lab-config` that can seamlessly run the code templates is not here yet.

src/runner.py
```diff
@@ -187,7 +197,28 @@ def run(self, args):
...
+        },
+        {
+            'name': '--ssh-port',
+            'help': "Storage SSH port number",
+            'type': int,
+        },
+        {
+            'name': '--ssh-user',
+            'help': "Storage SSH user name",
+        },
```
JenySadadia, line 220: Since these are the common command line arguments for `cmd_loop` and `cmd_run`, maybe we can store them in a separate list and use it to extend `opt_args` to drop the number of duplicate lines in the code.

src/runner.py
```diff
@@ -80,7 +89,8 @@ def _schedule_test(self, checkout_node, plan, device):
         self._logger.log_message(logging.INFO, "Creating test node")
         node = self._create_node(checkout_node, plan)
 
-        tmp = tempfile.TemporaryDirectory(dir=self._output)
+        tstamp = datetime.datetime.now().strftime('%Y%m%d_%H%M%S-')
+        tmp = tempfile.TemporaryDirectory(prefix=tstamp, dir=self._output)
```
JenySadadia, line 93: Out of curiosity, why do we need this prefix for the temporary directory?

JenySadadia:
@hardboprobot Do we have test results from staging?

src/runner.py
hardboprobot: I found it informative and useful while debugging to prepend a timestamp to the temp dir, although it's not really necessary.

src/runner.py
hardboprobot: Maybe something to do as a follow up after the merge?

src/runner.py
JenySadadia: I was just asking. If it is helpful, we can keep it.

src/runner.py
JenySadadia: Yes, that's fine.

JenySadadia (approved):

src/runner.py
```diff
@@ -31,6 +33,10 @@ def __init__(self, configs, args):
         runtime_config = configs['labs'][args.lab_config]
         self._runtime = kernelci.lab.get_api(runtime_config)
         self._output = args.output
+        self._ssh_host = args.ssh_host
+        self._ssh_key = args.ssh_key
+        self._ssh_port = args.ssh_port
+        self._ssh_user = args.ssh_user
```
gctucker, line 39: There's now a `kernelci.storage` package for handling uploads.  So please drop this and use it instead.  In fact `runner.py` and job templates will be updated to use it anyway, so maybe you can rebase this PR once it's done.  If you drop the `xfstests testplan integration` patch then I think this PR could be merged more easily as it would be all contained within the `fstests` directory.

hardboprobot: Are we still moving on with this? I had practically assumed this was going to be dropped altogether.

If there's a real plan to merge this I could give this higher priority (definitely not now), but at this point a big chunk of this might need to be rewritten and rebased to accomodate it to other changes, and it wouldn't be the first time that I have to rebase it because it has been left unmerged for too long. Note that this PR predates the `kernelci.storage` package.

gctucker: > Are we still moving on with this?

You mean running fstests?  Well I suppose so :)

gctucker:
It's been about 6 months now and as the code is still changing fast I think we should close this and start again when things are more stable.  We have just started running actual jobs with KUnit and kernel builds in k8s and baseline on QEMU in LAVA.  Adding Runtime implementations for on-demand VMs is going to be needed sooner or later, and at that point I think it will make a lot of sense to revive the whole fstests topic.

Issue #192, opened 2022-12-05
Use timeout__lt etc. to optimise timeout service


gctucker:
The timeout service queries the API regularly for any nodes that have reached their `timeout` or `holdoff` time.  This can be optimised by using the `timeout__lt` or `holdoff__lt` operators so that the API only returns the nodes that will timeout within the next polling period.  If there are any, then the time to sleep can be adjusted to deal with them earlier.  Nodes that have already reached their timeout / holdoff time will also be returned and dealt with from the same response.

PR #194, opened 2022-12-09
T39465 Optimize timeout service with comparison operators

JenySadadia:
Fixes https://github.com/kernelci/kernelci-pipeline/issues/192

- src/timeout: use `timeout__lt` filter to get timed-out nodes
- src/timeout: use `holdoff__lt` filter to get nodes with lapsed holdoff
- src/timeout: drop logic to check pending and available nodes
- src/timeout: add `_setup` method to `TimeoutService`

### DIFF
--- a/src/timeout.py
+++ b/src/timeout.py
@@ -96,7 +96,10 @@ class Holdoff(TimeoutService):
         super().__init__(configs, args, 'timeout-holdoff')
 
     def _get_available_nodes(self):
-        nodes = self._db.get_nodes({'state': 'available'})
+        nodes = self._db.get_nodes({
+            'state': 'available',
+            'holdoff__lt': datetime.isoformat(datetime.utcnow()),
+        })
         return {node['_id']: node for node in nodes}
 
     def _check_available_nodes(self, available_nodes):
--- a/src/timeout.py
+++ b/src/timeout.py
@@ -103,30 +103,22 @@ class Holdoff(TimeoutService):
         return {node['_id']: node for node in nodes}
 
     def _check_available_nodes(self, available_nodes):
-        now = datetime.utcnow()
-        sleep = timedelta(seconds=self._poll_period)
         timeout_nodes = {}
         closing_nodes = {}
         for node_id, node in available_nodes.items():
-            holdoff = datetime.fromisoformat(node['holdoff'])
-            if now > holdoff:
-                running = self._count_running_child_nodes(node_id)
-                if running:
-                    closing_nodes.update(
-                        self._get_child_nodes_recursive(node, 'available')
-                    )
-                    closing_nodes[node_id] = node
-                else:
-                    timeout_nodes.update(
-                        self._get_child_nodes_recursive(node)
-                    )
-                    timeout_nodes[node_id] = node
+            running = self._count_running_child_nodes(node_id)
+            if running:
+                closing_nodes.update(
+                    self._get_child_nodes_recursive(node, 'available')
+                )
+                closing_nodes[node_id] = node
             else:
-                self.log.debug(f"{node_id} holdoff left: {holdoff - now}")
-                sleep = min(sleep, holdoff - now)
+                timeout_nodes.update(
+                    self._get_child_nodes_recursive(node)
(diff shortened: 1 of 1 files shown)


src/timeout.py
```diff
@@ -70,28 +74,22 @@ def __init__(self, configs, args):
...
 
         while True:
-            pending_nodes = self._get_pending_nodes()
-            sleep_time = self._check_pending_nodes(pending_nodes)
-            sleep(sleep_time)
+            pending_nodes = self._get_pending_nodes({
+                'timeout__lt': datetime.isoformat(datetime.utcnow())
+            })
+            self._check_pending_nodes(pending_nodes)
+            sleep(ctx['poll_period'])
```
gctucker, line 92: The reason why the sleep period was dynamic was to be able to flag timeouts earlier than the sleep period, and in turn allow for a longer sleep period.  So we could have a poll period of 5min and still handle nodes that timeout in 30s.  I think this rework is losing the ability to do this.

JenySadadia: I don't think we need that ability anymore as we are only fetching timed-out nodes with `timeout__lt`.
It is possible to flag timeouts earlier if we have the information of the node that will timeout in the near future. Basically, we need to get all the pending nodes as per the previous implementation for that.
The filter helps to avoid getting all the pending nodes and then checking if the node is timed-out.

gctucker: I don't quite understand how this is working now.  Are you fetching the nodes that have timed out during the past polling period?  So if the polling period is set to 5min, the nodes may only actually be timed out up to 5min later?  If that's the case then I don't think we need comparison operators, as getting all the nodes not in done state with a timeout in the past would return the same thing.

JenySadadia: I have used filter `'timeout__lt': datetime.isoformat(datetime.utcnow()` to get timed out nodes.
It means we are sending the request to get nodes whose timeout is less that the current time of execution.
This will ensure that the node is indeed timed out and we don't need to check anything else for confirmation.

gctucker: OK, so if the polling period is 5min then the node could be timed out up to 5min later than its timeout time.

JenySadadia: Yes, that can be the scenario.
I am not sure about the side effects you are concerned about in that case.
We can always reduce the poll period to get over this issue because if we are using `timeout__lt` filter, we won't be able to get nodes that will timeout before the next iteration.

gctucker: My point was that we can have a long sleep period by default, and by querying when some nodes are going to timeout we can sleep a shorter amount of time to wake up when they've just timed out.  Combined with the ability to receive events, I think there are things we can do to have very few API calls and still handle timed out nodes without a long delay.

So based on this, the concerns I have is that if we use a polling period of 5min and several jobs in the chain hit a timeout they will each be adding 2.5min on average which in some cases can lead to something significant.  If we reduce the polling period then we reduce the API bandwidth.  I think with the current tests, we can easily have a 1min polling period but it's still not great when the pipeline is idle for a couple of hours and the timeout services keep sending tends of queries every minute.  It's not a real blocking issue though, and we'll have different scenarios in production which is what matters more.

JenySadadia: Okay. That makes sense.
So, now these are the two scenarios:
- By using `timeout__lt`, we can get only pending nodes. That will be no need to check manually for timed-out nodes but can not make sleep time dynamic
- By getting all the nodes, we need to externally check for timed-out nodes, and sleep time can be dynamic
The decision should be made by taking into account the possible future scenarios with production.
I am not sure what is the best thing to do here. What do you suggest then?

JenySadadia: Hello @gctucker, a gentle ping for your suggestion here.

gctucker: Sorry I don't have much time to look into this now as it's pretty low priority and I'm not sure what's the best way forward to be honest.

gctucker (approved):
OK I think this has been verified on staging so merging now.  If we need to tweak things further as we start running more jobs then we can have follow-up PRs to rework the logic.

Issue #191, opened 2022-12-05
Use Enum for node states


gctucker:
The states that a Node object can go through are simply defined as strings.  Using an Enum would provide some syntaxic definition of the states and make it easier to share the same definition between the API and the client-side services.  There is already this in `api.models` which may be reused on the client sides if we can import it somehow:

```python3
class StateValues(enum.Enum):
```

At the moment, the API package is not installed in client-side services so this can't be just imported as-is.  The advantage of having it separate is that it guarantees a clear distinction between the exposed web API and its internal implementation.  One way to solve this could be to have another shared package, but that's not really practical either.  We could just recreate an enum on the client side, but it's not clear whether that's the best approach either as they would need to be manually kept in sync which is likely to fail.

JenySadadia:
How about creating an endpoint to retrieve available node states from API?
That way the pipeline will be in sync with the API.

gctucker:
I think that would work if it was combined with an API version, in production.  While we're still in development phase we can still implement this and have an empty API version or 0.0 until we start enforcing the version.  My point is that services can't just read the list of state names and adjust their implementation.  For example if a new state was introduced in a non-backwards compatible way the version check would ensure older clients would not use it and report an error.  It's not like user data such as the kernel revisions etc. which can be anything.

Then yes the API could return some JSON data structure which could be used to construct an Enum I suppose.  @mgalka Do you see any implementation details with a dynamic enum definition?  Maybe the state names would be fixed in the class but the strings they map to would be received from the API in an ordered list to pass as arguments.  We would then have different subclasses for when new API versions have a different list of states.

JenySadadia:
Alternatively, to make this more generic we can return a dictionary with state names and values from API.
In the pipeline, we can construct dynamic enum with [function-call syntax](https://docs.python.org/3/library/enum.html#functional-api).
Fixed state names in the pipeline class can cause issues in case the existing state name changes in the API.

gctucker:
The states _have_ to be fixed in the pipeline, at least for each API version, as the implementation depends on it.  How do you think services can deal with state names that change?

JenySadadia:
Of course, services won't be able to deal with it but I suggested using state names and values from API considering that we have changed node states a few times and also added new states.
However, those changes were due to the API implementation evolving. So, yes we should have fixed state names per API version.

JenySadadia:
I was looking into this issue and there is no way to construct a dynamic `Enum` definition.
[...]

PR #195, opened 2022-12-19
T39494 src/timeout: use `db.get_node_states`

JenySadadia:
Depends on https://github.com/kernelci/kernelci-core/pull/1587
Fixes https://github.com/kernelci/kernelci-pipeline/issues/191

Instead of statically defining pending node states, use `db.node_states` enum to get the states from `kernelci-core`.
This will make it easier to use the node states across all the client services.

### DIFF
--- a/src/timeout.py
+++ b/src/timeout.py
@@ -24,10 +24,12 @@ from base import Service
 
 class TimeoutService(Service):
 
-    PENDING_STATES = ['running', 'available', 'closing']
-
     def __init__(self, configs, args, name):
         super().__init__(configs, args, name)
+        self._pending_states = [
+            state.value for state in self._db.node_states
+            if state != state.DONE
+        ]
 
     def _setup(self, args):
         return {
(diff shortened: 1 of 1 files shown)


JenySadadia:
As a follow-up, we can have `_setup()` method implemented for the `timeout` and add the polling period and pending states to context.

src/timeout.py
```diff
@@ -22,26 +22,27 @@
...
 
     def _get_pending_nodes(self, filters=None):
         nodes = {}
         node_filters = filters.copy() if filters else {}
-        for state in self.PENDING_STATES:
-            node_filters['state'] = state
+        for state in self._node_states:
+            if state.name == 'DONE':
+                continue
+            node_filters['state'] = state.value
```
gctucker, line 36: The state names are fixed for a given API version, so this should be initialised in the constructor rather than every time the method is called.  Something like this I guess:
```python
self._pending_states = [state for state in self._db.node_states if state.name != 'DONE']
```
Also I think there's an Enum now we can use instead of the plain `DONE` string.

gctucker: We could even have the node filter created in the constructor as that's also constant.

gctucker (commented):
Please rebase

src/timeout.py
JenySadadia: Rebased and updated the list to directly have `state.value`.

gctucker (approved):

JenySadadia:
The dependency https://github.com/kernelci/kernelci-core/pull/1587 is not merged yet.

gctucker:
Oops, sorry that's fixed now

Issue #183, opened 2022-11-25
Introduce Job class


mgalka:
At the moment the `Runner` service is responsible for defining, creating and running jobs. Job definitions should be decoupled from the running logic.

Job objects should be extracted from the current Runner service.  The base `Job` class should be introduced to unify structure and common functionality of the jobs. The Runner service should be responsible only for instantiating and running jobs which derive from the `Job` class. 

 - [ ]  Job class created
 - [ ]  Job related code refactored and moved into Job class derivatives
 - [ ]  Runner service logic adjusted to run Job class instances

gctucker:
I believe we also need to reshape the Lab class into something like Runtime and resolve the difference between the template arguments passed to the LAVA job generators and the new runtime templates using shell or k8s.  I think the Runtime class would have implementations that provide the template data to the job generation, basically we could have some test templates that only rely on common variables and the base runtime templates would be tied to the implementations to pass the relevant things as required.  The input data would still be the same, such as the incoming node which the job to be generated depended upon (sorry if that's a bit abstract...).  That probably also means we'll need to get the nodes dependency implementation done before sorting out the Runtime implementations.  In practice I guess we'll have to do things in incremental steps with some incomplete features and shortcuts or hard-coded things to get the whole thing bootstrapped.  Maybe some design / documentation would be nice to have to keep all the efforts well coordinated.  Let me know if you need some help with that.

PR #196, opened 2022-12-28
T39644 Implement job class

JenySadadia:
Fixes https://github.com/kernelci/kernelci-pipeline/issues/183

`Job` class has been implemented in order to separate logic for generating and scheduling various jobs (e.g. tests and builds) from `Runner`.
Instantiate it in the `Runner` and use it in `RunnerLoop` and `RunnerSingleJob` classes.

### DIFF
--- a/src/job.py
+++ b/src/job.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+"""KernelCI Job module"""
+
+import os
+import tempfile
+import json
+import requests
+
+import kernelci
+
+
+class Job():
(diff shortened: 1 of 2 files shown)


gctucker (commented):
Erm, I'm really not sure about that.  The kernelci-core tools are getting redesigned to cope with job generation for the new API so I don't see how this is going to work.

JenySadadia:
> 

This PR is about separating job related logic from `runner` service as described in the issue https://github.com/kernelci/kernelci-pipeline/issues/183. 
When everything will be in place for generating jobs in core tools, we can drop `Job` class and `runner` will have minimal changes as it will already be refactored.

gctucker:
Maybe, but I think this PR is a bit premature and a lot of the code might need to be dropped.  As long as these changes don't make things more difficult then I'm OK with having it merged.

gctucker (approved):
OK I guess this is likely to have to be reworked at some point but we need to merge it or drop it to resolve this.  So merging it now.

PR #198, opened 2023-01-13
config/pipeline.yaml: Append API version to config URL

JenySadadia:
Append API version `latest` to the URL for `docker-host` and `staging.kernelci.org` DB configurations.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -60,11 +60,11 @@ db_configs:
 
   docker-host:
     db_type: kernelci_api
-    url: http://172.17.0.1:8001
+    url: http://172.17.0.1:8001/latest/
 
   staging.kernelci.org:
     db_type: kernelci_api
-    url: https://staging.kernelci.org:9000
+    url: https://staging.kernelci.org:9000/latest/
 
 
 labs:


gctucker:
This doesn't actually work with `kernelci-core` as the `urllib` method to join URLs discards the path at the end.  So I've added a `version` field to the API config objects which should also help with supporting multiple versions, and that's now used in `kernelci-core` to insert the version in the path.

JenySadadia:
> 

This was not working at the beginning and I appended `/` at the end along with `latest` and then the parsing was fine and it worked locally.
Even the staging was working fine when I last checked with this PR merged on staging.

JenySadadia:
> OK I see where the problem is coming from:
> 
> ```
> >>> urllib.parse.urljoin('http://something.com:1234', 'bar')
> 'http://something.com:1234/bar'
> >>> urllib.parse.urljoin('http://something.com:1234/', 'bar')
> 'http://something.com:1234/bar'
> >>> urllib.parse.urljoin('http://something.com:1234/foo', 'bar')
> 'http://something.com:1234/bar'
> >>> urllib.parse.urljoin('http://something.com:1234/foo/', 'bar')
> 'http://something.com:1234/foo/bar'
> ```
> 
> So maybe you could review at least the last 2 commits in [kernelci/kernelci-core#1650](https://github.com/kernelci/kernelci-core/pull/1650)?
> 
> I think there are reasons for adding a `version` attribute in the YAML config:
> 
> * It's very easy to not put the trailing `/` as it's still a valid URL, so users are likely to get caught by that.  For example, if you visit https://staging.kernelci.org:9088/ the browser will actually drop the trailing `/` (with Chromium and Firefox at least...).  Who would think there's a problem here?
> 
> ```yaml
>   staging.kernelci.org:
>     url: https://staging.kernelci.org:9000/latest
> ```
> 
> * We'll most likely need to support multiple API versions in `kernelci-core`, so the `version` attribute can be used for that.
> 
> If you think these arguments don't justify the code change then please add some comments on the PR and I may just drop them and fix the URL strings there with a trailing `/`. Or an alternative would be to handle the missing trailing `/` in the code but not add a `version` attribute.

I agree that we should introduce `version` attribute instead. It is definitely a more efficient way than handling trailing `/`.

gctucker:
> I agree that we should introduce `version` attribute instead. It is definitely a more efficient way than handling trailing `/`.

OK, thanks for confirming.

PR #226, opened 2023-03-02
src/base.py: use kernelci.api.get_api()

gctucker:
Use the new kernelci.api.get_api() loader to get an API object rather than the kernelci.db one which is now deprecated.

### DIFF
--- a/src/base.py
+++ b/src/base.py
@@ -9,7 +9,6 @@ import logging
 import os
 
 import kernelci
-from kernelci.db import kernelci_api
 
 from logger import Logger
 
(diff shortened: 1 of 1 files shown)


gctucker:
Skipping for now while the new API class is being worked on.

gctucker:
The services have now all been updated with the new `API` and `APIHelper` classes.  Tested OK locally, now testing this on staging.

gctucker:
Tested OK on staging too.  However this should be merged at the same time as #228 and all the dependencies in `kernelci-core` that provide the new `API` and `APIHelper` classes.

PR #227, opened 2023-03-03
src/job.py: drop the templates_paths argument

gctucker:
The templates_paths argument has been dropped from Runtime.generate() as the templates are now all located in the config/runtime directory. Update the code in src/job.py accordingly.

Depends on https://github.com/kernelci/kernelci-core/pull/1777


gctucker:
Merging this now as well since https://github.com/kernelci/kernelci-core/pull/1777 has been merged.

PR #231, opened 2023-03-21
src/timeout: fix `base.Service` import

JenySadadia:
Fix the below import error for `from .base import Service` in timeout service:
`ImportError: attempted relative import with no known parent package`.


gctucker (approved):
Tested OK on staging.

PR #235, opened 2023-03-24
Run `kunit` and `kunit-x86_64`  in Kubernetes

gctucker:
As a temporary measure until the YAML configuration can provide a full description of all the jobs to run, rework the `runner.py loop --plan` argument as `--plans` to include a number of test plans and iterate over them when receiving a checkout node.  Update `docker-compose.yaml` accordingly with `kunit kunit-x86_64` for the `runner-k8s` service to run KUnit both on the default UML architecture and on `x86_64`.


gctucker:
Tested OK on staging

kci-show-002

kci-show-003

(small formatting issue with long test names...)

PR #241, opened 2023-03-28
src/base.py: use KCI_API_TOKEN environment variable

gctucker:
Look for the KCI_API_TOKEN environment variable rather than API_TOKEN to align with kernelci-core.

This is to follow a convention of having a KCI_ prefix for all the KernelCI environment variables to avoid name clashes.

### DIFF
--- a/src/base.py
+++ b/src/base.py
@@ -21,7 +21,7 @@ class Service:
         self._name = name
         self._logger = Logger("config/logger.conf", name)
         self._api_config = configs['api_configs'][args.api_config]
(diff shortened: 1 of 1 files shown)


JenySadadia (commented):
`https://kernelci.org/docs/api/getting-started/#configure-the-api-token` needs to be updated to reflect this change.

gctucker:
> https://kernelci.org/docs/api/getting-started/#configure-the-api-token needs to be updated to reflect this change.

https://github.com/kernelci/kernelci-api/pull/235

PR #243, opened 2023-04-03
src/runner.py: drop single run command

gctucker:
Now that we have "kci job" to generate a job and submit it, drop the "run" command used to run a single job in the runner service so it can be simplified as part of the core rework.

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -109,46 +109,6 @@ class RunnerLoop(Runner):
         return True
 
 
-class RunnerSingleJob(Runner):
-    """Runner subclass to execute a single job"""
-
-    def _get_node_from_commit(self, git_commit):
-        nodes = self._api.get_nodes({
-            "revision.commit": git_commit,
-        })
-        return nodes[0] if nodes else None
-
-    def _setup(self, args):
-        if args.node_id:
-            checkout_node = self._api.get_node(args.node_id)
-        elif args.git_commit:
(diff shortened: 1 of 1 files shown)


src/runner.py
```diff
@@ -109,46 +109,6 @@ def _run(self, sub_id):
...
 
 
-class RunnerSingleJob(Runner):
-    """Runner subclass to execute a single job"""
-
-    def _get_node_from_commit(self, git_commit):
-        nodes = self._api.get_nodes({
-            "revision.commit": git_commit,
-        })
-        return nodes[0] if nodes else None
```
JenySadadia, line 119: I see `kci job` command doesn't have provision to generate job from git commit.
Are you planning to add it?
If so, this should be dropped after adding the provision.

gctucker: Generally speaking, the pipeline scripts aren't command line tools so the trend should be to remove virtually all the code from them and have functions implemented in kernelci-core.  Then we'll have the `kci` command line interface on top of it for manual work, and the pipeline services for automation (and we might implement the service loops in kernelci-core directly eventually but that's a grey area still).

JenySadadia: Okay.
The only concern I had was to make sure the change is integrated to `kci job` before we drop it from here.
I think we are good to go.

JenySadadia (approved):

PR #246, opened 2023-04-04
docker-compose.yaml: disable kcidb bridge service

gctucker:
Comment out the section in docker-compose.yaml for the kcidb bridge service as it's currently broken and rebuilding the image takes a while on staging.  This should be addressed again once the core services have settled down.  It might be a good idea to move kcidb to a separate docker-compose file so users with a local deployment don't build the image etc. for nothing.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -92,21 +92,21 @@ services:
       - '--settings=${SETTINGS:-/home/kernelci/config/kernelci.conf}'
       - 'run'
 
-  kcidb:
-    container_name: 'kernelci-pipeline-kcidb'
-    build: {context: 'docker-kcidb'}
-    env_file: ['.env']
-    stop_signal: 'SIGINT'
-    command:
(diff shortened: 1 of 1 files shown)


JenySadadia:
It used to be working fine on staging for a while.
What errors are we receiving on staging atm?

gctucker:
It's kind of broken in several ways, and I don't want to spend time on this now as everything else is evolving around it.  So once things have settled down we'll definitely fix it properly.  If you want to investigate on your own you can already do that, but I haven't saved any logs.  Also it's taking a lot of time to build the Docker image just for this service which really gets in the way for working on the core pipeline services that still need to be developed.

JenySadadia:
> 

OK.
Could you please provide me the access to the staging instance so that I can take a look at it?

JenySadadia:
Adding `staging-skip` to this to test https://github.com/kernelci/kernelci-pipeline/pull/250

PR #249, opened 2023-04-07
Add KernelBuild data to the kbuild job

gctucker:
Populate the data associated with `kbuild` jobs in the template and with parameters in `pipeline.yaml`.

Depends on #248


config/runtime/kbuild.jinja2
```diff
@@ -55,17 +65,17 @@ echo '# {command}' | tee -a {job_log}
             os.remove(job_log_path)
 
         steps = {
-            'config': 'make ARCH=x86_64 x86_64_defconfig',
-            'kernel': 'make ARCH=x86_64 bzImage --jobs=$(nproc)',
-            'modules': 'make ARCH=x86_64 modules --jobs=$(nproc)',
+            'config': f"make ARCH=x86_64 {KBUILD_PARAMS['defconfig']}",
+            'kernel': "make ARCH=x86_64 bzImage --jobs=$(nproc)",
+            'modules': "make ARCH=x86_64 modules --jobs=$(nproc)",
```
JenySadadia, line 70: Shouldn't we use `KBUILD_PARAMS['arch]` for `ARCH` here?

gctucker: Yeah I guess. This template will only work with x86_64 anyway, but I'll update it.

JenySadadia: Can't we have a generic template?

gctucker: Actually it's going to make the code really messy to add formatting to each line and it'll just make it harder to rework later, so I don't think it's worth updating it here.

What we need instead is a series of templates for each build step, like sub-jobs.  That's also to be able to separate the actual `make` commands from the pipeline logic, with potentially a different image for each step in the case of Kubernetes jobs.

gctucker:
Alright, merging this now so we can keep making progress.  It'll all get refactored and properly support building for any architecture etc.

PR #250, opened 2023-04-10
src/send_kcidb: fix `_run` with `APIHelper`

JenySadadia:
Fix below error as the method renamed to `receive_event_node` as per the `APIHelper` implementation.
`AttributeError: 'APIHelper' object has no attribute 'receive_node_event'`

### DIFF
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -62,7 +62,7 @@ class KCIDBBridge(Service):
         self.log.info("Press Ctrl-C to stop.")
 
         while True:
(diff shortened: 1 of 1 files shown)


gctucker (approved):
OK, I thought there were other issues but if it's only that then fine.

However I still don't think we should enable it by default in `docker-compose.yaml` for local instances, and we might also disable it on staging if building the Docker image for KCIDB takes too much time in each iteration.  But that's separate from fixing the script.

PR #252, opened 2023-04-11
docker-compose-kcidb.yaml: add separate file with kcidb bridge

gctucker:
Move the kcidb bridge service to a separate docker-compose-kcidb.yaml file so it can be run only when required by the user.  It also avoid building the Docker image for it when just doing the default "docker-compose build".

### DIFF
--- a/docker-compose-kcidb.yaml
+++ b/docker-compose-kcidb.yaml
@@ -1,0 +1,29 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Collabora Limited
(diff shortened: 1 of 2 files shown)


gctucker:
Tested OK on staging:
```
04/17/2023 12:10:59 AM UTC [INFO] Listening for events... 
04/17/2023 12:10:59 AM UTC [INFO] Press Ctrl-C to stop.
04/17/2023 12:16:58 AM UTC [INFO] Submitting node to KCIDB: 643c8d44e9e758f21b3f600b
```

PR #253, opened 2023-04-19
config/runtime: kunit: parse JSON results if the file is present

gctucker:
Always parse the JSON results file if it's present, regardless of the status of the 'exec' step.  If any test case failed, the result will be "fail" but the data in the JSON file will still be present.

Fixes: 8015b47b18f0 ("config/runtime/kunit.jinja2: rework with separate steps")

### DIFF
--- a/config/runtime/kunit.jinja2
+++ b/config/runtime/kunit.jinja2
@@ -115,7 +115,7 @@ cd {src_path}
                 print(job_log_file.read())
                 print("--------------------------------------------------")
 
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging with mainline KUnit tests (exec step fails but overall job succeeds)
https://storage.staging.kernelci.org/api/kunit-6444ea8bbe20b960952aeb91/?C=M&O=D

PR #257, opened 2023-05-05
Update templates with new params for `kver` and `kunit`

gctucker:
Update how parameters are used in `kver.jinja2` and `kunit.jinja2` following the rework in https://github.com/kernelci/kernelci-core/pull/1919.


gctucker:
Tested OK on staging and locally.

PR #259, opened 2023-05-16
Add initial LAVA callback service

gctucker:
Add an initial LAVA callback service.  This is not yet suitable for a production-like environment as the docker-compose setup uses the development Werkzeug built-in server that comes with Flask rather than a proper WSGI application server.  Also it doesn't handle authentication tokens from the callback but relies on an environment variable.  These things will be addressed in a follow-up PR.

This current PR contains the logic to parse a LAVA callback and send a hierarchy of test results extracted from it to the API.  The `login` test case is created like with the legacy `kernelci-backend` implementation based on the job status.  This does not parse the logs yet, which is also something to address in a follow-up.

It can be safely tested by capturing a LAVA callback using plain HTTP and no token into a file, then sending that HTTP request manually to the service (e.g. using `nc`).  I've tested this with a real LAVA job running a `baseline` test on QEMU; will be adding more details in comments.


gctucker:
This worked on staging:
```
05/26/2023 01:34:35 PM UTC [INFO] 2023-05-26 13:34:35.084375  f915a3305353  6470b3e1d3444e718e39c1b8  Done       Pass      baseline-x86
```
```
[pid: 1|app: 0|req: 2/3] 164.92.146.209 () {44 vars in 776 bytes} [Fri May 26 13:34:34 2023] POST /node/6470b3e1d3444e718e39c1b8 => generated 4576 bytes in 166 msecs (HTTP/1.1 200) 2 headers in 73 bytes (1 switches on core 0)
```
LAVA job: https://lava.collabora.dev/scheduler/job/10465956/definition

gctucker:
Here's a node hierarchy with a kernel build and related baseline test:
kci-show-005

PR #260, opened 2023-05-19
Get API config and token from callback data

gctucker:
Rather than using a hard-coded API config and loading the API token from the environment, get the API config name from the job's metadata and the token from the HTTP Authorization header sent by lava.


gctucker:
Removing skip label here and on #259 so that @VinceHillier can work on adding the Nginx configuration and verify it's working with the service.

PR #265, opened 2023-05-23
config/runtime: use NODE in kbuild and kunit

gctucker:
The NODE_ID global variable has been removed so adjust the kbuild and kunit templates accordingly to use the NODE object instead.


JenySadadia (approved):
LGTM

PR #267, opened 2023-05-29
T42060 src/regression_tracker: check for regression in all child nodes

### LINKED ISSUE #36: Regression tracking
In a similar way that test regressions are being tracked in `kernelci-backend`, a service needs to be added to the new pipeline to listen to events with new results, determine whether they are regressions and if so submit it to the API.

The main difference with the `kernelci-backend` implementation is that the schema is different, now we need to check for regressions between Node objects which could be any kind of result (build, test, anything in between).

> Note: A model for regressions needs to be added to the API with endpoints to be able to submit and retrieve them.

JenySadadia:
When runner submits a hierarchy of nodes at a single shot, the API will publish event just for parent node.
In that case, the parent node can have "pass" result, even if one of its child has failed.
Tweak regression tracker to listen to events when "checkout" node reaches to "done" state instead of getting nodes with just "fail" result.
That way we can traverse through all the child nodes using the checkout node and check for regressions.

Fixes #36

### DIFF
--- a/src/regression_tracker.py
+++ b/src/regression_tracker.py
@@ -27,7 +27,6 @@ class RegressionTracker(Service):
     def _setup(self, args):
         return self._api_helper.subscribe_filters({
             'state': 'done',
-            'result': 'fail',
         })
 
     def _stop(self, sub_id):
@@ -43,29 +42,19 @@ class RegressionTracker(Service):
         regression['regression_data'] = [last_successful_node, failed_node]
         self._api_helper.submit_regression(regression)
 
-    def _run(self, sub_id):
-        """Method to run regression tracking"""
-        self.log.info("Tracking regressions... ")
-        self.log.info("Press Ctrl-C to stop.")
-        sys.stdout.flush()
+    def _detect_regression(self, node):
+        """Method to check and detect regression"""
+        previous_nodes = self._api.get_nodes({
+            'name': node['name'],
+            'group': node['group'],
+            'path': node['path'],
+            'revision.tree': node['revision']['tree'],
+            'revision.branch': node['revision']['branch'],
+            'revision.url': node['revision']['url'],
+            'created__lt': node['created'],
+        })
 
-        while True:
-            node = self._api_helper.receive_event_node(sub_id)
-            if not node['group']:
-                continue
-
-            previous_nodes = self._api.get_nodes({
-                'name': node['name'],
-                'group': node['group'],
-                'revision.tree': node['revision']['tree'],
-                'revision.branch': node['revision']['branch'],
-                'revision.url': node['revision']['url'],
-                'created__lt': node['created'],
-            })
-
-            if not previous_nodes:
-                continue
-
+        if previous_nodes:
             previous_nodes = sorted(
                 previous_nodes,
                 key=lambda node: node['created'],
@@ -76,6 +65,32 @@ class RegressionTracker(Service):
                 self.log.info(f"Detected regression for node id: \
 {node['id']}")
                 self._create_regression(node, previous_nodes[0])
+
+    def _get_all_failed_child_nodes(self, failures, root_node):
+        """Method to get all failed nodes recursively from top-level node"""
+        child_nodes = self._api.get_nodes({'parent': root_node['id']})
+        for node in child_nodes:
+            if node['result'] == 'fail':
+                failures.append(node)
+            self._get_all_failed_child_nodes(failures, node)
+
+    def _run(self, sub_id):
+        """Method to run regression tracking"""
+        self.log.info("Tracking regressions... ")
+        self.log.info("Press Ctrl-C to stop.")
+        sys.stdout.flush()
+
+        while True:
+            node = self._api_helper.receive_event_node(sub_id)
+
+            if node['name'] == 'checkout':
+                continue
+
+            failures = []
+            self._get_all_failed_child_nodes(failures, node)
+
+            for node in failures:
+                self._detect_regression(node)
 
             sys.stdout.flush()
         return True


JenySadadia:
It seems like some node has an invalid value for `timeout` on staging. It is breaking the regression tracker:
```
today at 17:10:0105/29/2023 11:40:01 AM UTC [INFO] Tracking regressions... 
today at 17:10:0105/29/2023 11:40:01 AM UTC [INFO] Press Ctrl-C to stop.
today at 17:22:0905/29/2023 11:52:09 AM UTC [ERROR] Traceback (most recent call last):
today at 17:22:09  File "/home/kernelci/pipeline/base.py", line 69, in run
today at 17:22:09    status = self._run(context)
today at 17:22:09  File "/home/kernelci/pipeline/regression_tracker.py", line 83, in _run
today at 17:22:09    failures.extend(self._api.get_nodes({
today at 17:22:09  File "/usr/local/lib/python3.10/site-packages/kernelci-1.1-py3.10.egg/kernelci/api/latest.py", line 115, in get_nodes
today at 17:22:09    resp = self._get('nodes', params=params)
today at 17:22:09  File "/usr/local/lib/python3.10/site-packages/kernelci-1.1-py3.10.egg/kernelci/api/__init__.py", line 139, in _get
today at 17:22:09    resp.raise_for_status()
today at 17:22:09  File "/usr/local/lib/python3.10/site-packages/requests/models.py", line 1021, in raise_for_status
today at 17:22:09    raise HTTPError(http_error_msg, response=self)
today at 17:22:09requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://staging.kernelci.org:9000/latest/nodes?group=kunit&result=fail&limit=100&offset=100
today at 17:22:09
today at 17:28:21Container stopped
```

src/regression_tracker.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, configs, args):
     def _setup(self, args):
         return self._api_helper.subscribe_filters({
             'state': 'done',
-            'result': 'fail',
+            'name': 'checkout',
```
gctucker, line 30: I think it will still be useful to catch regressions earlier than when the top-level checkout node is done.

src/regression_tracker.py
```diff
@@ -77,6 +66,28 @@ def _run(self, sub_id):
...
+            node = self._api_helper.receive_event_node(sub_id)
+            group_nodes = self._api.get_nodes({'parent': node['id']})
+
+            failures = []
+            for node in group_nodes:
+                if node['result'] == 'fail':
+                    failures.append(node)
+                failures.extend(self._api.get_nodes({
+                    'group': node['group'],
+                    'result': 'fail',
```
gctucker, line 85: I think this will also add the node in `group_nodes`, and the query should be more restrictive to pick nodes with the same parent path and same kernel revision to avoid other test results that have the same group name.  Or we could add a convention to have unique group names, like a group ID but I'm not sure if that's a good idea.

JenySadadia: This method is to get failed child nodes of the parent `checkout`.
I think you are referring to the logic for checking regression nodes. That is in  `_detect_regression` method.
So, first I am fetching all the child nodes with `fail` result recursively in `_run` and then the tracker will look for regression by iterating over each failed node in `_detect_regression`.

JenySadadia: Maybe I should add `path` filter while getting `previous_nodes`.

src/regression_tracker.py
JenySadadia: Yes, that can also be done. Either we check one by one for every `done` node or look for all the nodes at once from top-level node, it will be the same at the end.

JenySadadia: In both cases, we will need a different set of preconditions e.g. in case we look for each node one by one, we need to have some checks to see if the node has failed and if the node is not checkout.
We should go with the approach that is better in terms of performance. Maybe we can compare the number of API calls.
Also, in the approach with `checkout` node, the script will track regression after the timeout has been reached and not as soon as it occurred. That is another downside.

JenySadadia:
Fixed an issue with the logic to stop getting all the failed nodes matching `node.group` from the past rather than just looking for nodes related to the current `checkout`.

JenySadadia:
> Fixed an issue with the logic to stop getting all the failed nodes matching `node.group` from the past rather than just looking for nodes related to the current `checkout`.

However, this can also get nodes from another `checkout`. Looking into it right now.

JenySadadia:
> > Fixed an issue with the logic to stop getting all the failed nodes matching `node.group` from the past rather than just looking for nodes related to the current `checkout`.
> 
> However, this can also get nodes from another `checkout`. Looking into it right now.

Updated logic to fix this.
Tested OK on staging by failing immediate child nodes of `kbuild` and `kunit`.
```
today at 12:19:5806/05/2023 06:49:58 AM UTC [INFO] Detected regression for node id: 647d85955b7a776509a9fa1e
today at 12:19:5806/05/2023 06:49:58 AM UTC [INFO] Detected regression for node id: 647d85955b7a776509a9fa1f
today at 12:19:5906/05/2023 06:49:59 AM UTC [INFO] Detected regression for node id: 647d85955b7a776509a9fa20
today at 12:19:5906/05/2023 06:49:59 AM UTC [INFO] Detected regression for node id: 647d85955b7a776509a9fa21
today at 12:19:5906/05/2023 06:49:59 AM UTC [INFO] Detected regression for node id: 647d85955b7a776509a9fa22
```

gctucker:
See also `kci node`:
```
$ ./kci node count --api-config=staging.kernelci.org kind=regression created__gt=2023-06-05T00:08:52.174000
11
```
```
$ ./kci node find --api-config=staging.kernelci.org kind=regression created__gt=2023-06-05T00:08:52.174000 > regressions.json.txt
```
[regressions.json.txt](https://github.com/kernelci/kernelci-pipeline/files/11653998/regressions.json.txt)

gctucker:
Also I think this is more accurate, using the kernel revision rather than the timestamps:
```
$ ./kci node find --api-config=staging.kernelci.org kind=regression revision.describe=staging-next-20230605.0
```
Several revisions might be tested at the same time and regression nodes could be created in parallel for them.  In this particular case, the output is identical.

JenySadadia:
> See also `kci node`:
> 
> ```
> $ ./kci node count --api-config=staging.kernelci.org kind=regression created__gt=2023-06-05T00:08:52.174000
> 11
> ```
> 
> ```
> $ ./kci node find --api-config=staging.kernelci.org kind=regression created__gt=2023-06-05T00:08:52.174000 > regressions.json.txt
> ```
> 
> [regressions.json.txt](https://github.com/kernelci/kernelci-pipeline/files/11653998/regressions.json.txt)

The `kci` tools seem pretty useful. Thanks.

PR #269, opened 2023-05-29
src/notifier: fix program name in `parse_opts`

JenySadadia:
Correct command line program name argument from `runner` to `notifier` for `kernelci.cli.parse_opts` method. The method is used to get `Options` object with command line arguments and user settings.


gctucker (approved):
Sorry I noticed that back in Feb but never got round to sending the fix.  So yes that's already tested :+1:

PR #271, opened 2023-06-07
config/pipeline.yaml: use new base_url for staging storage

gctucker:
Use the new base URL for the staging storage which allows plaintext HTTP as well as HTTPS (part of the regular staging storage Nginx setup).


gctucker:
Tested OK on staging.

Issue #221, opened 2023-02-09
Job configuration implementation


gctucker:
Once we've established the principles for reworking the pipeline job configuration, this is about doing the initial implementation.  It should be possible to use it locally with the `kci` command line tool as well as with the automated pipeline.

For example, it could be running KUnit in k8s and also in parallel run 3 kernel builds (e.g. x86_64, arm64, arm defconfigs), each of them would then be running baseline and baseline-nfs on one target platform (e.g. x86 QEMU, some arm64 Chromebook and common arm boards) for a total of 15 LAVA tests.

- [x] minimum features implemented to have trigger / checkout / kunit / kernel-build / baseline jobs running
- [ ] ability to spawn multiple jobs in parallel with a series of parameters
- [ ] schema definitions for the new YAML job configuration created to help with validation
- [ ] tested locally with `kci` tool
- [ ] tested on staging

nuclearcat:
Closing as completed and superseded. The current pipeline supports configured parallel jobs, validation, local tooling, and production deployment. Any remaining configuration improvements should be tracked separately.

PR #284, opened 2023-06-23
config/pipeline.yaml: update to use scheduler events

gctucker:
Update the YAML pipeline config to use event attributes in the scheduler configuration rather than the job run_on attributes.

Fixes: https://github.com/kernelci/kernelci-pipeline/issues/221


gctucker:
Tested on staging in combination with https://github.com/kernelci/kernelci-core/pull/1974

PR #288, opened 2023-07-27
config/kernelci.toml: use dotted keys

JenySadadia:
As the syntax for TOML settings file has been switched from using `":"` to dotted keys for child section of components like database and storage, use the new syntax in the settings file.


gctucker (approved):

gctucker:
Tested OK manually (staging is temporarily using Azure Files so it's not testing this config right now).

PR #289, opened 2023-07-27
config/pipeline.yaml: rename `staging.kernelci.org` to `staging`

JenySadadia:
Update YAML configuration for staging instance. Rename `staging.kernelci.org` to `staging` in order to simplify TOML section names related to staging. We need to surround `staging.kernelci.org` with double quotes to use dotted key  syntax for the section as the name includes a dot. e.g. `db."staging.kernelci.org"`. After renaming the config name, it will become `db.staging`. It is good to use config options names with a dot for the sake of simplicity.


gctucker (approved):
Tested OK on staging.

Note: We might eventually have a prefix for standard kernelci.org configs e.g. `kci-staging` in case some users have their own kind of "staging" etc.  But that's maybe something to refine later.

PR #292, opened 2023-07-27
lava: look up yaml_config in settings

gctucker:
Replace the hard-coded 'config/pipeline.yaml' path to the YAML config file with the value found in the TOML settings file.  Keep using 'config/pipeline.yaml' by default if it's not defined in the settings file.


gctucker (commented):
Tested OK on staging.

PR #293, opened 2023-07-27
lava: use new dotted-key syntax for storage settings section

gctucker:
Update how the name of the storage settings section is defined by using the new dotted-key syntax and wrapping the config name in double quotes.


src/lava_callback.py
```diff
@@ -29,8 +29,7 @@ def _get_api_helper(api_config_name, api_token):
 
 def _get_storage(storage_config_name):
     storage_config = CONFIGS['storage_configs'][storage_config_name]
-    section = SETTINGS[(':'.join(('storage', storage_config_name)))]
-    storage_cred = section['storage_cred']
+    storage_cred = SETTINGS['storage'][storage_config_name]['storage_cred']
```
JenySadadia, line 32: Can we use `parse_opts` instead to read the settings from TOML like other services?

gctucker: No because this is not a service like the other ones, it's a Flask application.  There might be things we can do to try to make it look like a service but I made the design decision to rely directly on the YAML config and TOML settings as it seemed like the best choice in this case.

JenySadadia: Yes, I am aware it's a Flask app but I was wondering if we could use `kernelci` package somehow to read settings. But it seems not. Looks fine otherwise.

JenySadadia (approved):

PR #294, opened 2023-07-28
Convert email configs .env variables to TOML settings

JenySadadia:
To keep the settings in standard way, remove environment variables for email sender and recipient and convert them to command line arguments.

### DIFF
--- a/config/kernelci.toml
+++ b/config/kernelci.toml
@@ -23,6 +23,8 @@ kcidb_project_id = "kernelci-production"
 origin = "kernelci"
 
 [test_report]
+email_sender = "bot@kernelci.org"
+email_recipient = "kernelci-results-staging@groups.io"
 
 [timeout]
 
--- a/src/kernelci_pipeline/email_sender.py
+++ b/src/kernelci_pipeline/email_sender.py
@@ -16,11 +16,11 @@ import smtplib
 
 class EmailSender:
     """Class to send email report using SMTP"""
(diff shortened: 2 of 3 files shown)


JenySadadia:
`EMAIL_USER` can be removed as it should be the same as `email_sender`.

gctucker (approved):
LGTM - the whole thing for email reports will need to be rewritten anyway but yes the sending account settings will need to be stored in TOML so I think that's here to stay, more or less.

JenySadadia:
Added `email_sender` and `email_recipient` to `staging.kernelci.org.secrets.toml`.

gctucker:
The `email_recipient` recipient setting is typically the first thing that will be removed as it should end up in YAML instead for the system-wide settings.  We'll probably also have user-defined email preferences at some point in the future.

gctucker:
Somehow the emails are still not getting through on https://groups.io/g/kernelci-results-staging/topics

There are no errors in the logs, the report content is being generated as expected.  So I'm not sure what's going on there.

JenySadadia:
> Somehow the emails are still not getting through on https://groups.io/g/kernelci-results-staging/topics
> 
> There are no errors in the logs, the report content is being generated as expected. So I'm not sure what's going on there.

The staging config file `staging.kernelci.org.secrets.toml` is already been updated with the email sender and recipient settings.
That should have worked.

gctucker:
> That should have worked.

Except it doesn't :sweat_smile:

PR #297, opened 2023-08-01
[WIP] src/test_report.py: Include patchwork patch metadata into test report

yurinnick:
- Include patch `title`, `hash` and `web_url` in the report, when available
- Pass `args`  from `__init__` into `_setup()` in `Service` class


gctucker:
As with https://github.com/kernelci/kernelci-pipeline/pull/299, I think this will depend on the rest of the code to have settled into a functional implementation first.  Adding `staging-skip` label for now as it's likely to break every time the other PRs get updated.

PR #300, opened 2023-08-30
Improve events log format

gctucker:
Print the events log directly on stdout and add titles for the columns to improve readability.

PR #301, opened 2023-08-30
src/monitor.py: rename from notifier

gctucker:
This service is to monitor the API Pub/Sub events rather than sending
notifications which is more what the email report service does.  As
such, rename it from notifier to monitor.

Depends on #300


gctucker:
Merging now as that will need to be aligned with the documentation.

PR #302, opened 2023-08-30
config/pipeline.yaml: add early access API entry

gctucker:
Add an API config entry in pipeline.yaml for the Early Access instance deployed in AKS.


gctucker:
Tested OK with the early access deployment (monitor and trigger).

PR #303, opened 2023-08-30
kube/aks: add Dockerfile.production

gctucker:
Add Dockerfile.production to create a kernelci/pipeline Docker image suitable for production deployments in AKS.


gctucker:
Tested OK with early access deployment.

PR #304, opened 2023-08-30
kube/aks: add monitor.yaml to create the monitor pod

gctucker:
Add monitor.yaml to deploy the monitor in AKS.
    
    This is not strictly speaking a service in Kubernetes terminology as
    it doesn't expose any ports and it's only a Pod.  It is kind of a
    service as far as the pipeline is concerned.

### DIFF
--- a/kube/aks/monitor.yaml
+++ b/kube/aks/monitor.yaml
@@ -1,0 +1,25 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Collabora Limited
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK with early access deployment in AKS:
```
$ kubectl --namespace=kernelci-pipeline logs -f monitor-599dc75c6-gnxkb
08/30/2023 10:01:13 AM UTC [INFO] Listening for events... 
08/30/2023 10:01:13 AM UTC [INFO] Press Ctrl-C to stop.
Time                        Commit        Node Id                  State      Result    Name
```

PR #307, opened 2023-08-31
config/pipeline.yaml: add early-access-azure storage

gctucker:
Add a early-access-azure storage entry to be able to use the AzureFile storage under a different share name for early-access and keep it separate from staging or individual users.


gctucker (commented):
Tested OK with AKS deployment.

PR #316, opened 2023-09-01
src/timeout: only update nodes owned by the current user

gctucker:
In order to avoid hitting issues with API permissions when updating nodes that aren't owned by the current user, even with an admin API token, ignore nodes that have a different owner to the current user running the timeout service.

This is most likely a temporary measure until API permissions have been fixed as the timeout service should in principle be able to transition all nodes from all users.

### DIFF
--- a/src/timeout.py
+++ b/src/timeout.py
@@ -27,6 +27,8 @@ class TimeoutService(Service):
             state.value for state in self._api.node_states
             if state != state.DONE
         ]
+        self._user = self._api.whoami()
+        self._username = self._user['profile']['username']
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on AKS deployment.  This is probably more of a workaround which will need to be revisited.  We might accept the fact that only nodes that belong to a particular user or group get updated with the timeout services, so that the "main" pipeline doesn't timeout nodes from other users.  It would probably seem like an expected thing to happen anyway as users decide the timeout value in nodes they create.

PR #317, opened 2023-09-01
kube/aks: add timeout.yaml to create the timeout pods

gctucker:
Add timeout.yaml to create the timeout, holdoff and closing pods.


gctucker:
Tested OK with AKS deployment.

PR #285, opened 2023-07-05
adding kubernetes manifests to deploy and test kernelci-pipeline on minikube

sbdtu5498:
Fixes https://github.com/kernelci/kernelci-api/issues/273
@gctucker @JenySadadia 
 Will be adding README.md soon.

PR #322, opened 2023-09-04
kube/minikube: Add k8s manifest files for pipeline

JenySadadia:
Replacing https://github.com/kernelci/kernelci-pipeline/pull/285 with this PR.

The former pull request is on top of @sbdtu5498's pipeline fork i.e. https://github.com/sbdtu5498/kernelci-pipeline.
The fork is not up-to-date and hence it is not able to rebase the #285 for the latest pipeline changes i.e. configurations in `pipeline.yaml` and renaming pipeline services.
Because of the time constraints creating this new PR and cherry-picked all the commits from the older one.


gctucker:
Please also add your `Signed-off-by` trailer to each commit you're picking up from Sanskar.

JenySadadia:
> Please also add your `Signed-off-by` trailer to each commit you're picking up from Sanskar.

Done.

gctucker:
Ah I see, there was only one.  Maybe it's also worth fixing the first commit message format while keeping the same meaning.  Actually it doesn't have a description but just a very long subject...

config/pipeline.yaml
```diff
@@ -23,6 +23,9 @@ api_configs:
   early-access:
     url: https://kernelci-api.eastus.cloudapp.azure.com
 
+  k8s-host:
+    url: http://kernelci-api:8001
```
gctucker, line 27: How is this going to resolve when using the command locally on the actual host?  Is this meant to only be accessible from within the cluster?

kube/minikube/deployments/regression-tracker-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+            - "/home/kernelci/pipeline/regression_tracker.py"
+            - "--settings=$(SETTINGS)"
+            - "run"
+            - "--api-config=k8s-host"
+          volumeMounts:
+            - name: pipeline-volume
+              mountPath: /home/kernelci/pipeline
+            - name: config-volume
+              mountPath: /home/kernelci/config
+          # Resource limits needs to be discussed.
```
gctucker, line 46: I guess it's up to the user to decide, maybe some minimal values can be provided here as an example?

JenySadadia:
> Ah I see, there was only one. Maybe it's also worth fixing the first commit message format while keeping the same meaning. Actually it doesn't have a description but just a very long subject...

Oh yes, I missed it.
I'll fix.

kube/minikube/deployments/scheduler-docker-deployment.yaml
```diff
@@ -0,0 +1,77 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
```
gctucker, line 1: I thought we said we shouldn't be running this one in Kubernetes?

kube/minikube/deployments/tarball-deployment.yaml
```diff
@@ -0,0 +1,68 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
```
gctucker, line 1: How are the credentials for the storage being passed to the container?

kube/minikube/deployments/timeout-closing-deployment.yaml
```diff
@@ -0,0 +1,55 @@
...
+                  name: kernelci-pipeline-config
+                  key: settings
+          command:
+            - "/usr/bin/env"
+            - "python3"
+            - "/home/kernelci/pipeline/timeout.py"
+            - "--settings=$(SETTINGS)"
+            - "run"
+            - "--mode=closing"
+            - "--api-config=k8s-host"
```
gctucker, line 41: Is this not already in the TOML settings file?

JenySadadia:
> Ah I see, there was only one. Maybe it's also worth fixing the first commit message format while keeping the same meaning. Actually it doesn't have a description but just a very long subject...

Fixed the commit.

kube/minikube/scripts/deploy.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
+
+# Function to check if a resource exists
+check_resource_exist() {
+  local kind=$1
+  local name=$2
+
+  # Wait for the resource to be available
+  while [[ -z "$(kubectl get $kind $name -o json)" ]]; do
```
gctucker, line 9: How about `kubectl wait`?

kube/minikube/deployments/scheduler-docker-deployment.yaml
JenySadadia: Yes, I'll drop this.

kube/minikube/deployments/timeout-closing-deployment.yaml
JenySadadia: In the case of Minikube, TOML file is being used from `/home/kernelci/config/kernelci.toml` location.
That's why `api-config` needs to be passed here.
I think I need to add a different TOML file for Minikube as well like we have one for AKS.

kube/minikube/deployments/tarball-deployment.yaml
JenySadadia: I uploaded SSH key to the Minikube node.
Please see comments https://github.com/kernelci/kernelci-pipeline/pull/285#issuecomment-1670817784 and https://github.com/kernelci/kernelci-pipeline/pull/285#issuecomment-1671101131.

config/pipeline.yaml
JenySadadia: Yes, `k8s-host` is for accessing API within the cluster.
To use it locally, we need to `post-forword` API service and then `http://localhost:8001` would work fine.
The command for port forward has been mentioned [here](https://github.com/kernelci/kernelci-api/blob/main/kube/minikube/README.md).

gctucker: Right, and then that would require a different URL in the YAML config.

Another way to do it I guess would be to define a service, for example:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web
spec:
  replicas: 1
  selector:
    matchLabels:
      app: web
  template:
    metadata:
      labels:
        app: web
    spec:
      containers:
      - name: web
        image: nginx:1.21.3
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: web
spec:
  type: LoadBalancer
  ports:
  - port: 80
  selector:
    app: web
```

kube/minikube/deployments/tarball-deployment.yaml
gctucker: OK, and then not defining any credentials in the TOML settings?  In principle it should have the SSH key name (when using SSH storage).

kube/minikube/deployments/timeout-closing-deployment.yaml
gctucker: Yes, I think it's expected to have a different settings file for every use-case.  We may also need to use YAML config overlays, but that's a bit different.

kube/minikube/deployments/tarball-deployment.yaml
gctucker: Ah yes, sorry I missed that.  I was still thinking of the AKS deployment which has a secret token in `storage_cred` for Azure Files.

kube/minikube/deployments/regression-tracker-deployment.yaml
JenySadadia: Not sure if it has some default values. I'll check the documentation.

kube/minikube/scripts/deploy.sh
JenySadadia: Yes, it looks like a better option. I'll update the script.
Thanks for pointing it out.

JenySadadia: It turns out that `kubectl wait` is not useful for checking resource existence.
Please see https://github.com/kubernetes/kubernetes/issues/83242
Also, the `wait` tool is experimental. Hence we should keep `kubectl get` instead.

kube/minikube/deployments/regression-tracker-deployment.yaml
JenySadadia: There are no default values for resource limits if not provided.
We need to analyze the resource usage of different containers before adding limits here.

gctucker: Well I don't really think we can do that, it just depends on the scale of the deployment.  Having something very simple like 0.5 CPU and 256 MB of memory would probably be fine, then users can adjust it as needed?

gctucker: ...as limits, and maybe half of that as requirements?

gctucker:
OK I don't have MiniKube set up locally so it would be better if someone else could test and review this.

@sbdtu5498 @nuclearcat Is one of you able to do this?

nuclearcat:
I dont have minikube, but can setup k3s again, also i have small patch that makes this deployment scenario vendor-neutral. I will try to re-test it today with this patch

gctucker:
Is there anything in this minikube deployment that isn't vendor-neutral?

gctucker:
Also the plan is basically to have this tested and merged, then we'll be looking at how to consolidate the MiniKube and AKS deployments to have the common parts pulled together into a single set of YAML files and differences taken care of on top of that.

kube/minikube/deployments/regression-tracker-deployment.yaml
JenySadadia: Yes, we can do that.

kube/minikube/scripts/deploy.sh
JenySadadia: Thanks for sharing the link.
I have added `kubectl wait` for deployments.

gctucker:
OK thanks for the update.  As this isn't directly used in any deployment at the moment, if we don't hear anything today I think we can merge it tomorrow.  Then we can start consolidating the AKS deployment with it to have some common Kubernetes files and the common parts can be tested in AKS with the staging cluster.

kube/minikube/scripts/deploy.sh
```diff
@@ -0,0 +1,96 @@
...
+  echo "$kind $name is available!"
+}
+
+# Fork GitHub repository in Minikube node
+kubectl apply -f ../init/init-pod.yaml
+check_job_completed "github-cloning-pod"
+
+# Create Kubernetes secret
+kubectl create secret generic kernelci-pipeline-secrets --from-literal=kci-api-token=$KCI_API_TOKEN
+check_secret_key_exist secret kernelci-pipeline-secrets -o jsonpath='{.data.*}'
```
nuclearcat, line 45: Just got k3s testing now and got following while running script:
```
# ./deploy.sh
job.batch/pipeline-github-cloning-job created
job.batch/pipeline-github-cloning-job condition met
Init job is completed successfully
secret/kernelci-pipeline-secrets created
./deploy.sh: line 45: check_secret_key_exist: command not found
configmap/kernelci-pipeline-config created
configmaps kernelci-pipeline-config is available!
deployment.apps/kernelci-pipeline-monitor created
deployment.apps/kernelci-pipeline-monitor condition met
deployment.apps/kernelci-pipeline-regression-tracker created
deployment.apps/kernelci-pipeline-regression-tracker condition met
deployment.apps/kernelci-pipeline-scheduler created
deployment.apps/kernelci-pipeline-scheduler condition met
deployment.apps/kernelci-pipeline-scheduler configured
deployment.apps/kernelci-pipeline-scheduler condition met
deployment.apps/kernelci-pipeline-scheduler configured
deployment.apps/kernelci-pipeline-scheduler condition met
deployment.apps/kernelci-pipeline-tarball created
deployment.apps/kernelci-pipeline-tarball condition met
deployment.apps/kernelci-pipeline-test-report created
deployment.apps/kernelci-pipeline-test-report condition met
deployment.apps/kernelci-pipeline-timeout-closing created
deployment.apps/kernelci-pipeline-timeout-closing condition met
deployment.apps/kernelci-pipeline-timeout created
deployment.apps/kernelci-pipeline-timeout condition met
deployment.apps/kernelci-pipeline-timeout-holdoff created
deployment.apps/kernelci-pipeline-timeout-holdoff condition met
deployment.apps/kernelci-pipeline-trigger created
deployment.apps/kernelci-pipeline-trigger condition met
All components deployed successfully!
```
I think there is small typo in function name, s is missing

gctucker: So if this deployment works as-is both in MiniKube and k3s then it's already vendor-agnostic.  Is that right?

gctucker: (Well pending it actually works as there's a problem in the script apparently)

JenySadadia: Thanks for the review @nuclearcat 
Fixed it.

nuclearcat: Yes, correct @gctucker , it works with k3s as well.

gctucker (approved):
OK I guess there aren't any known issues with this any more so it should be good to go.  On the other hand, it's not really used by KernelCI or tested automatically so there might be hidden issues.

The next step is to consolidate this and the API counterpart with the AKS deployments, and the AKS deployments are used so we can verify they can get redeployed and work in the KernelCI staging cluster.

Issue #319, opened 2023-09-03
Add checkout node timeout argument to `trigger.py`


gctucker:
The timeout duration for checkout nodes is currently hard-coded in `trigger.py`.  In fact there are a couple of patches to adjust it in the code, which currently leads to unexpected behaviour with very short timeout on the AKS deployment.

To solve this issue, adding something like `--checkout-timeout` to the trigger service would make it possible to define the value in the user settings or command line.

PR #323, opened 2023-09-08
src/trigger: make checkout node timeout configurable

JenySadadia:
Fixes https://github.com/kernelci/kernelci-pipeline/issues/319

At the moment, timeout for `checkout` node has been calculated statically in the `trigger` service.
Make it configurable by adding optional command line arguments. Add default timeout configuration values in settings file.

### DIFF
--- a/config/kernelci.toml
+++ b/config/kernelci.toml
@@ -6,6 +6,7 @@ verbose = true
 [trigger]
 poll_period = 0
 startup_delay = 3
+timeout = 60
 
 [tarball]
 kdir = "/home/kernelci/data/src/linux"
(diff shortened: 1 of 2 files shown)


JenySadadia:
> That's very verbose and difficult to use I think. How about just setting a `--timeout` value in minutes?

Yes, that would work too. However, I wanted to make it more configurable by providing all the options we have for the timeout.

JenySadadia:
Updated to have single command line argument `--timeout`.

gctucker (approved):

gctucker (approved):
FYI here's a little snippet to double-check that `timedelta` can cope with this use-case:
```
>>> timedelta(minutes=90.5)
datetime.timedelta(seconds=5430)
```
So the timeout can be more than 60 minutes and it can be a decimal (float) number.

JenySadadia:
> FYI here's a little snippet to double-check that `timedelta` can cope with this use-case:
> 
> ```
> >>> timedelta(minutes=90.5)
> datetime.timedelta(seconds=5430)
> ```
> 
> So the timeout can be more than 60 minutes and it can be a decimal (float) number.

Yes, it does work with float values.

PR #335, opened 2023-10-13
src/timeout: update logic to get username of current user

JenySadadia:
Since the user schema has been changed in the API, update the logic to get `username` in `timeout` service.


JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/377

PR #338, opened 2023-10-17
src: use kernelci.legacy.cli

gctucker:
The kernelci.cli module is now being moved to kernelci.legacy.cli so we can have the new implementation in kernelci.cli instead.  Update all the pipeline services accordingly.


nuclearcat:
Works fine, observed how it fixes issue in staging

PR #339, opened 2023-10-17
pipeline.yaml: Update SAS token for reading files

nuclearcat:
Update SAS token to read Azure files after ZDI-CAN-22317.


nuclearcat:
Verified in Dozzle

gctucker (approved):

PR #342, opened 2023-10-21
src/patchset.py: Implement Patchset service

yurinnick:
Patchset service process patchset nodes:
- Wait for parent checkout node to be available
- Download checkout node tarball
- Apply patches and calculate patchset hash
- Upload new tarball

### DIFF
--- a/src/patchset.py
+++ b/src/patchset.py
@@ -1,0 +1,329 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# Author: Nikolay Yurin <yurinnick@meta.com>
+
+import os
+import sys
+import json
+import requests
+import time
+import tempfile
+import hashlib
+from datetime import datetime, timedelta
+from urllib.parse import urlparse
+from urllib.request import urlopen
+
+import kernelci
+import kernelci.build
+import kernelci.config
+from kernelci.legacy.cli import Args, Command, parse_opts
+import kernelci.storage
+
+from base import Service
+from tarball import Tarball
+
+
+class Patchset(Tarball):
+    TAR_CREATE_CMD = """\
+set -e
+cd {target_dir}
+tar --create --transform "s/^/{prefix}\\//" * | gzip > {tarball_path}
+"""
+
+    APPLY_PATCH_SHELL_CMD = """\
+set -e
+cd {checkout_path}
+patch -p1 < {patch_file}
+"""
+
+    # FIXME: I really don"t have a good idea what I"m doing here
+    # This code probably needs rework and put into kernelci.patch
+    def _hash_patch(self, patch_name, patch_file):
+        allowed_prefixes = {
+            b"old mode",  # Old file permissions
+            b"new mode",  # New file permissions
+            b"-",  # This convers both removed lines and source file
+            b"+",  # This convers both added lines and target file
+            # "@" I don"t know how we should handle hunks yet
+        }
+        hashable_patch_lines = []
+        for line in patch_file.readlines():
+            if not line:
+                continue
+
+            for prefix in allowed_prefixes:
+                if line.startswith(prefix):
+                    hashable_patch_lines.append(line)
+                    break
+
+        hashable_content = b"/n".join(hashable_patch_lines)
+        self.log.debug(
+            "Hashable content:\n" +
+            hashable_content.decode("utf-8")
+        )
+        patch_hash_digest = hashlib.sha256(hashable_content).hexdigest()
+        self.log.debug(f"Patch {patch_name} hash: {patch_hash_digest}")
+        return patch_hash_digest
+
+    # FIXME: move into kernelci.patch
+    def _apply_patch(self, checkout_path, patch_name, patch_url):
+        self.log.info(
+            f"Applying patch {patch_name}, url: {patch_url}",
+        )
+        try:
+            encoding = urlopen(patch_url).headers.get_charsets()[0]
+        except Exception as e:
+            self.log.warn(
+                "Failed to fetch encoding from patch "
+                f"{patch_name} headers: {e}"
+            )
+            self.log.warn("Falling back to utf-8 encoding")
+            encoding = "utf-8"
+
+        with tempfile.NamedTemporaryFile(
+            prefix="{}-{}-".format(
+                self._service_config.patchset_tmp_file_prefix,
+                patch_name
+            ),
+            encoding=encoding
+        ) as tmp_f:
+            if not kernelci.build._download_file(patch_url, tmp_f.name):
+                raise FileNotFoundError(
+                    f"Error downloading patch from {patch_url}"
+                )
+
+            kernelci.shell_cmd(self.APPLY_PATCH_SHELL_CMD.format(
+                checkout_path=checkout_path,
+                patch_file=tmp_f.name,
+            ))
+
+            return self._hash_patch(patch_name, tmp_f)
+
+    # FIXME: move into kernelci.patch
+    def _apply_patches(self, checkout_path, patch_artifacts):
+        patchset_hash = hashlib.sha256()
+        for patch_name, patch_url in patch_artifacts.items():
+            patch_hash = self._apply_patch(checkout_path, patch_name, patch_url)
+            patchset_hash.update(patch_hash.encode("utf-8"))
+
+        patchset_hash_digest = patchset_hash.hexdigest()
+        self.log.debug(f"Patchset hash: {patchset_hash_digest}")
+        return patchset_hash_digest
+
+    def _download_checkout_archive(self, download_path, tarball_url, retries=3):
+        self.log.info(f"Downloading checkout tarball, url: {tarball_url}")
+        tar_filename = os.path.basename(urlparse(tarball_url).path)
+        kernelci.build.pull_tarball(
+            kdir=download_path,
+            url=tarball_url,
+            dest_filename=tar_filename,
+            retries=retries,
+            delete=True
+        )
+
+    def _update_node(
+        self,
+        patchset_node,
+        checkout_node,
+        tarball_url,
+        patchset_hash
+    ):
+        patchset_data = checkout_node.get("data", {}).copy()
+        patchset_data["kernel_revision"]["patchset"] = patchset_hash
+
+        updated_node = patchset_node.copy()
+        updated_node["artifacts"]["tarball"] = tarball_url
+        updated_node["state"] = "available"
+        updated_node["data"] = patchset_data
+        updated_node["holdoff"] = str(
+            datetime.utcnow() + timedelta(minutes=10)
+        )
+
+        try:
+            self._api.node.update(updated_node)
+        except requests.exceptions.HTTPError as err:
+            err_msg = json.loads(err.response.content).get("detail", [])
+            self.log.error(err_msg)
+
+    def _setup(self, *args):
+        return self._api_helper.subscribe_filters({
+            "op": "created",
+            "name": "patchset",
+            "state": "running",
+        })
+
+    def _has_allowed_domain(self, url):
+        domain = urlparse(url).hostname
+        if domain not in self._service_config.allowed_domains:
+            raise RuntimeError(
+                "Forbidden mbox domain %s, allowed domains: %s",
+                domain,
+                self._service_config.allowed_domains,
+            )
+
+    def _get_patch_artifacts(self, patchset_node):
+        node_artifacts = patchset_node.get("artifacts")
+        if not node_artifacts:
+            raise ValueError(
+                "Patchset node %s has no artifacts",
+                patchset_node["id"],
+            )
+
+        for patch_mbox_url in node_artifacts.values():
+            self._has_allowed_domain(patch_mbox_url)
+
+        return node_artifacts
+
+    def _gen_checkout_name(self, checkout_node):
+        revision = checkout_node["data"]["kernel_revision"]
+        return "-".join([
+            "linux",
+            revision["tree"],
+            revision["branch"],
+            revision["describe"],
+        ])
+
+    def _process_patchset(self, checkout_node, patchset_node):
+        patch_artifacts = self._get_patch_artifacts(patchset_node)
+
+        # Tarball download implicitely removes destination dir
+        # there's no need to cleanup this directory
+        self._download_checkout_archive(
+            download_path=self._service_config.kdir,
+            tarball_url=checkout_node["artifacts"]["tarball"]
+        )
+
+        checkout_name = self._gen_checkout_name(checkout_node)
+        checkout_path = os.path.join(self._service_config.kdir, checkout_name)
+
+        patchset_hash = self._apply_patches(checkout_path, patch_artifacts)
+        patchset_hash_short = patchset_hash[
+            :self._service_config.patchset_short_hash_len
+        ]
+
+        tarball_path = self._make_tarball(
+            target_dir=checkout_path,
+            tarball_name=f"{checkout_name}-{patchset_hash_short}"
+        )
+        tarball_url = self._push_tarball(tarball_path)
+
+        self._update_node(
+            patchset_node=patchset_node,
+            checkout_node=checkout_node,
+            tarball_url=tarball_url,
+            patchset_hash=patchset_hash
+        )
+
+    def _mark_failed(self, patchset_node):
+        node = patchset_node.copy()
+        node.update({
+            "state": "done",
+            "result": "fail",
+        })
+        try:
+            self._api.node.update(node)
+        except requests.exceptions.HTTPError as err:
+            err_msg = json.loads(err.response.content).get("detail", [])
+            self.log.error(err_msg)
+
+    def _mark_failed_if_no_parent(self, patchset_node):
+        if not patchset_node["parent"]:
+            self.log.error(
+                f"Patchset node {patchset_node['id']} as has no parent"
+                "checkout node , marking node as failed",
+            )
+            self._mark_failed(patchset_node)
+            return True
+
+        return False
+
+    def _mark_failed_if_parent_failed(self, patchset_node, checkout_node):
+        if (
+            checkout_node["state"] == "done" and
+            checkout_node["result"] == "fail"
+        ):
+            self.log.error(
+                f"Parent checkout node {checkout_node['id']} failed, "
+                f"marking patchset node {patchset_node['id']} as failed",
+            )
+            self._mark_failed(patchset_node)
+            return True
+
+        return False
+
+    def _run(self, _sub_id):
+        self.log.info("Listening for new trigger events")
+        self.log.info("Press Ctrl-C to stop.")
+
+        while True:
+            patchset_nodes = self._api.node.find({
+                "name": "patchset",
+                "state": "running",
+            })
+
+            if patchset_nodes:
+                self.log.debug(f"Found patchset nodes: {patchset_nodes}")
+
+            for patchset_node in patchset_nodes:
+                if self._mark_failed_if_no_parent(patchset_node):
+                    continue
+
+                checkout_node = self._api.node.get(patchset_node["parent"])
+
+                if self._mark_failed_if_parent_failed(
+                    patchset_node,
+                    checkout_node
+                ):
+                    continue
+
+                if checkout_node["state"] == "running":
+                    self.log.info(
+                        f"Patchset node {patchset_node['id']} is waiting "
+                        f"for checkout node {checkout_node['id']} to complete",
+                    )
+                    continue
+
+                try:
+                    self.log.info(
+                        f"Processing patchset node: {patchset_node['id']}",
+                    )
+                    self._process_patchset(checkout_node, patchset_node)
+                except Exception as e:
+                    self.log.error(
+                        f"Patchset node {patchset_node['id']} "
+                        f"processing failed: {e}",
+                    )
+                    self.log.traceback()
+                    self._mark_failed(patchset_node)
+
+            self.log.info(
+                "Waiting %d seconds for a new nodes..." %
+                self._service_config.polling_delay_secs,
+            )
+            time.sleep(self._service_config.polling_delay_secs)
+
+
+class cmd_run(Command):
+    help = (
+        "Wait for a checkout node to be available "
+        "and push a source+patchset tarball"
+    )
+    args = [
+        Args.kdir, Args.output, Args.api_config, Args.storage_config,
+    ]
+    opt_args = [
+        Args.verbose, Args.storage_cred,
+    ]
+
+    def __call__(self, configs, args):
+        return Patchset(configs, args).run(args)
+
+
+if __name__ == "__main__":
+    opts = parse_opts("patchset", globals())
+    configs = kernelci.config.load("config")
+    status = opts.command(configs, opts)
+    sys.exit(0 if status is True else 1)
--- a/src/tarball.py
+++ b/src/tarball.py
@@ -5,13 +5,12 @@
 # Copyright (C) 2022 Collabora Limited
 # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
 # Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+# Author: Nikolay Yurin <yurinnick@meta.com>
 
 from datetime import datetime, timedelta
-import logging
 import os
 import re
 import sys
-import urllib.parse
 import json
 import requests
 
@@ -32,25 +31,30 @@ KVER_RE = re.compile(
 
 
 class Tarball(Service):
+    TAR_CREATE_CMD = """\
+set -e
+cd {target_dir}
+git archive --format=tar --prefix={prefix}/ HEAD | gzip > {tarball_path}
+"""
 
-    def __init__(self, configs, args):
-        super().__init__(configs, args, 'tarball')
-        self._build_configs = configs['build_configs']
-        self._kdir = args.kdir
-        self._output = args.output
-        if not os.path.exists(self._output):
-            os.makedirs(self._output)
-        self._verbose = args.verbose
-        self._storage_config = configs['storage_configs'][args.storage_config]
+    def __init__(self, global_configs, service_config):
+        super().__init__(global_configs, service_config, 'tarball')
+        self._service_config = service_config
+        self._build_configs = global_configs['build_configs']
+        if not os.path.exists(self._service_config.output):
+            os.makedirs(self._service_config.output)
+        storage_config = global_configs['storage_configs'][
+            service_config.storage_config
+        ]
         self._storage = kernelci.storage.get_storage(
-            self._storage_config, args.storage_cred
+            storage_config, service_config.storage_cred
         )
 
     def _find_build_config(self, node):
         revision = node['data']['kernel_revision']
         tree = revision['tree']
         branch = revision['branch']
-        for name, config in self._build_configs.items():
+        for config in self._build_configs.values():
             if config.tree.name == tree and config.branch == branch:
                 return config
 
@@ -61,13 +65,16 @@ class Tarball(Service):
         '''
         self.log.info(f"Updating repo for {config.name}")
         try:
-            kernelci.build.update_repo(config, self._kdir)
+            kernelci.build.update_repo(config, self._service_config.kdir)
         except Exception as err:
             self.log.error(f"Failed to update: {err}, cleaning stale repo")
             # safeguard, make sure it is git repo
-            if not os.path.exists(os.path.join(self._kdir, '.git')):
-                self.log.error(f"{self._kdir} is not a git repo")
-                raise Exception(f"{self._kdir} is not a git repo")
+            if not os.path.exists(
+                os.path.join(self._service_config.kdir, '.git')
+            ):
+                err_msg = f"{self._service_config.kdir} is not a git repo"
+                self.log.error(err_msg)
+                raise Exception(err_msg)
             # cleanup the repo and return True, so we try again
             kernelci.shell_cmd(f"rm -rf {self._kdir}")
             return True
@@ -75,24 +82,24 @@ class Tarball(Service):
         self.log.info("Repo updated")
         return False
 
-    def _make_tarball(self, config, describe):
-        name = '-'.join(['linux', config.tree.name, config.branch, describe])
-        tarball = f"{name}.tar.gz"
-        self.log.info(f"Making tarball {tarball}")
-        output_path = os.path.relpath(self._output, self._kdir)
-        cmd = """\
-set -e
-cd {kdir}
-git archive --format=tar --prefix={name}/ HEAD | gzip > {output}/{tarball}
-""".format(kdir=self._kdir, name=name, output=output_path, tarball=tarball)
+    def _make_tarball(self, target_dir, tarball_name):
+        self.log.info(f"Making tarball {tarball_name}")
+        tarball_path = os.path.join(
+            self._service_config.output,
+            f"{tarball_name}.tar.gz"
+        )
+        cmd = self.TAR_CREATE_CMD.format(
+            target_dir=target_dir,
+            prefix=tarball_name,
+            tarball_path=tarball_path
+        )
         self.log.info(cmd)
         kernelci.shell_cmd(cmd)
         self.log.info("Tarball created")
-        return tarball
+        return tarball_path
 
-    def _push_tarball(self, config, describe):
-        tarball_name = self._make_tarball(config, describe)
-        tarball_path = os.path.join(self._output, tarball_name)
+    def _push_tarball(self, tarball_path):
+        tarball_name = os.path.basename(tarball_path)
         self.log.info(f"Uploading {tarball_path}")
         tarball_url = self._storage.upload_single((tarball_path, tarball_name))
         self.log.info(f"Upload complete: {tarball_url}")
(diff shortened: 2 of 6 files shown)


yurinnick:
## Setup node
```
$ echo '{
  "name":"checkout",
  "path": ["checkout"],
  "revision":{
    "tree":"mainline",
    "url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
    "branch":"master",
    "commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26"
  }
}' | ./kci node submit
...
$ echo '{
  "name":"patchset",
  "path": [
    "checkout",
    "patchset"
  ],
  "parent": "6533024fc2a20de21962f5c6",
  "revision":{
    "tree":"mainline",
    "url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
    "branch":"master",
    "commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26"
  },
  "artifacts": {
    "0001-patch.patch": "https://patchwork.kernel.org/project/linux-mm/patch/2ea7cbce80ac7c62e90cbfb9653a7972f902439f.1697816692.git.raghavendra.kt@amd.com/raw"
  }
}' | ./kci node submit
...
```

## Patchset logs
```
kernelci-pipeline-patchset | 10/21/2023 08:51:00 AM UTC [INFO] Listening for new trigger events
kernelci-pipeline-patchset | 10/21/2023 08:51:00 AM UTC [INFO] Press Ctrl-C to stop.
kernelci-pipeline-patchset | 10/21/2023 08:51:41 AM UTC [INFO] Applying patch 0001-patch.patch to the repo, mbox url: https://patchwork.kernel.org/project/linux-mm/patch/2ea7cbce80ac7c62e90cbfb9653a7972f902439f.16
97816692.git.raghavendra.kt@amd.com/raw
kernelci-pipeline-patchset | 10/21/2023 08:51:43 AM UTC [INFO] Making tarball linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6
kernelci-pipeline-patchset | 10/21/2023 08:51:43 AM UTC [INFO] set -e
kernelci-pipeline-patchset | cd /home/kernelci/data/src/linux-patchset/linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b/
kernelci-pipeline-patchset | tar --create --transform "s/^/linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6/" * | gzip > /home/kernelci/data/output/
linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6.tar.gz
kernelci-pipeline-patchset |
kernelci-pipeline-patchset | 10/21/2023 08:52:14 AM UTC [INFO] Tarball created
kernelci-pipeline-patchset | 10/21/2023 08:52:14 AM UTC [INFO] Uploading linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6.tar.gz
kernelci-pipeline-patchset | 10/21/2023 08:52:14 AM UTC [INFO] Connected (version 2.0, client OpenSSH_8.4p1)
kernelci-pipeline-patchset | 10/21/2023 08:52:15 AM UTC [INFO] Authentication (publickey) successful!
kernelci-pipeline-patchset | 10/21/2023 08:52:16 AM UTC [INFO] Upload complete: http://172.17.0.1:8002/linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f
8e6.tar.gz
```

## Tarball logs
```
kernelci-pipeline-tarball | 10/21/2023 08:53:25 AM UTC [INFO] Updating repo for mainline
kernelci-pipeline-tarball | From https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux
kernelci-pipeline-tarball |  * branch                      HEAD       -> FETCH_HEAD
kernelci-pipeline-tarball | From git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux
kernelci-pipeline-tarball |  * branch                      HEAD       -> FETCH_HEAD
kernelci-pipeline-tarball | HEAD is now at 9c5d00cb7b6b Merge tag 'perf-tools-fixes-for-v6.6-2-2023-10-20' of git://git.kernel.org/pub/scm/linux/kernel/git/perf/perf-tools
kernelci-pipeline-tarball | 10/21/2023 08:53:31 AM UTC [INFO] Repo updated
kernelci-pipeline-tarball | 10/21/2023 08:53:31 AM UTC [INFO] Making tarball linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b.tar.gz
kernelci-pipeline-tarball | 10/21/2023 08:53:31 AM UTC [INFO] set -e
kernelci-pipeline-tarball | cd /home/kernelci/data/src/linux
kernelci-pipeline-tarball | git archive --format=tar --prefix=linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b/ HEAD | gzip > ../../output/linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b.tar.gz
kernelci-pipeline-tarball |
kernelci-pipeline-tarball | 10/21/2023 08:54:04 AM UTC [INFO] Tarball created
kernelci-pipeline-tarball | 10/21/2023 08:54:04 AM UTC [INFO] Uploading /home/kernelci/data/output/linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b.tar.gz
kernelci-pipeline-tarball | 10/21/2023 08:54:05 AM UTC [INFO] Upload complete: http://172.17.0.1:8002/linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b.tar.gz
```

## Check tarball
```
$ wget http://localhost:8002/linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6.tar.gz
$ tar -tf ~/Downloads/linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6.tar.gz | head
linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6/COPYING
linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6/CREDITS
linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6/Documentation/
linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6/Documentation/.gitignore
linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6/Documentation/ABI/
linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6/Documentation/ABI/README
linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6/Documentation/ABI/obsolete/
linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6/Documentation/ABI/obsolete/o2cb
linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6/Documentation/ABI/obsolete/procfs-i8k
linux-mainline-master-v6.6-rc6-285-g9c5d00cb7b6b-cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6/Documentation/ABI/obsolete/sysfs-bus-iio
```

src/patchset.py
```diff
@@ -0,0 +1,250 @@
...
+tar --create --transform "s/^/{prefix}\\//" * | gzip > {tarball_path}
+"""
+
+    APPLY_PATCH_SHELL_CMD = """\
+set -e
+cd {kdir}
+patch -p1 < {patch_file}
+"""
+
+    # FIXME: I really don't have a good idea what I'm doing here
```
yurinnick, line 45: I definitely need some help here, I am not confident enough to say which part we should leave and which we should redact before hashing.

src/tarball.py
```diff
@@ -59,25 +63,22 @@ def _update_repo(self, config):
...
+        return tarball_path
 
-    def _push_tarball(self, config, describe):
-        tarball_name = self._make_tarball(config, describe)
-        tarball_path = os.path.join(self._output, tarball_name)
-        self.log.info(f"Uploading {tarball_path}")
+    def _push_tarball(self, tarball_path):
+        tarball_name = os.path.basename(tarball_path)
+        self.log.info(f"Uploading {tarball_name}")
         tarball_url = self._storage.upload_single((tarball_path, tarball_name))
```
pawiecz, line 82: Please correct me if I'm wrong: `tarball_path` now also contains `tarball_name` (see L68). Should the `upload_single()` call be adjusted?

TODO(pawiecz): double check if upload was unaffected/needs adjusting

src/patchset.py
```diff
@@ -0,0 +1,250 @@
...
+                if checkout_node["state"] == 'running':
+                    continue
+
+                try:
+                    self._process_patchset(checkout_node, patchset_node)
+                except Exception as e:
+                    self._mark_failed(patchset_node)
+                    self.log.error(e)
+
+            time.sleep(10)
```
pawiecz, line 227: This service currently implements equivalent of `trigger` and extension to `tarball` - is it intended to be divided further with a separate trigger service? I don't suggest doing it right now, I'm rather curious if [trigger-like options](https://github.com/kernelci/kernelci-pipeline/blob/9e68e16e11c683af463aeb1d18f0ea46ff6b1b5a/src/trigger.py#L84) would also apply here

pawiecz (commented):
I really appreciate posting such a detailed verification steps and all the log information. These will greatly help verifying this patch and also maybe set up an E2E suite for Patchwork-related pipeline changes. Let me deploy these changes and get back to you on Monday if some other questions come up

src/patchset.py
yurinnick: The biggest difficulty I faced with this logic was the condition that we need to start `patchset` node processing after it's parent `checkout` node becomes `available` (otherwise no source tarball available for download).

If there's a better way to do this delayed processing I am all for it.

src/tarball.py
gctucker: FYI The Python abstract class for Storage needs to be reviewed anyway as the `upload()` arguments are a bit hard to grasp with 2-tuples etc.  It was just a way to get something to work and tick a box.

src/patchset.py
gctucker: But, why `sleep(10)` anyway?

src/patchset.py
gctucker: I think we've discussed this before.  If I remember correctly, and it's probably in the open hours notes, the consensus was to take all the contents of what `git format-patch` would produce including the commit details and author etc. as well as the base Git SHA1.  This basically would be the same as doing actual Git commits, as the SHA1 revisions include all the meta-data.  That way, you would only get the same patchset hash if and only if the exact same patch files were used with the same base Git revision.  So you know for sure that results would only be collated with other matching revisions, if another user submitted the same patchset with the same base commit.

gctucker: And we also said that an alternative which could be done as a follow-up but would be a bit more intricate, would be to only compute the hash with the diff and the commit message but leave out the rest of the meta-data and base Git commit.  Then clients would need to rely on both the base Git commit in `revision.commit` and `revision.patchset` as a pair of values to identify patchsets.  A same patchset applied on a different base commit would have the same `revision.patchset` hash but obviously a different `revision.commit` value.  The advantage here is to be able to compare same identical patchsets but applied on different base Git revisions, say once on a subsystem maintainer's branch and once on linux-next, as they will have the same `revision.patchset` regardless of how the patch files were created (different commit time etc.) as long as the diff and commit messages are the same.  So we could collate results with a bit more flexibility, at the expense of always checking the two fields.  Well we probably also need to deal with the two fields anyway when using the approach in my previous comment so maybe this is actually the way to go :p

I might be available on Wednesday for the Open Hours to discuss this one more time, maybe one last time though.

src/tarball.py
pawiecz: @gctucker right, thanks for clarifying. Taking point to look at it in the future, resolving thread for now.

Also, my initial comment made little sense: `tarball_path` *always* contained `tarball_name`, it was just processed differently.

src/patchset.py
gctucker: Ah I see, to periodically check if a new patchset is available.  I thought there would be a bridge service or some kind of callback to get a Pub/Sub event generated from Patchwork when that happened.

src/patchset.py
```diff
@@ -0,0 +1,250 @@
...
+        patchset_build_config = self._find_build_config(patchset_node)
+        if build_config != patchset_build_config:
+            raise ValueError(
+                f"Patchsets node {patchset_node['id']} build config "
+                f"doesn't match to the parent node {checkout_node['id']}"
+            )
+
+        self._download_checkout_archive(checkout_node["artifacts"]["tarball"])
+
+        chekout_name = '-'.join([
```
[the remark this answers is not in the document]
gctucker: About the patch format, this would ideally need to be clearly defined in a written form as far as what is expected from the API point of view when submitting a patchset.  It matters because it means the checksum will be different depending on what is being included.  I think the ideal approach is to require having just the commit subject, message and diff of course.  Additional meta-data e.g. commit author, date/time, Git notes etc. shouldn't be required.  This is to enable collating results together from a same patchset applied in different conditions.

Using `patch` to apply the diff is fine as that's the least common denominator, I don't think it's needed to recreate a local git history.  So regardless of what's in the patch files it should work and produce the same build.  Also, the initial kernel source tree comes from a tarball with no git history anyway.

yurinnick:
# Submit patchset node
```
$ cat patchset_node.json
{
  "name":"patchset",
  "path": [
    "checkout",
    "patchset"
  ],
  "parent": "65bd7cfdb656782e895218c3",
  "revision":{
    "tree":"mainline",
    "url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
    "branch":"master",
    "commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26"
  },
  "artifacts": {
    "0001-patch.patch": "https://patchwork.kernel.org/project/kvm/patch/20240203003518.387220-1-seanjc@google.com/raw"
  }
}
$ kci node submit ./patchset_node.json
```
## Patchset logs
```
02/08/2024 09:29:23 PM UTC [INFO] Listening for new trigger events
02/08/2024 09:29:23 PM UTC [INFO] Press Ctrl-C to stop.
02/08/2024 09:29:24 PM UTC [INFO] Waiting 30 seconds for a new nodes...
02/08/2024 09:29:54 PM UTC [DEBUG] Found patchset nodes: [{'id': '65c547ca316ff4bd1ddf9395', 'kind': 'node', 'name': 'patchset', 'path': ['checkout', 'patchset'], 'group': None, 'parent': '65bd7cfdb656782e895218c3', 'state': 'running', 'result': None, 'artifacts': {'0001-patch.patch': 'https://patchwork.kernel.org/project/kvm/patch/20240203003518.387220-1-seanjc@google.com/raw'}, 'data': None, 'debug': None, 'created': '2024-02-08T21:29:46.243000', 'updated': '2024-02-08T21:29:46.243000', 'timeout': '2024-02-09T03:29:46.243000', 'holdoff': None, 'owner': 'admin', 'user_groups': []}]
02/08/2024 09:29:54 PM UTC [INFO] Processing patchset node: 65c547ca316ff4bd1ddf9395
02/08/2024 09:29:54 PM UTC [INFO] Downloading checkout tarball, url: http://172.17.0.1:8002/linux-mainline-master-v6.8-rc2-397-g56897d51886f.tar.gz
02/08/2024 09:33:37 PM UTC [INFO] Applying patch 0001-patch.patch, url: https://patchwork.kernel.org/project/kvm/patch/20240203003518.387220-1-seanjc@google.com/raw
02/08/2024 09:33:42 PM UTC [DEBUG] Hashable content:
--- a/arch/x86/kvm/mmu/mmu.c
/n+++ b/arch/x86/kvm/mmu/mmu.c
/n+	/*
/n+	 * Pre-check for a relevant mmu_notifier invalidation event prior to
/n+	 * acquiring mmu_lock.  If there is an in-progress invalidation and the
/n+	 * kernel allows preemption, the invalidation task may drop mmu_lock
/n+	 * and yield in response to mmu_lock being contended, which is *very*
/n+	 * counter-productive as this vCPU can't actually make forward progress
/n+	 * until the invalidation completes.  This "unsafe" check can get false
/n+	 * negatives, i.e. KVM needs to re-check after acquiring mmu_lock.
/n+	 *
/n+	 * Do the pre-check even for non-preemtible kernels, i.e. even if KVM
/n+	 * will never yield mmu_lock in response to contention, as this vCPU is
/n+	 * *guaranteed* to need to retry, i.e. waiting until mmu_lock is held
/n+	 * to detect retry guarantees the worst case latency for the vCPU.
/n+	 */
/n+	if (mmu_invalidate_retry_gfn_unsafe(vcpu->kvm, fault->mmu_seq, fault->gfn)) {
/n+		kvm_release_pfn_clean(fault->pfn);
/n+		return RET_PF_RETRY;
/n+	}
/n+
/n--- a/include/linux/kvm_host.h
/n+++ b/include/linux/kvm_host.h
/n+
/n+/*
/n+ * This lockless version of the range-based retry check *must* be paired with a
/n+ * call to the locked version after acquiring mmu_lock, i.e. this is safe to
/n+ * use only as a pre-check to avoid contending mmu_lock.  This version *will*
/n+ * get false negatives and false positives.
/n+ */
/n+static inline bool mmu_invalidate_retry_gfn_unsafe(struct kvm *kvm,
/n+						   unsigned long mmu_seq,
/n+						   gfn_t gfn)
/n+{
/n+	/*
/n+	 * Use READ_ONCE() to ensure the in-progress flag and sequence counter
/n+	 * are always read from memory, e.g. so that checking for retry in a
/n+	 * loop won't result in an infinite retry loop.  Don't force loads for
/n+	 * start+end, as the key to avoiding infinite retry loops is observing
/n+	 * the 1=>0 transition of in-progress, i.e. getting false negatives
/n+	 * due to stale start+end values is acceptable.
/n+	 */
/n+	if (unlikely(READ_ONCE(kvm->mmu_invalidate_in_progress)) &&
/n+	    gfn >= kvm->mmu_invalidate_range_start &&
/n+	    gfn < kvm->mmu_invalidate_range_end)
/n+		return true;
/n+
/n+	return READ_ONCE(kvm->mmu_invalidate_seq) != mmu_seq;
/n+}

02/08/2024 09:33:42 PM UTC [DEBUG] Patch 0001-patch.patch hash: 96560b86f7238970362dc0f4220cddfc0c437e1f605f61eef8e074c9b37ef4c2
02/08/2024 09:33:42 PM UTC [DEBUG] Patchset hash 4f3428d5f61f53d0eabb4c30fe02ef47b9a7d7f4305017318d1c2116da6b8b96
02/08/2024 09:33:42 PM UTC [INFO] Making tarball linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5
02/08/2024 09:33:42 PM UTC [INFO] set -e
cd /home/kernelci/data/src/linux-patchset/linux-mainline-master-v6.8-rc2-397-g56897d51886f/
tar --create --transform "s/^/linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5\//" * | gzip > /home/kernelci/data/output/linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5.tar.gz

02/08/2024 09:38:53 PM UTC [INFO] Tarball created
02/08/2024 09:38:53 PM UTC [INFO] Uploading linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5.tar.gz
02/08/2024 09:38:53 PM UTC [INFO] Connected (version 2.0, client OpenSSH_8.4p1)
02/08/2024 09:38:54 PM UTC [INFO] Authentication (publickey) successful!
02/08/2024 09:39:04 PM UTC [INFO] Upload complete: http://172.17.0.1:8002/linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5.tar.gz
02/08/2024 09:39:05 PM UTC [INFO] Waiting 30 seconds for a new nodes...
```

## Check tarball
```
$ tar -tf ~/Downloads/linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5.tar.gz | head
linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5/COPYING
linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5/CREDITS
linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5/Documentation/
linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5/Documentation/.gitignore
linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5/Documentation/ABI/
linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5/Documentation/ABI/README
linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5/Documentation/ABI/obsolete/
linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5/Documentation/ABI/obsolete/o2cb
linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5/Documentation/ABI/obsolete/procfs-i8k
linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5/Documentation/ABI/obsolete/sysfs-bus-iio
```

## Check node
```
{
  "id": "65c547ca316ff4bd1ddf9395",
  "kind": "node",
  "name": "patchset",
  "path": [
    "checkout",
    "patchset"
  ],
  "group": null,
  "parent": "65bd7cfdb656782e895218c3",
  "state": "done",
  "result": null,
  "artifacts": {
    "tarball": "http://172.17.0.1:8002/linux-mainline-master-v6.8-rc2-397-g56897d51886f-4f3428d5f61f5.tar.gz"
  },
  "data": {
    "kernel_revision": {
      "tree": "mainline",
      "url": "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
      "branch": "master",
      "commit": "56897d51886fa7e9f034ff26128eb09f1b811594",
      "describe": "v6.8-rc2-397-g56897d51886f",
      "version": {
        "version": "6",
        "patchlevel": "8",
        "extra": "-rc2-397-g56897d51886f"
      },
      "patchset": "4f3428d5f61f53d0eabb4c30fe02ef47b9a7d7f4305017318d1c2116da6b8b96"
    }
  },
  "debug": null,
  "created": "2024-02-08T21:29:46.243000",
  "updated": "2024-02-08T21:49:39.798000",
  "timeout": "2024-02-09T03:29:46.243000",
  "holdoff": "2024-02-08T21:49:05.153000",
  "owner": "admin",
  "user_groups": []
}
```

src/patchset.py
yurinnick: I wish there would be a way to get event, and then reschedule it if `checkout_node` isn't available yet, but it doesn't seem to exist now, so I have to poll nodes from API.

nuclearcat: we are in development phase, so maybe we can introduce something, to make it event driven. For example we can introduce pipeline service that will trigger external callback and provide to external URL checkout_node details.

pawiecz: (will proceed with current implementation, adding a TODO to allow event-driven triggering for patchsets)

src/patchset.py
```diff
@@ -0,0 +1,322 @@
...
+        Args.verbose, Args.storage_cred,
+    ]
+
+    def __call__(self, configs, args):
+        return Patchset(configs, args).run(args)
+
+
+if __name__ == "__main__":
+    opts = parse_opts("patchset", globals())
+    configs = kernelci.config.load("config/pipeline.yaml")
```
pawiecz, line 320: File `pipeline.yaml` has grown too big and had to be split into separate smaller ones. Services are now being adjusted to load multiple config files from a directory. See [#435](https://github.com/kernelci/kernelci-pipeline/pull/435/commits/8b9e60895d20a6db02c502311766415274262c79)

```suggestion
    configs = kernelci.config.load("config")
```

pawiecz (approved):
LGTM :+1: 

I suggested a minor change to align with current PR related to configuration loading

src/patchset.py
```diff
@@ -0,0 +1,322 @@
...
+from kernelci.legacy.cli import Args, Command, parse_opts
+import kernelci.storage
+
+from tarball import Tarball
+
+# FIXME: make patchset service configuration option
+ALLOWED_DOMAINS = {"patchwork.kernel.org"}
+PATCHSET_SHORT_HASH_LEN = 13
+POLLING_DELAY_SECS = 30
+PATCH_TMP_FILE_PREFIX = "kernel-patch"
```
JenySadadia, line 31: You can use `kernelci.toml` to define all these configuration options.

src/patchset.py
```diff
@@ -0,0 +1,322 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022 Collabora Limited
```
JenySadadia, line 5: Please fix the year here. Also, this patch is not from `Collabora`.
```suggestion
# Copyright (C) 2024 Meta
```

src/patchset.py
```diff
@@ -0,0 +1,322 @@
...
+                    checkout_node
+                ):
+                    continue
+
+                if checkout_node["state"] == "running":
+                    self.log.info(
+                        f"Patchset node {patchset_node['id']} is waiting "
+                        f"for checkout node {checkout_node['id']} to complete",
+                    )
+                    continue
```
JenySadadia, line 281: Okay, now I understand why you are finding all the running nodes on every event.
According to the state machine and design, `checkout` node should only have children after its state changes to `available`.
Instead of manually creating `checkout` and `patchset` nodes, we can listen to `available checkout` node events in the patchset service and create a node automatically from the service itself.

JenySadadia:
@yurinnick Please rebase the PR.

src/patchset.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def _run(self, sub_id):
+        self.log.info("Listening for new trigger events")
+        self.log.info("Press Ctrl-C to stop.")
+
+        while True:
+            patchset_nodes = self._api.node.find({
+                "name": "patchset",
+                "state": "running",
+            })
```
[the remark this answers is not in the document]
yurinnick: I was under impression that I can't requeue nodes if they aren't ready to be processed. When I poll `patchset` one I only process one that have a parent `checkout` node in `done` state.

src/patchset.py
yurinnick: So, for every `patchset` node I should create a new `checkout` node and wait for it to become `available`? It will pause the queue for quite some time and seems to be quite inefficient.
However it gave me another idea. Ideally I want to apply patchset to whatever latest commit is available on a tree. Is it possible to query latest available checkout node?

yurinnick:
Rebased.

src/patchset.py
JenySadadia: Yes, that's what I am suggesting. You can listen to available checkout nodes and create patchset nodes from the service directly.
See, for instance, you can use the below filter [here](https://github.com/yurinnick/kernelci-pipeline/blob/patchset-service/src/patchset.py#L154) to get available checkout node events:
```
def _setup(self, *args):
        return self._api_helper.subscribe_filters({
            "name": "checkout",
            "state": "available",
        })
```
Then you can get `checkout` node from the event just like https://github.com/kernelci/kernelci-pipeline/blob/main/src/tarball.py#L130 and then you can create `patchset` node from it.

yurinnick:
Done:
- Rebase
- Pass all `args` into `self._service_config` instead of assigning individual values
- Rename `configs` into `global_configs` and `args` into `service_config` to better reflect their purpose
- Fix `patchset` service usage of `tarball` config
- Move `patchset` global variables into `patchset` config section
- Fix checkout dir path

Todo:
- Patchset `self._name` is still `tarball`, need to fix this
- Checkout nodes querying

src/patchset.py
yurinnick: So it should be done outside of `patchset` service, right? In an another service that will prepare `checkout` node and then pass it into `patchset`?

yurinnick: So I think the greatest challenge with checkout nodes so is picking the right one. I'm not sure how to determine an appropriate version to apply a patch and then correctly query a right node. If we have checkouts for `6.4`, `6.8` etc how to pick the right version. Maybe we should pick the latest stable and `next` branch and create two patchset nodes based on them.

a-wai:
Sorry for jumping into this discussion with naive questions, but why not extend the `tarball` service so it can generate checkout nodes with configured patchset(s) applied?

As it turns out, `kbuild` nodes currently listen for `checkout` nodes, so I'm fearing that adding another type of basically the same thing (i.e. provide a kernel source tree) could introduce unneeded complexity.

This would of course need some thinking around how to configure patchsets in `pipeline.yaml`, but it feels more intuitive to me (remember I don't have any background information/history wrt. this feature, though)

src/patchset.py
JenySadadia: It still doesn't fix the year in the copyright header.

src/patchset.py
JenySadadia: > In an another service that will prepare checkout node and then pass it to patchset?

The pipeline service `trigger` is already creating checkout nodes whenever a new commit is detected for a particular kernel revision. As I have not been involved in the patchset-related discussions, I am not aware of all the requirements, but do you think you can use `checkout` nodes created by the `trigger` service?
Once the trigger service creates a `checkout` node, the `tarball` service will create and push a tarball and change the node's status to `available`. Then you can use the below filter in the `patchset` service and create a patchset node from it:

```
    def _setup(self, *args):
        return self._api_helper.subscribe_filters({
            "op": "updated",
            "name": "checkout",
            "state": "available",
        })
    def _run(self, _sub_id):
        self.log.info("Listening for new trigger events")
        self.log.info("Press Ctrl-C to stop.")

        while True:
              checkout_node = self._api_helper.receive_event_node(sub_id)
              patchset_node = { create_patchset_node_from_checkout() }
              patchset_node = self._api.node.add(patchset_node)
              self._process_patchset(checkout_node, patchset_node)
```

yurinnick: As we discussed with @gctucker in kernelci-pipeline#295, `patchset` service only processed patchset nodes created by another service (patchwork<>kernelci connector). So patchset service should tream these nodes as created on-demand and process them accordingly.

I am open to suggestions on how to link existing checkout nodes and freshly created patchset node. I think it's a topic for a larger discussion on how we want to test kernel patches in general.

pawiecz: @JenySadadia @yurinnick I agree it will be an important next step to find the best way to connect `patchset` nodes with `checkout` ones created by the `trigger` service. Let keep the scope of this PR to just processing `patchset` nodes created by on-demand - with an assumption for an already existing parent checkout node (like in the verification steps posted in this PR).

I created #490 to track this task. @JenySadadia Please let me know if there something else that should be done before starting integration of this service.

pawiecz:
> but why not extend the `tarball` service so it can generate `checkout` nodes with configured patchset(s) applied? (snip)

@a-wai good point! I think we could fold some parts of the `patchset` service in its current form into `tarball` service like you suggested and convert/reduce this one to a ~`trigger-patchset` service as a next step.

Let's keep Patchwork integration as a separate service for now and revisit this idea once the patchset processing gets stable*ish* - I created #491 to track this.

src/patchset.py
JenySadadia: I think we should merge this PR now and improvements can be discussed as follow-ups.

yurinnick:
Rebased and fixed copyright header

yurinnick:
@a-wai this is what we can iterate to. I'd prefer `patchset` service to pick up existing node for the `next` tree and build against it.

pawiecz:
@JenySadadia Please have a look at merging this PR if there are no blockers (you're [the sole Pipeline maintainer](https://kernelci.org/docs/org/maintainers/#pipeline))

JenySadadia:
> @JenySadadia Please have a look at merging this PR if there are no blockers (you're [the sole Pipeline maintainer](https://kernelci.org/docs/org/maintainers/#pipeline))

Let's give it a go to the staging first.

JenySadadia:
Updated staging TOML and started `patchset` service. Staging looks OK:
```
today at 14:56:4403/28/2024 09:26:44 AM UTC [INFO] Listening for new trigger events
today at 14:56:4403/28/2024 09:26:44 AM UTC [INFO] Press Ctrl-C to stop.
today at 14:56:4703/28/2024 09:26:47 AM UTC [INFO] Waiting 30 seconds for a new nodes...
```

JenySadadia (approved):

PR #343, opened 2023-10-23
src/regression_tracker: re-format regression data for submission

hardboprobot:
Sync the regression data format to the API model definition.

### DIFF
--- a/src/regression_tracker.py
+++ b/src/regression_tracker.py
@@ -38,8 +38,10 @@ class RegressionTracker(Service):
         regression = {}
         for field in self._regression_fields:
             regression[field] = failed_node[field]
-        regression['parent'] = failed_node['id']
-        regression['regression_data'] = [last_successful_node, failed_node]
+        regression['data'] = {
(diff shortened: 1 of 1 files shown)


src/regression_tracker.py
```diff
@@ -38,8 +38,10 @@ def _create_regression(self, failed_node, last_successful_node):
         regression = {}
         for field in self._regression_fields:
             regression[field] = failed_node[field]
-        regression['parent'] = failed_node['id']
-        regression['regression_data'] = [last_successful_node, failed_node]
+        regression['data'] = {
+            'testrun_failed': failed_node['id'],
+            'testrun_passed': last_successful_node['id'],
```
gctucker, line 43: I think `testrun` is not a very good name, it's not a real word and more importantly it implies this is about tests.  Regressions with node results can be about anything, not just tests.  Something like `node_pass` and `node_fail` might be better.

hardboprobot: Good point

PR #346, opened 2023-10-23
config/pipeline.yaml: replace api_configs with api

gctucker:
The YAML configuration is being finalised and api_configs is now renamed to api.  Update pipeline.yaml accordingly.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -12,7 +12,7 @@ _anchors:
     state: available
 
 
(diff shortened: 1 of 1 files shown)


hardboprobot (approved):

hardboprobot (commented):
Looks trivial but it probably isn't. If merging this may cause other pending PRs to break and go through refactoring, prioritize them over this one and merge this after.

PR #347, opened 2023-10-23
config/pipeline.yaml: rename storage_configs to storage

gctucker:
The YAML pipeline configuration is getting finalised and storage_configs is being renamed to storage.  Update pipeline.yaml accordingly.


hardboprobot (approved):
Looks trivial but it probably isn't. If merging this may cause other pending PRs to break and go through refactoring, prioritize them over this one and merge this after.

PR #348, opened 2023-10-31
kube/aks: add restart.sh to restart all the pods

gctucker:
Add a basic restart.sh script to restart all the pipeline pods.


nuclearcat:
Thanks!

PR #349, opened 2023-10-31
config/pipeline.yaml: add context to qemu platform/device_type definition

pawiecz:
This patch adds required configuration for LAVA TestJob definition for qemu device_type. Context section was converted to a conditional one in the base template as a part of kernelci/kernelci-core#2153

This patch allows generating equivalent LAVA TestJob definitions.


nuclearcat:
Tested on staging

PR #352, opened 2023-11-08
pipeline.yaml: Add Chromebook for testing on x86-board kernels

nuclearcat:
Add first Chromebook in pipeline testing

Depends-on: https://github.com/kernelci/kernelci-core/pull/2176

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -178,6 +178,11 @@ device_types:
     boot_method: grub
     mach: x86
 
+  asus-cx9400-volteer:
+    arch: x86_64
+    boot_method: depthcharge
+    mach: x86
+
   kubernetes:
     base_name: kubernetes
     class: kubernetes
@@ -199,6 +204,16 @@ scheduler:
     platforms:
       - qemu-x86
       - minnowboard-turbot-E3826
+
+  - job: baseline-x86-board
+    event:
+      channel: node
+      name: kbuild-gcc-10-x86-board
+      result: pass
+    runtime:
+      type: lava
+    platforms:
+      - asus-cx9400-volteer
 
   - job: kbuild-gcc-10-x86
     event: *checkout-event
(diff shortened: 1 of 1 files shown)


pawiecz:
I should be able to verify this change (to certain extent) in my dev setup - let me check this patch and report back.

nuclearcat:
> I should be able to verify this change (to certain extent) in my dev setup - let me check this patch and report back.

It depends also on https://github.com/kernelci/kernelci-core/pull/2176 , i will add it on PR

gctucker:
It would be really good to complete the job configuration work first, so we don't have to fill the YAML file with all the possible combinations we'll want to run.  So while it's probably a good idea to add a couple of platforms like this, before adding any more the features should be implemented for this to scale with wildcards or regexes or something other than everything hard-coded.

pawiecz:
Unfortunately verification of this patch allowed me only to confirm the behavior observed on the staging instance.

Here's what I've done:
- shorten pipeline to depend on `kver` instead if `kbuild-*` job (no K8s in the local setup)
- run pipeline with `kernelci/staging-kernelci` images (requires LAVA submission credentials)
- run pipeline with custom container images [0] with mocked LAVA TestJob submission

This results in running `baseline-x86` job, but not the `-board` one. I'll dig into triggering mechanisms unless that's something @nuclearcat's working already (I don't want to duplicate work).

We can also split the investigation, e.g. I'll have a look bottom-up (scheduler-lava -> scheduler -> pipeline) while you check top-down (pipeline -> checkout -> kbuild -> trigger) if that makes sense - let me know what would work best for you.

[0] uploaded to: https://github.com/pawiecz/kernelci-core/tree/add-depthcharge
edited: `config/docker/fragment/kernelci.jinja2` and later pipeline's `docker-compose.yaml`
rebuilt with: `./kci docker build kernelci --prefix=boards/ --no-cache`

pawiecz:
Verified together with kernelci/kernelci-core#2176 (and a minor fixup PR) on a local instance:

1. LAVA job definition was properly rendered
2. Scheduler-LAVA job was triggered and mock TestJob sumission was run
3. LAVA TestJob completed successfully: https://lava.collabora.dev/scheduler/job/12007421

LGTM (after fixup is pulled and squashed to the patch)

PR #353, opened 2023-11-08
kbuild.jinja2: Add olddefconfig step to build

nuclearcat:
As sometimes defconfig and fragments might miss some options, we need to make sure at compile time config process wont restart and try to ask questions by using olddefconfig.

### DIFF
--- a/config/runtime/kbuild.jinja2
+++ b/config/runtime/kbuild.jinja2
@@ -101,6 +101,16 @@ class Config:
                 print(f"[_merge_frags] Failed to merge fragment")
             print(f"[_merge_frags] Merging fragment done")
             os.remove(f"/tmp/{i}.config")
+    
+    def olddefconfig(self, srcdir):
+        cwd = os.getcwd()
+        os.chdir(srcdir)
+        print(f"[_olddefconfig] Running olddefconfig")
+        cmd = f"make ARCH=x86_64 olddefconfig"
+        ret = subprocess.run(cmd, shell=True).returncode
+        if ret != 0:
+            print(f"[_olddefconfig] Failed to run olddefconfig")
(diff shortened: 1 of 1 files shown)


config/runtime/kbuild.jinja2
```diff
@@ -101,6 +101,16 @@ class Config:
...
+    
+    def olddefconfig(self, srcdir):
+        cwd = os.getcwd()
+        os.chdir(srcdir)
+        print(f"[_olddefconfig] Running olddefconfig")
+        cmd = f"make ARCH=x86_64 olddefconfig"
+        ret = subprocess.run(cmd, shell=True).returncode
+        if ret != 0:
+            print(f"[_olddefconfig] Failed to run olddefconfig")
+        print(f"[_olddefconfig] Running olddefconfig done")
```
gctucker, line 113: That's a lot of debug noise though, maybe it's useful right now but it should really be cleaned up at some point.  The job log should contain all the necessary information.

gctucker (commented):
Please take a look at what the legacy build jobs are doing.  It seems they always run `make olddefconfig` with CIP and ChromeOS fragments but I'm not sure about the others as they rely on `merge_config.sh` to be taking care of this already.

config/runtime/kbuild.jinja2
nuclearcat: Yes, it is just debug at moment to get build steps right

nuclearcat:
> Please take a look at what the legacy build jobs are doing. It seems they always run `make olddefconfig` with CIP and ChromeOS fragments but I'm not sure about the others as they rely on `merge_config.sh` to be taking care of this already.

Good point, i will investigate more what is happening, cause i think by some reason it is happening with x86-board fragment as well, while it shouldn't.

PR #355, opened 2023-11-15
kbuild.jinja2: Keep modules name simpler and neutral

nuclearcat:
modules.tar.xz complicates jinja templates and probably add some inflexibility in future, if we decide to change compression format. It is better to avoid also dots in artifact name.


gctucker (commented):
Also the artifacts might be improved with some meta-data like we have with `artifacts.json` in the legacy system, to help avoid such issues in the future.

nuclearcat:
> Also the artifacts might be improved with some meta-data like we have with `artifacts.json` in the legacy system, to help avoid such issues in the future.

Yes, definitely it needs rework as whole kbuild, i just want to make sure all basic functions works and then after i complete deploy - it is next thing to do.

nuclearcat:
Tested on staging

PR #356, opened 2023-11-15
pipeline.yaml: Add one more lab and board

nuclearcat:
As good example for future improvements and to detect bugs in corner case we can add also one more lab and device. This is also unfortunately required to add one more real hardware device, as production lab doesn't support properly Azure files (yet), and it might be also good case to improve (detect infrastructure failure).

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -91,6 +91,16 @@ runtimes:
         token: kernelci-api-token-early-access
         url: https://staging.kernelci.org:9100
 
+  lava-collabora-staging:
+    lab_type: lava
+    url: https://staging.lava.collabora.dev/
+    priority_min: 40
+    priority_max: 60
+    notify:
+      callback:
+        token: kernelci-api-token-lava-staging
+        url: https://staging.kernelci.org:9100
+
   shell:
     lab_type: shell
 
@@ -108,6 +118,9 @@ jobs:
     template: baseline.jinja2
 
   baseline-x86-board:
+    <<: *baseline-x86-job
+
+  baseline-x86-board-staging:
     <<: *baseline-x86-job
 
   kbuild-gcc-10-x86:
@@ -190,6 +203,11 @@ device_types:
     boot_method: depthcharge
     mach: x86
 
+  dell-latitude-3445-7520c-skyrim:
+    arch: x86_64
+    boot_method: depthcharge
+    mach: x86
+
   kubernetes:
     base_name: kubernetes
     class: kubernetes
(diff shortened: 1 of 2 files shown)


gctucker (commented):
I think @broonie also volunteered to have his lab enabled with early access.

gctucker:
About this particular PR, @pawiecz can confirm but the Collabora LAVA staging lab shouldn't be used for production use-cases.  So while it's probably OK to have it right now, it should be removed or somehow treated differently when the new API reaches actual production status.

nuclearcat:
Yes, if this works we can try to enable @broonie as well. At moment i am enabling it because Azure caching fix will be deployed on production only at December, so i need something else to test on.

gctucker:
What Azure caching fix?  Is this something on the Collabora LAVA lab side?

nuclearcat:
It is more incompatibility of Azure Files with caching systems.

Azure Files have certain implementation bugs:
https://github.com/MicrosoftDocs/azure-docs/blob/main/includes/storage-files-condition-headers.md
https://stackoverflow.com/questions/43706605/azure-file-storage-error-condition-headers-are-not-supported

gctucker:
Right but is the fix going to be deployed in the LAVA lab or on the Azure side?

nuclearcat:
LAVA lab, its more workaround than fix.

gctucker:
OK thanks.  Do you know if other labs may be impacted as well, if this is a widely used caching mechanism?

nuclearcat:
Yes, it is typical caching mechanism, and they might need to deploy similar workaround or we might need to change storage solution.

gctucker:
Right, so that's something that would be ideally addressed before making the first production deployment.  I believe one possibility mentioned in the past was to use MinIO on top of Azure storage.

Is there a GitHub issue about this?  There is definitely one about deciding which storage solution to use for production and which ones to support in general, but I don't remember seeing one about this particular caching problem.

nuclearcat:
It is mentioned since October: https://github.com/kernelci/kernelci-api/issues/381

gctucker:
RIght, so there isn't any GitHub issue about this particular problem.  Probably it would be worth creating one and resolving it before the first production deployment happens (in early January I guess).

nuclearcat:
381 is exactly about that, storage options. Probably i should add backend i tested recently and developed in go, as it is backward compatible with old one (means already implemented, i tested - it works), i will do that as will sort out current tasks.

gctucker:
Right - there just isn't an issue specifically about the fact that we can't use Azure Files with caching in LAVA.  Never mind me...

Issue #358, opened 2023-11-27
Unauthorized error should cause containers to quit or (better) re-read authentication information


nuclearcat:
In some cases token might get updated in k8s or docker container, but in current case we have following problem:
```
11/24/2023 04:01:46 PM UTC [INFO] Existing revision                kernelci_staging-stable          5be3078c819c6de773b07dba8bd45bb40f2caa96
11/24/2023 04:01:47 PM UTC [INFO] Existing revision                mainline                         f1a09972a45ae63efbd1587337c4be13b1893330
11/24/2023 04:01:47 PM UTC [INFO] Sleeping for 3600s
11/24/2023 05:01:47 PM UTC [INFO] New revision                     kernelci_staging-mainline        7b8e21d60e895185fb8bdcedf51b4161d0305d60
11/24/2023 05:01:47 PM UTC [ERROR] Unauthorized
11/24/2023 05:01:47 PM UTC [INFO] Existing revision                kernelci_staging-next            1c98619b589f17c2c6ee53abc55350f138e359bb
11/24/2023 05:01:47 PM UTC [INFO] Existing revision                kernelci_staging-stable          5be3078c819c6de773b07dba8bd45bb40f2caa96
11/24/2023 05:01:48 PM UTC [INFO] Existing revision                mainline                         f1a09972a45ae63efbd1587337c4be13b1893330
```
trigger process keep failing on Unauthorized, but not rereading token, staying non-functional.

gctucker:
In principle, when something like a token is revoked or replaced, then a production redeployment should happen to enable the new configuration.  Then the services would be restarted.  I think this is just one of the many corner-cases where something can go wrong but isn't part of the normal workflow.

JenySadadia:
One solution is to restart all the services whenever `.env` file changes are detected. The drawback is that if we use this kind of automation, the user will be unaware of when the services got restarted. That would also cause abrupt termination of ongoing tests or build execution.
I think the best way is to manually restart services when required instead of automating the process.

JenySadadia:
@nuclearcat One solution is to monitor `.env` file and restart all the services if file change is detected.
I created a script using `inotifywait` tool.
Here is the draft PR https://github.com/kernelci/kernelci-pipeline/pull/380

nuclearcat:
[...]

PR #359, opened 2023-11-27
trigger.py: Quit, if error 401

nuclearcat:
In some cases token might get updated, but to re-read it we need to restart container.
In future we need to implement re-read of token only.

Improves https://github.com/kernelci/kernelci-pipeline/issues/358

### DIFF
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -72,6 +72,11 @@ class Trigger(Service):
             detail = ex.response.json().get('detail')
             if detail:
                 self.log.error(detail)
+            # TODO(nuclearcat): Re-read authorization token if it's failing
+            # if error Unauthorized 401 - quit
+            if ex.response.status_code == 401:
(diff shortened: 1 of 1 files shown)


gctucker (commented):
I think this is a bad idea, there could be lots of reasons why this happens and if the trigger just quits it will never retry again (unless we're lucky and the infra restarts the service, but we don't know about that in the service itself).  Reporting the error is more important, e.g. with an email or something (but not sending an email _from_ the service obviously).

nuclearcat:
Agree, better i implement re-reading secret file.

gctucker:
Well I wouldn't even do that, just make sure the HTTP errors are well propagated.  Reloading the API token is just a wild guess that it's the cause of the error, there's an infinite number of reasons why things like this might fail.  In a normal scenario, the token doesn't need to be reloaded as when it changes the services should be redeployed.

PR #360, opened 2023-11-27
docker-compose.yaml: Restart on failure

nuclearcat:
In kubernetes it is expected containers might fail and quit, so we can duplicate this behavior in docker as well. While failure is not normal and we should avoid it, but we should not assume it is critical issue in container environment.

Failure might be expected in situations like this: https://github.com/kernelci/kernelci-pipeline/issues/358
Or for example if API service was temporary unreachable.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -10,6 +10,7 @@ services:
 
   monitor: &base-service
     container_name: 'kernelci-pipeline-monitor'
+    restart: on-failure
     image: 'kernelci/staging-kernelci'
     env_file: ['.env']
     stop_signal: 'SIGINT'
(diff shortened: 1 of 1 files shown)


gctucker (commented):
I think the reason why this wasn't enabled initially is because the docker-compose deployment is mostly meant to be used by developers, and it's better to not automatically restart services when they're failing while working on the code as it just gets in the way.

Whatever works best for developers.  If this was to address production use-cases then I don't think it's a good idea.  If it's because developers generally find it useful then it probably makes more sense.

@rcn What do you think?

nuclearcat:
I think I agree with the collective and reasoned opinion. The only problem, like recently during staging issues, when the server was overloaded, all the containers “died” due to the fact that containers couldn't contact the API server or due to problems with DNS resolving. On the other hand, they should not “crash” if they experience temporary problems with the network...

PR #363, opened 2023-11-28
src/monitor: print info only for node events with revision info

hardboprobot:
This should be safe to merge, it shouldn't break anything.

Related to https://github.com/kernelci/kernelci-core/pull/2212
Both PRs try to get rid of crashes when non-node events are published.

Note that these are provisional fixes. A proper fix would involve a better use of pydantic checks, and for that we'd need to move the models to kernelci-core.

### DIFF
--- a/src/monitor.py
+++ b/src/monitor.py
@@ -61,14 +61,18 @@ class Monitor(Service):
             event = self._api.receive_event(sub_id)
             obj = event.data
             dt = datetime.datetime.fromisoformat(event['time'])
-            print(self.LOG_FMT.format(
-                time=dt.strftime('%Y-%m-%d %H:%M:%S.%f'),
-                commit=obj['revision']['commit'][:12],
-                id=obj['id'],
-                state=state_map[obj['state']],
-                result=result_map[obj['result']],
-                name=obj['name']
-            ), flush=True)
+            # Print the status line only for events that contain
+            # "revision" data. Implement proper event filtering when the
+            # models are production-ready
+            if 'revision' in obj:
+                print(self.LOG_FMT.format(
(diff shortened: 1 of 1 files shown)


src/monitor.py
```diff
@@ -61,14 +61,18 @@ def _run(self, sub_id):
...
-                commit=obj['revision']['commit'][:12],
-                id=obj['id'],
-                state=state_map[obj['state']],
-                result=result_map[obj['result']],
-                name=obj['name']
-            ), flush=True)
+            # Print the status line only for events that contain
+            # "revision" data. Implement proper event filtering when the
+            # models are production-ready
+            if 'revision' in obj:
```
JenySadadia, line 67: Same comment as https://github.com/kernelci/kernelci-core/pull/2212.
Events on `node` channel will always contain `revision` field.
What are the non-node events?

hardboprobot: Sorry, I used the wrong words in the description. I meant node events that aren't well formatted. This is a workaround to prevent the monitor from crashing when it receives a node event that's not formatted as expected. But it's not a good one, as it could still crash if any of the other referenced node fields are missing. If this isn't urgent I'd rather find the proper way to sanitize these checks all the way from the endpoint inputs and through the API.

hardboprobot: I agree. If this isn't urgent I might as well close the PR and we'll figure out how to validate all inputs properly, although I think we'll need more validation and sanity checks throughout, not only at the endpoints.

JenySadadia: Sure. It would be nice if you could create a Github issue for this to keep track.

nuclearcat:
Do we still need this PR?

PR #364, opened 2023-12-05
trigger.py: Check latest commit nodes owned by token owner

nuclearcat:
We might have several different instances, and owners should not interfere with each other.

### DIFF
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -28,14 +28,17 @@ class Trigger(Service):
     def __init__(self, configs, args):
         super().__init__(configs, args, 'trigger')
         self._build_configs = configs['build_configs']
+        self._current_user = self._api.whoami()
 
     def _log_revision(self, message, build_config, head_commit):
         self.log.info(f"{message:32s} {build_config.name:32s} {head_commit}")
 
     def _run_trigger(self, build_config, force, timeout):
         head_commit = kernelci.build.get_branch_head(build_config)
+
         node_count = self._api.count_nodes({
             "revision.commit": head_commit,
+            "owner": self._current_user['username'],
         })
 
         if node_count > 0:


gctucker:
I think ideally it should actually be filtering events based on a group name.  Several instances or trigger services could be running as part of the same extended pipeline.  But just filtering based on the owner is fine as a starting point.

src/trigger.py
```diff
@@ -34,8 +34,10 @@ def _log_revision(self, message, build_config, head_commit):
 
     def _run_trigger(self, build_config, force, timeout):
         head_commit = kernelci.build.get_branch_head(build_config)
+        whoami = self._api.whoami()
```
gctucker, line 37: This could be one only once when the service starts, there's already a class to store the `whoami` data.  No need to keep asking the API if I'm still user, unless an identity crisis is likely to be triggered rather than a new kernel revision ;)

If something as critical as the username of the user running the pipeline changes, I would expect a full synchronous redeploy.

gctucker (commented):
LGTM overall - not tested, just one small comment.

src/trigger.py
nuclearcat: Ok, i will update code, agree as it might significantly increase load on API if it will poll whoami each time

src/trigger.py
```diff
@@ -28,14 +28,17 @@ class Trigger(Service):
     def __init__(self, configs, args):
         super().__init__(configs, args, 'trigger')
         self._build_configs = configs['build_configs']
+        self._whoami = self._api.whoami()
```
JenySadadia, line 31: How about renaming the variable holding the current user's information?
```suggestion
        self._current_user = self._api.whoami()
```

nuclearcat: Changed

PR #365, opened 2023-12-05
Changes to sync to the new node models

hardboprobot:
Companion PR to https://github.com/kernelci/kernelci-api/pull/395
Also related: https://github.com/kernelci/kernelci-core/pull/2224

### DIFF
--- a/src/regression_tracker.py
+++ b/src/regression_tracker.py
@@ -19,10 +19,7 @@ class RegressionTracker(Service):
 
     def __init__(self, configs, args):
         super().__init__(configs, args, 'regression_tracker')
-        self._regression_fields = [
-            'artifacts', 'group', 'name', 'path', 'revision', 'result',
-            'state',
-        ]
+        self._regression_fields = ['group', 'name', 'path']
 
     def _setup(self, args):
         return self._api_helper.subscribe_filters({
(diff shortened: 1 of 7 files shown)


src/regression_tracker.py
```diff
@@ -38,19 +35,25 @@ def _create_regression(self, failed_node, last_successful_node):
         regression = {}
         for field in self._regression_fields:
             regression[field] = failed_node[field]
-        regression['parent'] = failed_node['id']
-        regression['regression_data'] = [last_successful_node, failed_node]
-        self._api_helper.submit_regression(regression)
+
+        regression['kind'] = 'regression'
```
JenySadadia, line 39: This line can be eliminated as the default value for `Regression.kind` is already `regression`.

hardboprobot: This has changed now. All subclasses of `Node` are now `POST`ed and stored as plain `Node` objects, and their type is encoded in the `kind` field (see https://github.com/kernelci/kernelci-api/pull/395 and https://github.com/kernelci/kernelci-core/pull/2224). A next step, though, is to move parts of this code to the kernelci-core helpers, now that the models are there after refactoring. But we better do that in a next PR.

nuclearcat (approved):
Tested on staging manually, followed logs on each stage

PR #367, opened 2023-12-07
scheduler.py: Try to not fail if submit have exception

nuclearcat:
We are having quite often errors on LAVA submit, and this will cause scheduler to crash.
This is generic attempt to keep running on such failures, which might be handled better in future.

### DIFF
--- a/src/scheduler.py
+++ b/src/scheduler.py
@@ -80,7 +80,18 @@ class Scheduler(Service):
         data = runtime.generate(job, params)
         tmp = tempfile.TemporaryDirectory(dir=self._output)
         output_file = runtime.save_file(data, tmp.name, params)
-        running_job = runtime.submit(output_file)
+        try:
+            running_job = runtime.submit(output_file)
+        except Exception as e:
+            self.log.error(' '.join([
+                node['id'],
+                runtime.config.name,
+                platform.name,
+                job_config.name,
+                str(e),
+            ]))
+            return
+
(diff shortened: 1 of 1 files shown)


src/scheduler.py
```diff
@@ -80,7 +80,18 @@ def _run_job(self, job_config, runtime, platform, input_node):
         data = runtime.generate(job, params)
         tmp = tempfile.TemporaryDirectory(dir=self._output)
         output_file = runtime.save_file(data, tmp.name, params)
-        running_job = runtime.submit(output_file)
+        try:
+            running_job = runtime.submit(output_file)
+        except Exception as e:
+            self.log.error(' '.join([
```
pawiecz, line 86: It would probably be neater to collect node/runtime/job data first and append it with TestJob ID/error message for logging but it could be the next step

pawiecz (approved):
Could you point to the failing nodes? It would be useful to eliminate root cause of such issues in the future.

nuclearcat:
> Could you point to the failing nodes? It would be useful to eliminate root cause of such issues in the future.

yesterday at 11:39:02 PM12/07/2023 09:39:02 PM UTC [ERROR] 65723b76ea59c4c039f324ff lava-collabora-staging minnowboard-turbot-E3826 baseline-x86 400 Client Error: Bad Request for url: https://staging.lava.collabora.dev/api/v0.2/jobs/?format=json&limit=256

PR #368, opened 2023-12-08
pipeline: add "octopus" Chromebook

a-wai:
This enables testing on an additional device type. `octopus` is a good candidate as it's known to work fine with the legacy pipeline and also uses a pineview chip.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -208,6 +208,11 @@ device_types:
     boot_method: depthcharge
     mach: x86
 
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
Hurray!!!
```
today at 5:03:55 PM12/08/2023 03:03:55 PM UTC [INFO] 6573305a8a899924a8a75d50 lava-collabora hp-x360-12b-ca0010nr-n4020-octopus baseline-x86-board 12220468
```
https://lava.collabora.dev/scheduler/job/12220468
It boots and it works!

PR #369, opened 2023-12-13
kernelci.toml: Adjust default values to make local setup easier

nuclearcat:
In pipeline.yaml we have:
```
- 'data/output/:/home/kernelci/data/output'
```
Let's align it with kernelci.toml, as it might bite if we run local docker runner for kbuild.

### DIFF
--- a/config/kernelci.toml
+++ b/config/kernelci.toml
@@ -14,7 +14,7 @@ output = "/home/kernelci/data/output"
 storage_config = "docker-host"
 
 [scheduler]
-output = "/home/kernelci/output"
+output = "/home/kernelci/data/output"
 
 [notifier]
 


config/kernelci.toml
```diff
@@ -14,7 +14,7 @@ output = "/home/kernelci/data/output"
 storage_config = "docker-host"
 
 [scheduler]
-output = "/home/kernelci/output"
+output = "/home/kernelci/data/output"
```
JenySadadia, line 17: Is this working as expected?
I guess we also need to adjust the mounted volume `- './data/output:/home/kernelci/output'` for scheduler service to make this work.

nuclearcat: Good point, i will prepare fresh VM to retest all this changes from zero

PR #370, opened 2023-12-19
pipeline: add more Chromebooks

a-wai:
This enables baseline testing on more x86 devices, allowing some early stress testing of the pipeline and infrastructure.


nuclearcat (approved):
Tested on staging, visible in Dozzle

PR #374, opened 2023-12-20
src/trigger.py: use new API.User class

nuclearcat:
Use the new API.User bindings class in timeout for user-related operations.


gctucker:
Surely this is a duplicate of #374?

gctucker:
Converted to Draft as the changes in the bindings haven't been merged yet.  Also I think this should be closed as a duplicate anyway, it's very confusing why it was created with commits from the other PR.

nuclearcat:
One of commits still required, so reopening PR

PR #373, opened 2023-12-20
src: update all services to use new API.User class

gctucker:
Use the new API.User bindings class in all services for user-related operations.

PR #375, opened 2023-12-20
src/trigger.py: use new API.User class

nuclearcat:
Use the new API.User bindings class in timeout for user-related operations.


gctucker:
Duplicate of #372.  Please review the PR properly and highlight the issue rather than creating other dummy ones as that's preventing me from actually fixing the problem.

gctucker:
Sorry, this is covered by #373 not #372.

PR #377, opened 2023-12-21
pipeline.yaml: Comment as we can retrieve API token from os env

nuclearcat:
We have support in LAVA runtime to retrieve API token, if it is not set explicitly in config
Depends-on: https://github.com/kernelci/kernelci-core/pull/2258

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -79,7 +79,8 @@ runtimes:
     priority_max: 60
     notify:
       callback:
-        token: kernelci-api-token-staging
+        # By default retrieve token from environment variable
+        # token: kernelci-api-token-staging
         url: https://staging.kernelci.org:9100
 
   # ToDo: avoid creating a separate Runtime entry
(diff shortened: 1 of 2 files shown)


gctucker (changes requested):
No, the token is not needed at this point.  I think this PR should be closed and not merged.

nuclearcat:
> No, the token is not needed at this point. I think this PR should be closed and not merged.
Hmm.. but:
kernelci-core/config/runtime/base/lava.jinja2
```
notify:
  callback:
    content-type: {{ notify.callback['content-type']|default('json') }}
    dataset: {{ notify.callback.dataset|default('all') }}
    method: {{ notify.callback.method|default('POST') }}
    {%- if notify.callback.token %}
    token: {{ notify.callback.token }}
```
And in lava-callback code, it retrieve token from Authentication header, and use as *api_token*.

nuclearcat:
Not needed, thanks for explanation, follow up PR https://github.com/kernelci/kernelci-pipeline/pull/381

PR #379, opened 2023-12-28
scheduler: Add additional parameters for create_job_node

nuclearcat:
It will be useful to add runtime and platform name in node, as it will help to troubleshoot tests failures, build failures and etc.


JenySadadia:
`kci job new` can be extended to use these parameters. It can be a follow-up PR.

JenySadadia:
Tested OK on staging along with https://github.com/kernelci/kernelci-core/pull/2267

JenySadadia (approved):

PR #380, opened 2024-01-03
restart_services.sh: restart pipeline service when `.env` file changes

JenySadadia:
Environment variable `KCI_API_TOKEN` is required to run pipeline services. The variable is defined in `.env` file. Add a script to monitor `.env` file using `inotifywait` tool. If the file changes are detected, that is if the token is modified, restart all the services.

### DIFF
--- a/restart_services.sh
+++ b/restart_services.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+FILE=".env"
+inotifywait -m -e close_write $FILE | while read EVENT;
+do 
+  echo $EVENT      
+  echo ".env file changes detected. Restarting pipeline services..."
+  docker-compose down
+  docker-compose up --build --no-cache
+done


JenySadadia:
`inotifywait` perfectly detects changes when tested with `nano` and `visual studio` editors. 
I have monitored `closed_write` event for file changes that will detect when a file is closed after being opened in writable mode.

The script did not work with `vim` though. I investigated and the reason is, `vim` does not overwrite a file. Instead when a file is modified, it creates a new file, writes content to it, then deletes the old file and moves the new file with the old file name. That's why the tool is not able to detect the file changes as the original gets deleted every time a file is modified with `vim`.

Hence, the script solely depends on the editor's behavior.

nuclearcat (approved):
LGTM,thanks,  we might evolve this script to ENTRYPOINT later on

pawiecz:
I might be missing background information on requirements for this script. In case list of tracked events should be extended to also cover e.g. `vim`, I'd suggest `entr` which would also cover that (but would also require some rework in the script: `ls '.env' | <loop body>`).

There's a handy [list of file watchers](https://anarc.at/blog/2019-11-20-file-monitoring-tools/) in case this topic would be revisited in future.

JenySadadia:
@pawiecz Here is the requirements for the script https://github.com/kernelci/kernelci-pipeline/issues/358

restart_services.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+FILE=".env"
+inotifywait -m -e close_write $FILE |
+while read EVENT;
+do 
+  echo $EVENT      
+  echo ".env file changes detected. Restarting pipeline services..."
+  docker-compose down
+  docker-compose up --build
```
a-wai, line 10: Might be worth it to add `--no-cache` as well?

JenySadadia: Yes, added the flag.

restart_services.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+FILE=".env"
+inotifywait -m -e close_write $FILE |
+while read EVENT;
```
a-wai, line 5:
```suggestion
inotifywait -m -e close_write $FILE | while read EVENT
```

Optional, minor nitpick as I didn't understand immediately you were feeding the while loop with the output from `inotifywait`.

a-wai (approved):

restart_services.sh
JenySadadia: Thanks. Fixed it as it improves readability.

PR #381, opened 2024-01-08
lava_callback.py: Rework lava callback authentication

nuclearcat:
This is follow up on discussion of PR:
https://github.com/kernelci/kernelci-pipeline/pull/377 https://github.com/kernelci/kernelci-core/pull/2258

Now workflow of enabling LAVA lab is simplified:
LAVA lab need to enable one or two tokens (two a bit more secure, as one will be used to submit job, another for callback), and provide name(description) of one of tokens to specify in pipeline.yaml.
Also in toml file we need to specify callback token value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,43 @@ Follow instructions to [add a token and start the services](https://kernelci.org
 
 > **Note** The `trigger` service was run only once as it's not currently configured to run periodically.
 
+### Setting up LAVA lab
+
+For scheduling jobs, the pipeline needs to be able to submit jobs to a "LAVA lab" type of runtime and receive HTTP(S) callbacks with results over "lava-callback" service.
+Runtime is configured in yaml file following way, for example:
+```
+  lava-collabora: &lava-collabora-staging
+    lab_type: lava
+    url: https://lava.collabora.dev/
+    priority_min: 40
+    priority_max: 60
+    notify:
+      callback:
+        token: kernelci-api-token-staging
+        url: https://staging.kernelci.org:9100
+```
+
+- url is endpoint of LAVA lab API where job will be submitted.
+- notify.callback.url is endpoint of lava-callback service where LAVA will send results.
+- notify.callback.token is token DESCRIPTION used in LAVA job definition. This part is a little bit tricky: https://docs.lavasoftware.org/lava/user-notifications.html#notification-callbacks
+If you specify token name that does not exist in LAVA under user submitting job, callback will return token secret set to description. If following example it will be "kernelci-api-token-staging".
+If you specify token name that matches existing token in LAVA, callback will return token value (secret) from LAVA, which is usually long alphanumeric string.
+Tokens generated in LAVA in "API -> Tokens" section. Token name is "DESCRIPTION" and token value (secret) can be shown by clicking on green eye icon named "View token hash".
+
+The `lava-callback` service is used to receive notifications from LAVA after a job has finished.  It is configured to listen on port 8000 by default and expects in header "Authorization" token value(secret) from LAVA. Mapping of token value to lab name is done over toml file. Example:
+```
+[runtime.lava-collabora]
+runtime_token = "REPLACE-LAVA-TOKEN-GENERATED-BY-LAB-LAVA-COLLABORA"
+callback_token = "REPLACE-LAVA-TOKEN-GENERATED-BY-LAB-LAVA-COLLABORA"
+
+[runtime.lava-collabora-early-access]
+runtime_token = "REPLACE-LAVA-TOKEN-GENERATED-BY-LAB-LAVA-COLLABORA-EARLY-ACCESS"
+callback_token = "REPLACE-LAVA-TOKEN-GENERATED-BY-LAB-LAVA-COLLABORA"
+```
+In case we have single token, it will be same token used to submit job(by scheduler), runtime_token only, but if we use different to tokens to submit job and to receive callback, we need to specify both runtime_token and callback_token.
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -12,6 +12,38 @@ Follow instructions to [add a token and start the services](https://kernelci.org
...
+      callback:
+        token: kernelci-api-token-staging
+        url: https://staging.kernelci.org:9100
+```
+
+- url is endpoint of LAVA lab API where job will be submitted.
+- notify.callback.url is endpoint of lava-callback service where LAVA will send results.
+- notify.callback.token is token DESCRIPTION used in LAVA job definition. This part is a little bit tricky: https://docs.lavasoftware.org/lava/user-notifications.html#notification-callbacks
+If you specify token name that does not exist in LAVA under user submitting job, callback will return token secret set to description. If following example it will be "kernelci-api-token-staging".
+If you specify token name that match existing token in LAVA, callback will return token value (secret) from LAVA, which is usually long alphanumeric string.
```
pawiecz, line 35: minor: typo:

```suggestion
If you specify token name that matches existing token in LAVA, callback will return token value (secret) from LAVA, which is usually long alphanumeric string.
```

src/lava_callback.py
```diff
@@ -56,14 +59,34 @@ def hello():
 
 @app.post('/node/<node_id>')
 def callback(node_id):
+    settings = kernelci.settings.Settings(SETTINGS_FILE)
```
pawiecz, line 62: Settings are now read both here and https://github.com/kernelci/kernelci-pipeline/pull/381/files#diff-b5a78ae1c76918daeeabe1b33399cbda922d09cdf847815dd8c4e0b00e91c0daR19

I'd add at least a TODO entry to align `SETTINGS_FILE` usage throughout this service.

src/lava_callback.py
```diff
@@ -56,14 +59,34 @@ def hello():
...
+    lab_token = request.headers.get('Authorization')
+    # return 401 if no token
+    if not lab_token:
+        return 'Unauthorized', 401
+    # iterate over tokens and check if value of one matches
+    # retrieve lab name from key
+    lab_name = None
+    for token in tokens:
+        if tokens[token] == lab_token:
+            lab_name = tokens[token]
```
pawiecz, line 79: Naming here is a bit confusing: `tokens` keeps `{'lab-name': 'lab-token'}` dict. How about:

```suggestion
    for lab, token in tokens.items():
        if token == lab_token:
            lab_name = lab
```

(or at least change assignment before lookup break to `token` - otherwise token value gets assigned instead of lab name)

pawiecz (changes requested):
I really like added docs - it would be useful to pull them to https://kernelci.org/docs site later as well.

src/lava_callback.py
nuclearcat: Yes, we have "old" way of dealing with settings in pipeline services,  and already @JenySadadia  highlighted that in our development plan.

nuclearcat: I removed kernelci.settings module for now

src/lava_callback.py
nuclearcat: I updated code in this part with new approach (as we might have 2 tokens)

nuclearcat:
Tested on staging:
today at 11:58:20 AM[pid: 8|app: 0|req: 1/2] 159.223.210.122 () {44 vars in 778 bytes} [Wed Jan 10 09:58:17 2024] POST /node/659e69dfd67d060e2dfd99ae => generated 7286 bytes in 3901 msecs (HTTP/1.1 200) 2 headers in 73 bytes (1 switches on core 0)

README.md
nuclearcat: Updated

pawiecz: I don't see the change but I'm fine to do it in another change set to this file.

src/lava_callback.py
pawiecz: Do I understand this correctly that temporary removing `kernelci.settings` module and dropping `Settings`/`Secrets` is a transitional step while models are moved around API/Core repositories?

If so, this thread can be marked as resolved but I think it would be good to bring `Secrets` back here once the transition is completed.

nuclearcat: I follow the suggestions of initial author, @gctucker , on Slack
```
I didn't implement the callback service using kernelci.settings because it wasn't a good idea, it's better to just load the TOML settings directly in this case.  It's not a command-line tool or a service but a Flask application
```
Thats actually good point, as it is web app, it is better to keep it lightweight as possible.
Also using global SETTINGS is better, as it wont re-read and parse file on each request.

README.md
```diff
@@ -12,6 +12,43 @@ Follow instructions to [add a token and start the services](https://kernelci.org
...
+      callback:
+        token: kernelci-api-token-staging
+        url: https://staging.kernelci.org:9100
+```
+
+- url is endpoint of LAVA lab API where job will be submitted.
+- notify.callback.url is endpoint of lava-callback service where LAVA will send results.
+- notify.callback.token is token DESCRIPTION used in LAVA job definition. This part is a little bit tricky: https://docs.lavasoftware.org/lava/user-notifications.html#notification-callbacks
+If you specify token name that does not exist in LAVA under user submitting job, callback will return token secret set to description. If following example it will be "kernelci-api-token-staging".
+If you specify token name that matches existing token in LAVA, callback will return token value (secret) from LAVA, which is usually long alphanumeric string.
```
nuclearcat, line 35: @pawiecz i think now it is fixed, match / matches. :) sorry i missed to commit last time

nuclearcat (commented):
Tested on staging, need to merge to fix next bug - log truncation issue

PR #383, opened 2024-01-12
config/runtime/kunit: fix artifact field names

hardboprobot:
Normalize field names, replace '.' characters with '_', as dots in fields are problematic with mongodb


nuclearcat (approved):
Tested on staging

PR #384, opened 2024-01-12
src/test_report: sync to latest Node model changes

hardboprobot:


JenySadadia:
Added `staging-skip` as it conflicts with https://github.com/kernelci/kernelci-pipeline/pull/389.

PR #386, opened 2024-01-16
lava_callback: Fix artifact name and move auth at top

nuclearcat:
From security POV authentication should be done before anything. Then, we cannot use log.txt in new data model, dot in key means nested attribute in mongodb.


nuclearcat:
Tested on staging

PR #388, opened 2024-01-16
lava_callback: Job processing can be done asynchronously

nuclearcat:
LAVA have very short timeouts for callbacks, so better to do job submission process asynchronously.


nuclearcat:
Tested on staging, now answering within 100ms and also log file is not truncated. This is intermediate solution.

PR #392, opened 2024-01-19
scheduler: catch errors in create_job_node and print them

hardboprobot:
Related to https://github.com/kernelci/kernelci-core/pull/2305

### DIFF
--- a/src/scheduler.py
+++ b/src/scheduler.py
@@ -72,8 +72,12 @@ class Scheduler(Service):
         self._cleanup_paths()
 
     def _run_job(self, job_config, runtime, platform, input_node):
-        node = self._api_helper.create_job_node(job_config, input_node,
-                                                runtime, platform)
+        try:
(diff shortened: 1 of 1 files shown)


src/scheduler.py
```diff
@@ -72,8 +72,12 @@ def _stop(self, sub_id):
...
 
     def _run_job(self, job_config, runtime, platform, input_node):
-        node = self._api_helper.create_job_node(job_config, input_node,
-                                                runtime, platform)
+        try:
+            node = self._api_helper.create_job_node(job_config, input_node,
+                                                    runtime, platform)
+        except Exception as e:
+            self.log.error(f"Error generating job node: {e}")
+            return
```
JenySadadia, line 80: The `try-catch` block is already implemented in the base class https://github.com/kernelci/kernelci-pipeline/blob/main/src/base.py#L72.

hardboprobot: True! I'm dropping this PR then

PR #399, opened 2024-01-25
docker-compose: use toml config file for scheduler-lava

a-wai:
Commit 6169407 ("docker-compose: provide kernelci.toml as settings file") changed all services to use `kernelci.toml` instead of `kernelci.conf`, except for `scheduler-lava`. This leads to the corresponding container to exit with a "File not found" error unless a `kernelci.conf` file is also present in the `config/` dir.


nuclearcat (approved):
Thanks for spotting, LGTM.

nuclearcat:
I will merge after next staging run, if everything is ok

PR #395, opened 2024-01-23
pipeline.yaml: Add arm64 chromebook builds

nuclearcat:
Add arm64 Chromebook to build as test of cross-compilation support.

PR #400, opened 2024-01-25
pipeline: add baseline tests for arm64 chromebooks

a-wai:
This PR based on #395 which provides the necessary `kbuild` job. It adds baseline tests for 2 different arm64 chromebooks on top of it so we can assess how this use-case is handled and add more devices in the future.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -20,6 +20,10 @@ _anchors:
     <<: *x86_64-device
     boot_method: depthcharge
 
+  arm64-chromebook-device: &arm64-chromebook-device
+    arch: arm64
+    boot_method: depthcharge
+
 api:
 
   docker-host:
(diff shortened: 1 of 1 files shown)


a-wai:
Tested the concept of this PR using a manually-crafted [LAVA job](https://lava.collabora.dev/scheduler/job/12629642), which passed.

a-wai:
v2:
* fixed the trogdor dtb name (it was missing the `-r1` part)
* prepended dtb names with `dtbs/` like in legacy system, that will probably help with finding the dtb in the artifacts list

nuclearcat (approved):
Tested on staging, congratulations with first arm64 chromebook: https://staging.kernelci.org:9000/viewer?node_id=65b78862f1c6fc5a4cc810b2

PR #402, opened 2024-01-26
scheduler: abort job generation if params are invalid

a-wai:
The `get_params()` function might return `None`, in which case we should abort job generation as it will fail anyway.


nuclearcat:
Tested on staging, works well.

nuclearcat (approved):
Tested on staging, works well.

PR #406, opened 2024-01-29
Debug facilities (part 1)

hardboprobot:
Requirements:

- https://github.com/kernelci/kernelci-core/pull/2326
- https://github.com/kernelci/kernelci-core/pull/2328

Add support for bypassing (dry-running) the initial pipeline stages and add a launcher script (debug_run.sh) to make it easier to run pipeline services individually in a local setup. See the commit messages and the script documentation for more info.

Running the trigger service with `--dry-run <result>` will generate checkout nodes with "dry_run" as a debug parameter and a forced result so that certain subsequent stages will simply run the logic and code generation and bypass actions that require external factors. This is meant as a way to exercise the pipeline stages locally without relying on a full staging instance.

### DIFF
--- a/config/runtime/kbuild.jinja2
+++ b/config/runtime/kbuild.jinja2
@@ -50,10 +50,14 @@ def main(args):
     build.set_storage_config(STORAGE_CONFIG_YAML)
     build.write_script("build.sh")
     build.serialize("_build.json")
-    r = os.system("bash -e build.sh")
     build2 = KBuild.from_json("_build.json")
     build2.verify_build()
-    results = build2.submit(r)
+    if NODE.get('debug') and NODE['debug'].get('dry_run'):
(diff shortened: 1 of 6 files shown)


nuclearcat:
Lets test it

nuclearcat:
Causing crash in tarball service:
```
today at 6:08:19 PM  File "/home/kernelci/pipeline/base.py", line 69, in run
today at 6:08:19 PM    status = self._run(context)
today at 6:08:19 PM             ^^^^^^^^^^^^^^^^^^
today at 6:08:19 PM  File "/home/kernelci/./pipeline/tarball.py", line 155, in _run
today at 6:08:19 PM    version, tarball_url)
today at 6:08:19 PM    ^^^^^^^
today at 6:08:19 PMNameError: name 'version' is not defined
```

nuclearcat:
do we still need this PR?

PR #415, opened 2024-02-05
scheduler.py: Add failcode and fail node when some job parameters missing

nuclearcat:
In some cases when some of job parameters doesn't match, for example dtb, we need to issue error and fail node early.
This should be registered, as in some cases this might be regression, for example if dtb got renamed.

Depends-on: https://github.com/kernelci/kernelci-core/pull/2348

### DIFF
--- a/src/scheduler.py
+++ b/src/scheduler.py
@@ -87,6 +87,14 @@ class Scheduler(Service):
                 job_config.name,
                 "Invalid job parameters, aborting...",
             ]))
+            node['state'] = 'done'
+            node['result'] = 'fail'
+            node['data']['error_code'] = 'invalid_job_params'
+            try:
+                self._api.node.update(node)
+            except requests.exceptions.HTTPError as err:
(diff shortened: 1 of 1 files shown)


src/scheduler.py
```diff
@@ -87,6 +87,14 @@ def _run_job(self, job_config, runtime, platform, input_node):
                 job_config.name,
                 "Invalid job parameters, aborting...",
             ]))
+            node['data']['state'] = 'done'
+            node['data']['result'] = 'fail'
+            node['data']['failcode'] = 'lava_job_params'
```
hardboprobot, line 92: I don't think the scheduler should mention anything lava-specific, this is supposed to work with any of the runtimes.

nuclearcat: good point, in this case it is generic code.

nuclearcat:
Tested at: https://staging.kernelci.org:9000/viewer?node_id=65c343edd3c7969a82814dba

PR #416, opened 2024-02-06
src/scheduler: remove debug log

JenySadadia:
Remove debug log for printing job node.

Fixes: d916ba8 ("src/scheduler: store k8s context in the job node")


nuclearcat (approved):
Tested on dozzle, debug removed

PR #418, opened 2024-02-07
Add initial implementation for Tast tests

a-wai:


a-wai:
Depends on https://github.com/kernelci/kernelci-core/pull/2356

a-wai:
Marking as ready as it was successfully tested on `puff`: https://lava.collabora.dev/scheduler/job/12768684

Dropping the actual tast test job so it can be worked on in a separate PR: #435

PR #420, opened 2024-02-08
timeout.py: Fix result, error_code for nodes that timed out

nuclearcat:
In case of LAVA, for example, if task was not submitted properly, or callback didn't reached, timeout service will close such node after while, but in obscure state: job_id exist, state is done, and result is nul. Let's set result as 'incomplete', fill error_code for better clarity.

Probably fixes: https://github.com/kernelci/kernelci-api/issues/470

### DIFF
--- a/src/timeout.py
+++ b/src/timeout.py
@@ -65,12 +65,20 @@ class TimeoutService(Service):
                 ))
         return recursive
 
-    def _submit_lapsed_nodes(self, lapsed_nodes, state, log=None):
+    def _submit_lapsed_nodes(self, lapsed_nodes, state, mode):
         for node_id, node in lapsed_nodes.items():
             node_update = node.copy()
             node_update['state'] = state
-            if log:
-                self.log.debug(f"{node_id} {log}")
+            self.log.debug(f"{node_id} {mode}")
+            if mode == 'TIMEOUT':
+                if node['kind'] != 'checkout':
+                    if 'data' not in node_update:
+                        node_update['data'] = {}
+                    node_update['result'] = 'incomplete'
+                    node_update['data']['error_code'] = 'node_timeout'
+                else:
+                    node_update['result'] = 'pass'
+
             try:
                 self._api.node.update(node_update)
             except requests.exceptions.HTTPError as err:


src/timeout.py
```diff
@@ -69,8 +69,18 @@ def _submit_lapsed_nodes(self, lapsed_nodes, state, log=None):
         for node_id, node in lapsed_nodes.items():
             node_update = node.copy()
             node_update['state'] = state
+            if node['kind'] != 'checkout':
+                node_update['result'] = 'incomplete'
+            else:
+                node_update['result'] = 'pass'
```
JenySadadia, line 75: As the method `_submit_lapsed_nodes` is generic for different cases such as `holdoff`, `timeout`, and `done`, we need to add a condition here to make it specific for incomplete nodes:
```suggestion 
            if log !=  "DONE":
                if node['kind'] != 'checkout':
                    node_update['result'] = 'incomplete'
                else:
                    node_update['result'] = 'pass'
```

JenySadadia, line 75: As the method `_submit_lapsed_nodes` is generic for different cases such as `holdoff`, `timeout`, and `done` nodes, we need to add a condition here to make it specific for incomplete nodes:
```suggestion 
            if log ==  "TIMEOUT":
                if node['kind'] != 'checkout':
                    node_update['result'] = 'incomplete'
                else:
                    node_update['result'] = 'pass'
```
This would make the execution flow driven by `log` variable. So, we need to make sure the correct `log` is sent to the function. Hence, the definition becomes:
```
def _submit_lapsed_nodes(self, lapsed_nodes, state, log):
```

JenySadadia: I'd also rename the `log` variable as it is no longer just being used for logging info.

nuclearcat: Any ideas/suggestions for new name of log variable?

src/timeout.py
```diff
@@ -69,8 +69,18 @@ def _submit_lapsed_nodes(self, lapsed_nodes, state, log=None):
...
             if log:
```
JenySadadia, line 76: Also, this condition can be removed here as the argument will become mandatory.

src/timeout.py
JenySadadia: How about renaming it to `mode`?

nuclearcat: done

JenySadadia (approved):
LGTM.
Let's wait until we get the error code set for timed-out node on staging.

JenySadadia:
Tested OK on staging.

PR #423, opened 2024-02-09
lava_callback.py: Handle failed jobs properly

nuclearcat:
As mentioned in https://github.com/kernelci/kernelci-api/issues/474, even if job failed, node still registered as pass, and trying to upload non-existing log file. Fix that, if job is not completed, set node status to incomplete and do not upload log file.

Fixes: https://github.com/kernelci/kernelci-api/issues/474

Depends on: https://github.com/kernelci/kernelci-core/pull/2363


nuclearcat:
Example of incomplete job:
https://staging.kernelci.org:9000/viewer?node_id=65c62b61bd8e66257ec39e5c#line=42

Issue #403, opened 2024-01-29
`kver` failure on staging


JenySadadia:
Need to debug the issue of `kver` failing on staging:
```
today at 05:36:0001/29/2024 12:06:00 AM UTC [INFO] Listening for available checkout events
today at 05:36:0001/29/2024 12:06:00 AM UTC [INFO] Press Ctrl-C to stop.
today at 05:37:3701/29/2024 12:07:37 AM UTC [INFO] 65b6ec49f1c6fc5a4cc80c69 shell shell kver 7
today at 05:38:13Getting kernel source tree...
today at 05:38:13Source directory: /tmp/kci/linux-kernelci-staging-mainline-staging-mainline-20240129.0
today at 05:38:13Running job...
today at 05:38:13Checking kernel revision
today at 05:38:13Result: fail
today at 06:37:2901/29/2024 01:07:29 AM UTC [INFO] 65b6fa51f1c6fc5a4cc80dd1 shell shell kver 8
today at 06:38:04Getting kernel source tree...
today at 06:38:04Source directory: /tmp/kci/linux-mainline-master-v6.8-rc2
today at 06:38:04Running job...
today at 06:38:04Checking kernel revision
today at 06:38:04Result: fail
```

JenySadadia:
Started investigating.

JenySadadia:
Investigation & root cause:

When `kver` job is submitted, the job node is created and sent to the runtime.

`version` information stored in the job node is as follows: 
```
{'version': '6', 'patchlevel': '8', 'extra': '-rc3-10-g6d280f4d760e'}
```
Version details in linux source Makefile: 
```
VERSION = 6
PATCHLEVEL = 8
SUBLEVEL = 0
EXTRAVERSION = -rc3
```
In the `kver` job, both the version information is compared. The job is failing as the job node version fields are `str` and `Makefile` as them as `int` (See `VERSION` and `PATCHLEVEL` fields.). That's why the `==` operator fails.

There are 2 issues here:
- Somehow pipeline service sends nodes with version fields in `str` values.
- The API should not allow the insertion of `str` values where the `pydantic` field type is specified `int`.
  The model definition of `version` is:
   ```
   class KernelVersion(BaseModel):
    """Linux kernel version model"""
    version: int = Field(
        description="Major version number e.g. 4 in 'v4.19'"
    )
    patchlevel: int = Field(
        description="Minor version number or 'patch level' e.g. 19 in 'v4.19'"
    )
[...]

PR #424, opened 2024-02-09
config/runtime/kver: workaround for kver job failure

JenySadadia:
This is a temporary fix until the actual fixes are in place in the API.

### DIFF
--- a/config/runtime/kver.jinja2
+++ b/config/runtime/kver.jinja2
@@ -51,10 +51,10 @@ class Job(BaseJob):
                 return False
 
         return (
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
Tested on staging: https://staging.kernelci.org:9000/viewer?node_id=65c601cda6c428dc601cb5e5
All good now, please open issue in project with label techdebt, so we issue proper fix later.

JenySadadia:
I think we can keep https://github.com/kernelci/kernelci-pipeline/issues/403 open until we have an actual fix.

PR #427, opened 2024-02-12
pipeline: use the new platform config for the new system

a-wai:
Switch to using the new `platforms` config instead of the legacy `device_types`.

Also remove `base_name` attributes where unnecessary (it is now inferred from the node name if left unspecified) and drop now-unneeded `class` attributes.


nuclearcat:
tested on staging

PR #431, opened 2024-02-13
lava_callback.py: Handle case when invalid node id submitted

nuclearcat:
In some cases LAVA callback might return invalid node id, we need to handle that properly, without crashing.

### DIFF
--- a/src/lava_callback.py
+++ b/src/lava_callback.py
@@ -66,6 +66,9 @@ def async_job_submit(api_helper, node_id, job_callback):
     '''
     results = job_callback.get_results()
     job_node = api_helper.api.node.get(node_id)
+    if not job_node:
+        print(f'Node {node_id} not found')
+        return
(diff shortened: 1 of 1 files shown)


src/lava_callback.py
```diff
@@ -66,6 +66,9 @@ def async_job_submit(api_helper, node_id, job_callback):
     '''
     results = job_callback.get_results()
     job_node = api_helper.api.node.get(node_id)
+    if not job_node:
+        print(f'Node {node_id} not found')
```
a-wai, line 70: Maybe add a bit more context? like `Node <id> not found for callback (<other relevant information if available>)`

nuclearcat: It is quite exceptional case, it happened with me only when i manually edited LAVA job with random node id, just need small notification if such abnormal thing will happen and to prevent lava-callback crashing on it.

nuclearcat:
Tested on staging

PR #432, opened 2024-02-14
docker-compose.yaml: Add proper mapping of gateway hostname

nuclearcat:
Using 172.17.0.1 assuming default docker subnet and single instance, that might not work in some cases.
Proper scenario is to use mapping in docker-compose file to hostname host.docker.internal.

This is part of the efforts to make local development setup easier.


nuclearcat:
Tested on staging

PR #433, opened 2024-02-14
lava_callback.py: Handle failed jobs properly

nuclearcat:
As mentioned in https://github.com/kernelci/kernelci-api/issues/474, even if job failed, node still registered as pass, and trying to upload non-existing log file. Fix that, if job is not completed, set proper node status, and dont upload non-existing log file, if it is not present.

Fixes: https://github.com/kernelci/kernelci-api/issues/474

Depends on: https://github.com/kernelci/kernelci-core/pull/2363


nuclearcat:
Tested on staging

PR #434, opened 2024-02-14
kunit.jinja2: Update node information up to date before submit

nuclearcat:
We need to update node to latest before submit, to not miss jobid and such stuff.

### DIFF
--- a/config/runtime/kunit.jinja2
+++ b/config/runtime/kunit.jinja2
@@ -142,6 +142,8 @@ cd {src_path}
     def _submit(self, result, node, api):
         # Ensure top-level name is kept the same
         result = result.copy()
+        # update node to newest
+        node = api.node.get(node['id'])
         result['node']['name'] = node['name']
         api_helper = kernelci.api.helper.APIHelper(api)
         api_helper.submit_results(result, node)


config/runtime/kunit.jinja2
```diff
@@ -142,6 +142,8 @@ cd {src_path}
     def _submit(self, result, node, api):
         # Ensure top-level name is kept the same
         result = result.copy()
+        # update node to newest
+        node = api.node.get(node['id'])
```
JenySadadia, line 146: These lines are not required as `api_helper.submit_results` already takes care of it.

nuclearcat: I had to remove it from submit_results, because artifacts get removed, for example from lava_callback, just noticed that yesterday.

As you can see in [lava-callback](https://github.com/kernelci/kernelci-pipeline/blob/e79e3743c2caeb106b99e7eaabd5904d0bca950f/src/lava_callback.py#L79) we are adding artifact to job_node, it is second argument to submit_results.
But in [submit_results](https://github.com/kernelci/kernelci-core/blob/b68f435edf231240e724d9c633ed787f4bb88e3d/kernelci/api/helper.py#L194) we are discarding everything and just using id to fetch fresh node data, which leads to complicated situation, that any added data to node will be discarded. And if we try to handle it in submit_results, it might be different in each case, which node and field have priority.
For example kunit job_node is generated directly during job creation, and doesn't contain jobid, but lava-callback have job_node retrieved "late", only when callback received and might contain job id, so it will be logical to handle this logic in each job differently, and make submit_results more simple, but flexible.

nuclearcat: https://staging.kernelci.org:9000/viewer?node_id=65cd5b8f76955310afcf8227
This is also example, where all works ok, because lava-callback naturally supply updated node data, and only kunit need refresh in the right place.

hardboprobot: Wouldn't the right solution be to fix `submit_results` instead? There's something that feels a bit off to me. Ideally, if there's a function to submit results, it should be generic for all use cases. I have the feeling that we're going in the opposite direction, adding more code to handle specific scenarios out of that function.

First of all, `submit_results` seems to do maybe too many things: it can be used to submit a hierarchy of nodes, but then here it's used with a root node that contains information that isn't in the DB, so now `submit_results` has to be aware of that? Shouldn't this be split into two processes? First, update the root node, then submit the hierarchy of children nodes?

The first thing that `submit_results` does is to fetch the root node from the DB, so it obviously isn't expecting anything new in that node. Another way to go around it would be to fetch the node from the DB, then take the root node as passed to the function and merge them properly (with any existing field in the DB node overwritten by the new values passed in the parameter node). But this needs to be documented very clearly!

Instead of moving forward with this, I'd take a couple of deep breaths and think of the possible use cases we're facing and how to address them in a way that's ergonomic rather than in a piecemeal way, finding individual fixes that might not be compatible with each other. It shouldn't take too long to find the right solution for this.

nuclearcat: At moment it is fixing bug, otherwise we will not have lava_log, so we have to come to some conclusion in reasonable time :) And i should fix this bug before deploying on production at monday.

hardboprobot: As always, I won't stop a merge from happening, just leaving my thoughts on it. It's up to you to decide how urgent this is and whether it's valuable to hold it to find a more complete solution or not

nuclearcat: I will prefer right solution, this is why i am asking

JenySadadia: > Another way to go around it would be to fetch the node from the DB, then take the root node as passed to the function and merge them properly (with any existing field in the DB node overwritten by the new values passed in the parameter node). But this needs to be documented very clearly!

I completely agree with what Ricardo has pointed out here.

@nuclearcat If you want to make a quick fix without waiting for `submit_results` to get fixed, the simplest solution would be to update the node from lava callback before calling submit results. That would preserve data in `submit_results` method as well.

nuclearcat: But that also means removing node update from submit results?
Otherwise any new data i provide with node to submit_results is getting discarded.

hardboprobot: The deep merge route is quite immediate, really. Something like this could work:

``` python
def merge(a: dict, b: dict):
    """Deep merges dicts a and b, returning a new dict containing
    dictionary a updated with the contents of dictionary b.

    """
    result = {}
    for key in a:
        result[key] = a[key]
        if key in b:
            if isinstance(a[key], dict) and isinstance(b[key], dict):
                result[key] = merge(a[key], b[key])
            else:
                result[key] = b[key]
    for key in b:
        if key not in a:
            result[key] = b[key]
    return result
```

So that in `submit_results` you can do something like this:

``` python
def submit_results(self, results, root):
    # Get latest node data added after node creation
    root_from_db = self.api.node.get(root['id'])
    root_node = merge(root_from_db, root)
    ...
```

wouldn't that work for the cases we're looking at? I still think this is a workaround rather than a fix, and this behavior (the node merging) should be very explicitly documented with a big TODO banner so we have a look at this when we have the time

nuclearcat: That still doesn't solve problem of extra API call at lava-callback and thats fundamental problem of submit_results.
In lava-callback node is _fresh_, because it does not have node, just id, when callback received.
In kunit (and similar runtimes that can submit job by themself) - they have copy of whole "stale" node created at moment of initial job/node creation that is guaranteed to not have job_id and other fields, this is why we need to fetch fresh copy of node.
I will add your solution atm into submit results just to fix bug, open issue and we will need to solve/discuss that properly.

PR #435, opened 2024-02-15
config: pipeline: add initial Tast tests

a-wai:
This change adds a `tast-quickcheck-x86` job, running on 2 Intel-based devices in order to help debug the new Tast implementation.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -1,0 +1,62 @@
+_anchors:
+
+  kbuild-gcc-10-arm64-chromeos: &kbuild-gcc-10-arm64-chromeos-job
+    template: kbuild.jinja2
+    kind: kbuild
+    image: kernelci/staging-gcc-10:arm64-kselftest-kernelci
+    params: &kbuild-gcc-10-arm64-chromeos-params
+      arch: arm64
+      compiler: gcc-10
+      cross_compile: 'aarch64-linux-gnu-'
+      cross_compile_compat: 'arm-linux-gnueabihf-'
+      defconfig: 'cros://chromeos-{krev}/{arch}/chromiumos-{arch}-generic.flavour.config'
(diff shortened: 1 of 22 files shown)


a-wai:
So, this PR ends up being a significant rework of the pipeline config as much as it is about adding Tast tests.

As such, the more comments & suggestions about this proposal, the better :)

(and given the scope of changes I've added `staging-skip` in order to ensure I don't break the whole system)

nuclearcat:
Starting staging test

config/jobs-chromeos.yaml
```diff
@@ -0,0 +1,62 @@
...
+        - x86-board
+        - CONFIG_MODULE_COMPRESS=n
+
+jobs:
+
+  baseline-arm64-chromeos-mediatek: &baseline-job
+    template: baseline.jinja2
+    kind: test
+
+  baseline-arm64-chromeos-qualcomm: *baseline-job
```
pawiecz, line 230: I might be missing something: what is the purpose of duplicating job definitions with the same body? Is it for logging/tracking progress? Wouldn't adding `job: baseline-job` in the schedule item definition cut it?

pawiecz (approved):
LGTM (except Tast-specific bits in second-to-last patch - I don't have any expertise there)

config/jobs-chromeos.yaml
a-wai: That's a good question, maybe @nuclearcat knows?

I confess I kept things as they were previously (1:1 mapping between jobs and scheduler entries), but I'd be happy to drop duplicates if we can actually schedule multiple instances of a given job.

a-wai:
Changed a bit how tast tests are listed, now depends on https://github.com/kernelci/kernelci-core/pull/2405

nuclearcat:
Tested on staging

PR #443, opened 2024-02-23
scheduler: return early if node creation skipped

a-wai:
Due to platform filters, `create_job_node()` could skip a node creation and return `None`. In such cases, we should stop processing immediately and quietly return.


nuclearcat:
Tested on staging

PR #444, opened 2024-02-23
Refine node search and regression generation

r-c-n:
Depends on https://github.com/kernelci/kernelci-core/pull/2378

Restrict the search of past results to those with the same platform, defconfig, compiler, platform, etc. than the original failed test.

Use the submit_regression helper to generate and publish the regression.


nuclearcat:
do we still need this PR?

PR #445, opened 2024-02-26
tarball: Implement gentle clean/retry mechanism

nuclearcat:
As we experienced several times git directory corruption, and it seems repeating issue, we need to implement failover mechanism, such as clean git directory and retry.

### DIFF
--- a/src/tarball.py
+++ b/src/tarball.py
@@ -55,9 +55,25 @@ class Tarball(Service):
                 return config
 
     def _update_repo(self, config):
+        '''
+        Return True - if failed to update repo and need to retry
+        Return False - if repo updated successfully
+        '''
         self.log.info(f"Updating repo for {config.name}")
-        kernelci.build.update_repo(config, self._kdir)
+        try:
+            kernelci.build.update_repo(config, self._kdir)
+        except Exception as err:
+            self.log.error(f"Failed to update: {err}, cleaning stale repo")
+            # safeguard, make sure it is git repo
+            if not os.path.exists(os.path.join(self._kdir, '.git')):
+                self.log.error(f"{self._kdir} is not a git repo")
+                raise Exception(f"{self._kdir} is not a git repo")
+            # cleanup the repo and return True, so we try again
+            kernelci.shell_cmd(f"rm -rf {self._kdir}")
+            return True
+
         self.log.info("Repo updated")
+        return False
 
     def _make_tarball(self, config, describe):
         name = '-'.join(['linux', config.tree.name, config.branch, describe])
@@ -133,7 +149,13 @@ git archive --format=tar --prefix={name}/ HEAD | gzip > {output}/{tarball}
             if build_config is None:
                 continue
 
-            self._update_repo(build_config)
+            if self._update_repo(build_config):
+                self.log.error("Failed to update repo, retrying")
+                if self._update_repo(build_config):
+                    # critical failure, something wrong with git
+                    self.log.error("Failed to update repo again, exit")
+                    os._exit(1)
+
             describe = kernelci.build.git_describe(
                 build_config.tree.name, self._kdir
             )


src/tarball.py
```diff
@@ -56,8 +56,20 @@ def _find_build_config(self, node):
...
+        try:
+            kernelci.build.update_repo(config, self._kdir)
+        except Exception as err:
+            self.log.error(f"Failed to update repo: {err}")
+            # safeguard, make sure it is git repo
+            if not os.path.exists(os.path.join(self._kdir, '.git')):
+                self.log.error(f"{self._kdir} is not a git repo")
+                return
+            # cleanup the repo and return False, so we try again
+            kernelci.shell_cmd(f"rm -rf {self._kdir}")
```
JenySadadia, line 68: Although I haven't tested it I think this will raise permission issues.

JenySadadia (commented):
This looks like a good workaround to me. I have added a few comments though.

JenySadadia:
I understand this is a workaround but shouldn't we look into the issue of why the repo update terminates abruptly sometimes?

src/tarball.py
nuclearcat: Its already tested on staging, as @pawiecz was guessing why there is no results over weekend, we had by some reason git.lock problem popped up again, so it was perfect opportunity to test this patch.
Verifying .git is just safeguard, to not rm -rf something important.

src/tarball.py
```diff
@@ -133,7 +145,13 @@ def _run(self, sub_id):
             if build_config is None:
                 continue
 
-            self._update_repo(build_config)
+            if not self._update_repo(build_config):
+                self.log.error("Failed to update repo, retrying")
+                if not self._update_repo(build_config):
+                    # critical failure, something wrong with git
+                    self.log.error("Failed to update repo, critical failure")
+                    os._exit(1)
```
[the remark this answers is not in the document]
nuclearcat: True, i will change logic, hopefully it will work, cause it is a bit hard to simulate this bug.
This failure was tested just for current patch.

src/tarball.py
JenySadadia: Okay. That's great.

src/tarball.py
```diff
@@ -56,8 +56,20 @@ def _find_build_config(self, node):
...
-        kernelci.build.update_repo(config, self._kdir)
+        try:
+            kernelci.build.update_repo(config, self._kdir)
+        except Exception as err:
+            self.log.error(f"Failed to update repo: {err}")
+            # safeguard, make sure it is git repo
+            if not os.path.exists(os.path.join(self._kdir, '.git')):
+                self.log.error(f"{self._kdir} is not a git repo")
+                return
+            # cleanup the repo and return False, so we try again
```
a-wai, line 67: This comment says "return False" and the next return is actually `return True`.

FWIW, I personnally prefer that functions return True on success and False on error (0 == success is a C thing and shouldn't work the same with booleans)

src/tarball.py
```diff
@@ -56,8 +56,20 @@ def _find_build_config(self, node):
...
         self.log.info(f"Updating repo for {config.name}")
-        kernelci.build.update_repo(config, self._kdir)
+        try:
+            kernelci.build.update_repo(config, self._kdir)
+        except Exception as err:
+            self.log.error(f"Failed to update repo: {err}")
+            # safeguard, make sure it is git repo
+            if not os.path.exists(os.path.join(self._kdir, '.git')):
+                self.log.error(f"{self._kdir} is not a git repo")
+                return
```
a-wai, line 66: Please use the same return type throughout the function

a-wai (changes requested):

src/tarball.py
nuclearcat: Also same for exit/return code of shell programs, 0 == success, so i think i follow Jeny suggestion and just fix a comment.

src/tarball.py
nuclearcat: I think in such situation better even to raise exception

nuclearcat:
Tested on staging, if i missed anything from suggestions let me know, i think implemented all of them (except function exit code, as there was matter of choice - i went with Jeny suggestion).

src/tarball.py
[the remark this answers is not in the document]
JenySadadia: This looks like unresolved.

src/tarball.py
JenySadadia: I agree with @a-wai. We usually use `return True` for success and `return False` otherwise.

nuclearcat:
I am getting confused here. Code works as following:
1)_update_repo() on failure(exception) will clean up old source directory and return true.
I am not retrying recursively in same function, because in case of insufficient disk space or memory we might end up with infinite recursion.
2)In caller we verify, if _update_repo() returned True, we need to retry once again to fetch sources, and if we failed again it means failure cannot be repaired by deleting old directory (returned True again) - then quit tarball service totally with exit code 1, to indicate failure.

src/tarball.py
a-wai: > Also same for exit/return code of shell programs, 0 == success

This is different: 0 is an int, not a bool.

The options are:
* either fully adhere to C/unix conventions and then return an int (which I think don't really make sense in python and is probably a bad idea tbh)
* or use booleans and stick to the most expected behaviour of having `True` for success and `False` for errors

nuclearcat:
I am merging as is, to not keep PR hanging and added issue to rework it later.

PR #447, opened 2024-02-28
Refine node search and regression generation (alternative)

hardboprobot:
Alternative version of https://github.com/kernelci/kernelci-pipeline/pull/444 that doesn't depend on https://github.com/kernelci/kernelci-core/pull/2378

### DIFF
--- a/src/regression_tracker.py
+++ b/src/regression_tracker.py
@@ -7,6 +7,8 @@
 
 import sys
 
+import json
+
 import kernelci
 import kernelci.config
 import kernelci.db
@@ -19,7 +21,6 @@ class RegressionTracker(Service):
 
     def __init__(self, configs, args):
         super().__init__(configs, args, 'regression_tracker')
-        self._regression_fields = ['group', 'name', 'path']
 
     def _setup(self, args):
         return self._api_helper.subscribe_filters({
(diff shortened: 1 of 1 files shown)


src/regression_tracker.py
```diff
@@ -30,44 +31,71 @@ def _stop(self, sub_id):
...
         """Method to create a regression"""
         regression = {}
-        for field in self._regression_fields:
-            regression[field] = failed_node[field]
-
         regression['kind'] = 'regression'
+        regression['name'] = failed_node['name']
+        regression['path'] = failed_node['path']
+        regression['group'] = failed_node['group']
+        regression['state'] = 'done'
```
JenySadadia, line 41: I had added  `self._regression_fields` for easy field assignment. May I know the reason for dropping it?

src/regression_tracker.py
```diff
@@ -30,44 +31,71 @@ def _stop(self, sub_id):
...
         regression['data'] = {
             'fail_node': failed_node['id'],
-            'pass_node': last_successful_node['id'],
+            'pass_node': last_pass_node['id'],
+            'arch': failed_node['data'].get('arch'),
+            'defconfig': failed_node['data'].get('defconfig'),
+            'compiler': failed_node['data'].get('compiler'),
+            'platform': failed_node['data'].get('platform'),
```
JenySadadia, line 48: We can create a new list with `Node.data` fields and use it to assign these values the same way as with `self._regression_fields`.

JenySadadia (commented):
Some minor comments. LGTM otherwise.

src/regression_tracker.py
hardboprobot: Some of the fields to match against (the specific ones inside `data`) are nested, so traversing the fields in `self._regression_fields` wouldn't be as easy as going through a list. Considering that this is only done once in the code (`self._regression_fields` isn't referenced anywhere else) I though it'd be simpler to discard it.

src/regression_tracker.py
hardboprobot: My two reasons for not doing that are that for this code, where these fields are referenced only once, during the regression creation, factoring them out to `self._regression_fields`  wouldn't be of any gain. The other reason is that this is supposed to be eventually replaced by a helper, as done in https://github.com/kernelci/kernelci-pipeline/pull/444, so this is just an intermediate step to that.

src/regression_tracker.py
JenySadadia: Yes, that's why suggested having a different list for data fields.
I'd do something like this:
```
        for field in self._regression_data_fields:
            regression['data'][field] = failed_node['data'].get(field)
```
IMO that would reduce lines of code and it's also easy while adding more data fields. This would just require adding the field to the list.
If we decide to replace this with the helper function, we can use this same logic in that function as well.

src/regression_tracker.py
JenySadadia: Any reason for creating this alternate PR?

src/regression_tracker.py
hardboprobot: The reason to use a helper function for this is to avoid coding this logic in the client code, ie. the logic is in the models: see https://github.com/kernelci/kernelci-pipeline/pull/444, https://github.com/kernelci/kernelci-core/pull/2378 and https://github.com/kernelci/kernelci-core/blob/82c15a524f28feb0fd6beed53a3770943550903e/kernelci/api/models.py#L483

src/regression_tracker.py
hardboprobot: Yes, https://github.com/kernelci/kernelci-pipeline/pull/444 depends on https://github.com/kernelci/kernelci-core/pull/2378, which can't be merged until staging runs on python 3.10

src/regression_tracker.py
hardboprobot: Just to clarify: the code to generate regressions is already done in the regression model class and the helper that uses that code is already written, only it can't be merged yet and we can't use it. So it won't be necessary to translate any of the regression generation logic elsewhere, that's been done already.

JenySadadia: Oh okay. So it's already been merged.
Got it. Sorry for the noise.

JenySadadia (approved):
LGTM. Need to verify on staging.

PR #450, opened 2024-03-01
pipeline.yaml: Use new kubernetes cluster

nuclearcat:
As discussed on weekly, we are able to run inexpensive cluster, separate for staging at least. Let's try to use it.


nuclearcat:
Tested on staging

PR #453, opened 2024-03-01
config: runtime: adapt to python template changes

a-wai:
The `base/python.jinja2` template has evolved with changes to the `Job` class API. Reflect and leverage those changes in both the `kunit` and `kbuild` templates.

Depends on https://github.com/kernelci/kernelci-core/pull/2428


nuclearcat:
Tested on staging

PR #454, opened 2024-03-04
kbuild.jinja2: Add dtbs_check parameter

nuclearcat:
Add dtbs_check as parameter for kbuild

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -179,6 +179,16 @@ jobs:
   kbuild-gcc-10-arm64:
     <<: *kbuild-gcc-10-arm64-job
 
+  kbuild-gcc-10-arm64-dtbscheck:
+    <<: *kbuild-gcc-10-arm64-job
+    kind: test
+    params:
+      <<: *kbuild-gcc-10-arm64-params
+      dtbs_check: true
+    rules:
+      tree: next
+      branch: 'master'
+
   kbuild-gcc-10-armel:
     <<: *kbuild-job
     image: kernelci/staging-gcc-10:arm64-kselftest-kernelci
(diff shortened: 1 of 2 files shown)


config/pipeline.yaml
```diff
@@ -179,6 +179,15 @@ jobs:
   kbuild-gcc-10-arm64:
     <<: *kbuild-gcc-10-arm64-job
 
+  kbuild-gcc-10-arm64-dtbscheck:
+    <<: *kbuild-gcc-10-arm64-job
+    params:
+      arch: arm64
+      compiler: gcc-10
+      cross_compile: 'aarch64-linux-gnu-'
+      defconfig: defconfig
```
a-wai, line 188:
```suggestion
      <<: *kbuild-gcc-10-arm64-params
```

config/pipeline.yaml
```diff
@@ -384,6 +393,12 @@ scheduler:
   - job: kbuild-gcc-10-arm64
     <<: *build-k8s-all
 
+  - job: kbuild-gcc-10-arm64-dtbscheck
+    <<: *build-k8s-all
+    rules:
+      tree: next
+      branch: 'master'
```
a-wai, line 400: Those should go into the job definition, not the scheduler entry (this could be a future TODO though).

a-wai (changes requested):

config/pipeline.yaml
```diff
@@ -179,6 +179,15 @@ jobs:
   kbuild-gcc-10-arm64:
     <<: *kbuild-gcc-10-arm64-job
 
+  kbuild-gcc-10-arm64-dtbscheck:
+    <<: *kbuild-gcc-10-arm64-job
+    params:
```
a-wai, line 184:
```suggestion
    kind: test
    params:
```
`kbuild-gcc-10-arm64-job` has `kind: kbuild`, I think this one should be different

a-wai: yes, that's what I meant, and also why we keep using the kbuild template; but the node/results should clearly be identified as `test` for easier understanding.

nuclearcat:
Thanks all, merging

PR #455, opened 2024-03-04
src/regression_tracker: add config_full to queries and regression objects

hardboprobot:
Depends on https://github.com/kernelci/kernelci-core/pull/2432


a-wai (approved):

nuclearcat:
Thanks, merging!

PR #457, opened 2024-03-04
Add rules based on legacy config

a-wai:
This PR adds a few rules matching those currently present in the legacy configs. Most of those will have no effect for now, as we're either not monitoring the corresponding tree (e.g. `android`) or already ensure the proper defconfig/fragments are used through job definitions and platforms allocation.

The lab rules make sense nonetheless as we'll likely start monitoring the Android tree at some point.

I also think the rules for Mediatek Chromebooks make sense due to fairly recent upstream support of some of those.

I'd happily get rid of per-platform rules on "generic" devices (those found in `pipeline.yaml`) though and defconfig/fragments rules for Chromebooks (see `platforms-chromeos.yaml`).

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -128,6 +128,9 @@ runtimes:
       callback:
         token: kernelci-new-api-callback
         url: https://staging.kernelci.org:9100
+    rules:
+      tree:
+      - '!android'
 
   lava-collabora: &lava-collabora-staging
     lab_type: lava
@@ -138,6 +141,9 @@ runtimes:
       callback:
         token: kernelci-api-token-staging
         url: https://staging.kernelci.org:9100
+    rules:
+      tree:
+      - '!android'
 
   # ToDo: avoid creating a separate Runtime entry
   # https://github.com/kernelci/kernelci-core/issues/2088
@@ -149,10 +155,8 @@ runtimes:
         url: https://staging.kernelci.org:9100
 
   lava-collabora-staging:
-    lab_type: lava
+    <<: *lava-collabora-staging
     url: https://staging.lava.collabora.dev/
-    priority_min: 40
-    priority_max: 60
     notify:
       callback:
         token: kernelci-api-token-lava-staging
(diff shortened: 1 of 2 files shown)


nfraprado:
Hey @a-wai . Didn't know from the top of my head, but going through https://github.com/kernelci/kernelci-deploy/commit/d6aedbad9699e11cb1c9892751bef44d7331e811, I checked when each series was merged, and the first release that includes them all is v6.4 (v6.3 has everything but the GPU). So I suggest v6.4 for mt8192-spherion to have full support  (It also has all configs btw, as [they were added](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=900785741289) in 6.3). 6.1 is only usable if we apply those patches on top, but my understanding is that you're adding generic rules for all branches, so we don't want to rely on patches.

Similarly, for Tomato, based on the [patch here](https://github.com/kernelci/kernelci-deploy/tree/main/patches/linux/linux-6.6.y-arm64-chromeos), it was only merged in v6.7, so to be able to boot without patches we'd need that version. Although note that the defconfig for Tomato only landed in v6.8, so if we don't want to rely on fragments, that'd need to be the first version.

I'm not sure about jacuzzi.

nfraprado:
For jacuzzi [the defconfig changes landed in 6.2](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=83f3da586828), so we can use that to avoid the need for fragments.

nfraprado:
Oh and one more thing, MT8186-Steelix support is scheduled for v6.9, and v6.9-rc1 should land in ~3 weeks from now. No defconfig change should be needed. So depending on how long you expect this PR to take to be merged it might make sense to already set that as the rule. (otherwise you can keep it as is and update it in 3 weeks from now)

a-wai:
@nfraprado Thanks! For Chromebooks we currently only use CrOS defconfigs + our fragments, so no need to have the full defconfig upstream. Based on that, I'll use 6.4 for spherion, 6.7 for tomato, 6.9 for steelix (jobs won't run until we get 6.9-rc1, but I don't see this as an issue, we're still in testing phase)

For jacuzzi I'm going for 6.1+, assuming it has everything needed except the defconfig changes, please correct me if that's wrong.

nfraprado:
@a-wai A quick test run of the dt-kselftest on jacuzzi running 6.1 (looking for "not ok" in https://lava.collabora.dev/scheduler/job/12949055) shows that all the main peripherals are there, so yes, it should run ok.

a-wai:
Dropped the platform rules from pipeline.yaml as they're useless, ready to merge

PR #458, opened 2024-03-05
scheduler: import artifacts from parent node before running job

a-wai:
In most situations, jobs need artifacts generated by their parent. As those ar no longer automatically propagated, we need to import them when needed.


nuclearcat (approved):
LGTM, tested on staging

PR #459, opened 2024-03-05
Implement param formatting in scheduler

a-wai:
Counterpart to and depends on https://github.com/kernelci/kernelci-core/pull/2438


a-wai:
Set `staging-skip` as it conflicts with #457

a-wai:
Dropped the config changes, this should fix staging

PR #460, opened 2024-03-05
scheduler: apply rules before running job

a-wai:
Scheduler config entries can now contain rules to ensure jobs are only executed when certain conditions are met. This change ensures this check is properly conducted before actually attempting to run a job.

Depends on https://github.com/kernelci/kernelci-core/pull/2439


nuclearcat (approved):
LGTM

PR #461, opened 2024-03-06
src/regression_tracker: collect error code and msg from failed node

hardboprobot:
Companion to https://github.com/kernelci/kernelci-core/pull/2440 but not dependent on it.

### DIFF
--- a/src/regression_tracker.py
+++ b/src/regression_tracker.py
@@ -49,6 +49,25 @@ class RegressionTracker(Service):
                 logs = self._collect_logs(parent)
         return logs
 
+    def _collect_errors(self, node):
+        """Returns a dict containing the 'error_code' and 'error_msg'
+        data fields of <node>. If <node> doesn't have any info in them,
+        it searches upwards through parent nodes until it reaches a node
+        that has them.
+        """
+        if node['data'].get('error_code'):
+            return {
+                'error_code': node['data']['error_code'],
+                'error_msg': node['data']['error_msg']
+            }
+        elif node.get('parent'):
+            parent = self._api.node.get(node['parent'])
+            return self._collect_errors(parent)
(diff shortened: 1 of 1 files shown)


src/regression_tracker.py
```diff
@@ -31,6 +31,25 @@ def _stop(self, sub_id):
...
+        that has them.
+        """
+        if node['data'].get('error_code'):
+            return {
+                'error_node': node['data']['error_code'],
+                'error_msg': node['data']['error_msg']
+            }
+        elif node.get('parent'):
+            parent = self._api.node.get(node['parent'])
+            return self._collect_errors(parent)
```
JenySadadia, line 47: I am not sure if we should set error information of parent node.
Specifically for `kbuild` nodes, for example, the parent node of `https://staging.kernelci.org:9000/latest/node/65e6d3578c896c56c9f5e9ff` has completely different runtime than the child one.

hardboprobot: In that particular case, that node has `error_code` and `error_msg`, so that's what the regression node will get.
For context, I wrote this code based on what I'm seeing in the node results:
- test leaf nodes (test cases) don't contain all the info about a test, the parent node (the test _suite_ node) does. So if the failed test of a regression is a leaf node, we need to get certain info from the first parent node that has it, which most of the time should be the test suite node
- Other type of nodes that can cause regressions (kbuilds) are supposed to contain all the required info themselves.

It's likely that we'll have to change (and generalize) this logic if we ever change how results are stored and propagated in nodes.

JenySadadia: Okay. Let's keep it like this as of now.

nuclearcat:
Merging, as tested on staging and it is new subsystem

PR #463, opened 2024-03-06
config: chromeos: take advantage of params formatting

a-wai:
Lots of parameters for chromebooks are mostly identical, only differing by a few characters which are device/family-specific.

For jobs, this mostly affects the kbuild defconfigs. Regarding platforms, each entry must specify the full URL to the Tast tarball, even though the format is identical.

Taking advantage of the recent ability to format string params using the object's own attributes (or other params), this change aims at simplifying the current Chromebook configs.

Depends on #459 https://github.com/kernelci/kernelci-core/pull/2438

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -9,7 +9,8 @@ _anchors:
       compiler: gcc-10
       cross_compile: 'aarch64-linux-gnu-'
       cross_compile_compat: 'arm-linux-gnueabihf-'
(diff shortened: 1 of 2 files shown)


a-wai:
Setting `staging-skip` due to conflicts with #457, will rebase once the latter is merged

a-wai:
It has been sitting here for a while, no one objected, so merging to avoid conflicts with other PRs.

PR #469, opened 2024-03-12
Enable DT kselftest for MediaTek Chromebooks

nfraprado:
Enable the DT kselftest for MediaTek Chromebooks.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -350,6 +350,13 @@ jobs:
   baseline-x86-stoneyridge: *baseline-job
   baseline-x86-stoneyridge-staging: *baseline-job
 
+  kbuild-gcc-10-arm64-chromebook:
+    <<: *kbuild-gcc-10-arm64-chromeos-job
+    params:
+      <<: *kbuild-gcc-10-arm64-chromeos-params
+      cross_compile_compat:
+      defconfig: defconfig
+
   kbuild-gcc-10-arm64-chromeos-mediatek:
     <<: *kbuild-gcc-10-arm64-chromeos-job
     params:
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -241,6 +241,7 @@ jobs:
     template: kselftest.jinja2
     kind: test
     params:
+      nfsroot: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240221.0/{arch}'
       collections: dt
       job_timeout: 10
 
--- a/config/scheduler-chromeos.yaml
+++ b/config/scheduler-chromeos.yaml
@@ -104,6 +104,9 @@ scheduler:
     platforms:
     - dell-latitude-3445-7520c-skyrim
 
+  - job: kbuild-gcc-10-arm64-chromebook
+    <<: *build-k8s-all
+
   - job: kbuild-gcc-10-arm64-chromeos-mediatek
     <<: *build-k8s-all
 
--- a/config/scheduler-chromeos.yaml
+++ b/config/scheduler-chromeos.yaml
@@ -119,6 +119,14 @@ scheduler:
   - job: kbuild-gcc-10-x86-chromeos-stoneyridge
     <<: *build-k8s-all
 
+  - job: kselftest-dt
+    <<: *lava-job-collabora
+    event:
+      channel: node
+      name: kbuild-gcc-10-arm64-chromebook
+      result: pass
+    platforms: *mediatek-platforms
+
   - job: tast-basic-arm64-mediatek
     <<: *test-job-mediatek
 


config/pipeline.yaml
```diff
@@ -304,6 +305,16 @@ platforms:
     mach: amlogic
     dtb: dtbs/amlogic/meson-g12b-a311d-khadas-vim3.dtb
 
+  mt8192-asurada-spherion-r0:
```
a-wai, line 308: This will likely be a problem: this platform is already defined in `platforms-chromeos.yaml`, and all YAML files are merged into a single config object, meaning we could end up with an undefined behaviour here depending on which file is parsed first...

Technically this PR could be reduced to the added platform line in the scheduler entry, but OTOH I'd love to keep `pipeline.yaml` self-contained, in the sense that it shouldn't reference a platform defined in a separate file.

For now the best way to be to add an entry in `scheduler-chromeos.yaml` for executing the needed job on spherion (no need for an additional job entry, as we assume `pipeline.yaml` is always available, and therefore can reference jobs defined here in other files).

Also, the `-chromeos` naming of those config files may be confusing, it refers to "devices meant to be running ChromeOS but that can run any mainline Linux-based system", not "ChromeOS-specific tests and jobs".

a-wai (changes requested):

config/pipeline.yaml
nfraprado: Ah ok, I thought the chromeos files were an overlay for the chromeos instance, like in the old system. I'll just add the test entry in the chromeos file like you mentioned then.

nfraprado: @a-wai on second thought, we need the `nfsroot` to be set to the kselftest rootfs, but the base device definition in `arm64-chromebook-device:` also sets it for the chromebooks. Is there a guarantee in the order that `params` are processed? Can we set `nfsroot` to the kselftest rootfs in the job or the scheduler definition and have it always override the device one?

a-wai: Hmm, it's actually the [other way around](https://github.com/kernelci/kernelci-core/blob/main/kernelci/runtime/__init__.py#L152): platform params override job params... We could probably change that, but it needs a bit more thinking...

nfraprado: Sounds reasonable to me. The only issue would be if we ever want to override a param for a specific platform without needing to define it for every other platform, but I also can't think of an example right now, so maybe it won't ever be an issue.

Anyway, I've updated the PR now like you suggested, and took the opportunity to enable the test for all MTK Chromebooks instead of just Spherion. So as soon as the change in param overriding order is done this should be ready to go. Are you going to look into that or should I? @a-wai

a-wai: Done as https://github.com/kernelci/kernelci-core/pull/2465

nfraprado: Thank you!

nfraprado:
I forgot to consider the lack of the arm64-chromebook fragment in the main build job, so the tests never ran after I started using the chromeos platform definitions. Now I've added a new build with the fragment so the tests should start getting scheduled.

config/jobs-chromeos.yaml
```diff
@@ -350,6 +350,11 @@ jobs:
   baseline-x86-stoneyridge: *baseline-job
   baseline-x86-stoneyridge-staging: *baseline-job
 
+  kbuild-gcc-10-arm64-chromebook:
+    <<: *kbuild-gcc-10-arm64-job
+    fragments:
+      - arm64-chromebook
```
a-wai, line 356:
```suggestion
  kbuild-gcc-10-arm64-chromebook:
    <<: *kbuild-gcc-10-arm64-chromeos-job
    params:
      <<: *kbuild-gcc-10-arm64-chromeos-params
      cross_compile_compat:
      defconfig: defconfig
```

`kbuild-gcc-10-arm64-job` is only defined in `pipeline.yaml` (at least for now), and having it referenced in a different file causes parsing issues as cross-files anchors unfortunately aren't supported. jobs-chromeos.yaml does have the `kbuild-gcc-10-arm64-chromeos-job` (and corresponding `-params`) aliases though, so you can use those.

`fragments` already contain the proper value (unless you want/need to get rid of `CONFIG_MODULE_COMPRESS=n`), but `defconfig` should be set back to the default upstream defconfig, and I have a feeling (which might be wrong) `cross_compile_compat` should be reset to null in this case.

a-wai (changes requested):

nfraprado:
@a-wai shall we remove the staging-skip tag to try the PR again? :)

a-wai:
yep, should be fine now, I dropped the tag

a-wai (approved):
Looks good, let's see how it looks after a staging run

nuclearcat:
Works on staging: https://staging.kernelci.org:9000/viewer?search=parent%3D6602db1b75be4a7966ccb9f8

PR #471, opened 2024-03-13
pipeline.yaml: Fix dtbs_check rules to trigger test

nuclearcat:
Reason of not triggering job that rules was not properly set.


nfraprado:
@nuclearcat Thank you so much for fixing this! At least for now, testing only on the master branch of the next tree, like is set in this PR, already gives us all the results we need. So LGTM!

PR #472, opened 2024-03-13
src/regression_tracker: fix regression detection

hardboprobot:
When searching for previous test runs, don't assume all search parameters will be present, as some of them might be null.

~~**NOTE**: Pending to test in staging~~ Tested OK

### DIFF
--- a/src/regression_tracker.py
+++ b/src/regression_tracker.py
@@ -97,7 +97,7 @@ class RegressionTracker(Service):
     def _detect_regression(self, fail_node):
         """Detects if <fail_node> (a failed job) produces a regression,
         ie. if the previous job run with the same parameters passed"""
-        previous_nodes = self._api.node.find({
+        search_params = {
             'name': fail_node['name'],
             'group': fail_node['group'],
             'path': fail_node['path'],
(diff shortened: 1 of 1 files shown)


src/regression_tracker.py
```diff
@@ -107,14 +107,16 @@ def _detect_regression(self, fail_node):
...
-            'data.defconfig': fail_node['data']['defconfig'],
-            'data.config_full': fail_node['data']['config_full'],
-            'data.compiler': fail_node['data']['compiler'],
-            'data.platform': fail_node['data']['platform'],
             'created__lt': fail_node['created'],
-            'state': 'done'
-        })
+            'state': 'done',
+            # Parameters that may be null in some test nodes
+            'data.arch': fail_node['data'].get('arch'),
```
JenySadadia, line 113: API queries with `field=None` are not supported yet.
```
$ curl 'https://staging.kernelci.org:9000/latest/nodes?data.debug=None&limit=1'
{"items":[],"total":0,"limit":1,"offset":0}
```
```
$ curl 'https://staging.kernelci.org:9000/latest/nodes?data.debug=null&limit=1'
{"items":[{"id":"65a12eabe98651d0fe81e3ed","kind":"checkout","name":"checkout","path":["checkout"],"group":null,"parent":null,"state":"done","result":null,"artifacts":null,"data":{"kernel_revision":{"tree":"kernelci","url":"https://github.com/kernelci/linux.git","branch":"staging-mainline","commit":"3145cd9dd8e634f31e1e14cb4f02d23573354585"}},"debug":null,"created":"2024-01-12T12:20:59.469000","updated":"2024-01-12T13:21:55.301000","timeout":"2024-01-12T13:20:59.406000","holdoff":null,"owner":"admin","user_groups":[]}],"total":1093301,"limit":1,"offset":0}
```

hardboprobot: `'None' != None`. The query you ran with `curl` tries to match `data.debug` to the string `None`. In the code I wrote, the parameters are matched against `None` the _value_. But I'll set them to `null` explicitly so that the intention is more explicit.

JenySadadia: As per the default behavior of `FastAPI`, all the query params are sent as `str` to API.
API translates them later based on the field type. Hence, the API will receive `'None'` and not `None` from the above query.

> But I'll set them to null explicitly so that the intention is more explicit.

Bdw this would also need a translation method of query param values with `null` as if you send `'null'` directly in the query, it will be again sent as `str`.

JenySadadia: OK. I tried and API is working with string `'null'`.

JenySadadia (approved):

PR #484, opened 2024-03-19
src/test_report: sync to new Node path format

r-c-n:
Print the full path string for all nodes.

**DEPENDS ON** https://github.com/kernelci/kernelci-core/pull/2466


JenySadadia (approved):
LGTM.
Let's wait for staging results.

PR #495, opened 2024-03-22
Fix Tast tests on `trogdor-kingoftown` with kernels >= 6.4

a-wai:
This PR reworks the `tast` template so it uses the new `device_dtb` parameter instead of the old `platform_config.dtb`. It also honors `platform_config.params.flash_kernel.dtb` if present, falling back to `device_dtb` otherwise.

The config for `trogdor-kingoftown` is updated so it can use either the old DTB file name (up until 6.3.x) or the new one (renamed starting with 6.4-rc1), while still using the correct file name for the "flash kernel" (which is based on 6.1).

Depends on https://github.com/kernelci/kernelci-core/pull/2471

### DIFF
--- a/config/platforms-chromeos.yaml
+++ b/config/platforms-chromeos.yaml
@@ -150,12 +150,15 @@ platforms:
     <<: *arm64-chromebook-device
     base_name: trogdor
     mach: qcom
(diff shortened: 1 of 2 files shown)


nuclearcat (approved):
Tested on staging

PR #501, opened 2024-03-25
regression_tracker: sync to latest model changes

hardboprobot:
See https://github.com/kernelci/kernelci-core/pull/2474

To be applied after kernelci-core commit

models: rename regression.data.failed_kernel_revision to kernel_revision


nuclearcat:
do we still need this PR?

Issue #483, opened 2024-03-19
Enable ARCH=um - UM Linux compiling tests - for Android Common Kernels


helen-fornazier:
Reference: https://lore.kernel.org/kernelci/CAHRSSEwe1z7pajTmLNJRHbBv1Ghnz-siAyw4CU5kraN3_hLqpg@mail.gmail.com/T/#t

It was requested to add UML compiling tests `ARCH=um SUBARCH=x86_64` to kernelci for Android Common Kernels

Add them into the new system and check the best way to report errors from it

pawiecz:
Here's a list of steps to enable setting env variables for the builds in the new system:

1. Start by making them an argument to [`_build*`](https://github.com/kernelci/kernelci-core/blob/6572f1c403506d4ce2c1117167cf86054adb1db3/kernelci/kbuild.py#L526) functions in the [kernel build abstraction package](https://github.com/kernelci/kernelci-core/blob/main/kernelci/kbuild.py)
2. Once it ^ accepts parameters, adjust [kbuild job template](https://github.com/kernelci/kernelci-pipeline/blob/main/config/runtime/kbuild.jinja2) to pass these parameters (as `KBUILD_PARAMS`) to executed job
3. When ^ job template is adjusted, set env vars in your new kbuild job definition; see [example gcc-10-arm64 definition](https://github.com/kernelci/kernelci-pipeline/blob/e2bcc7dc6ae24df321aaf360805ee360b797b369/config/pipeline.yaml#L183) and its neighbors

musamaanjum:
@helen-fornazier All the Android builds and baseline tests on arm/arm64 have been enabled. Is there anything else remaining for this issue? The PR, https://github.com/kernelci/kernelci-pipeline/pull/528 seems outdated.

PR #474, opened 2024-03-15
Enable baylibre lab

JenySadadia:

PR #502, opened 2024-03-25
pipeline: Add android-mainline

helen-fornazier:
Add definitions of android-mainline from legacy system to the new system.
Add android-mainline tree
~~Add lab-kontron and lab-pengutronix definitions~~
~~Add respective platforms~~

NOTE: please review this, I didn't test and I wrote based on what I saw from current code and legacy code, so I'm very unsure about this. Please let me know if there is any better way to check/test it.

Reference: https://github.com/kernelci/kernelci-pipeline/issues/483

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -286,6 +286,9 @@ trees:
   mediatek:
     url: 'https://git.kernel.org/pub/scm/linux/kernel/git/mediatek/linux.git'
 
+  android:
+    url: 'https://android.googlesource.com/kernel/common'
+
 platforms:
 
   docker:
@@ -592,4 +595,9 @@ build_configs:
   mediatek_for_next:
     tree: mediatek
     branch: 'for-next'
+    variants: *build-variants
+
+  android_mainline:
+    tree: android
+    branch: 'android-mainline'
     variants: *build-variants
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -198,14 +198,26 @@ jobs:
       <<: *kbuild-gcc-10-arm64-params
       dtbs_check: true
 
-  kbuild-gcc-10-armel:
+  kbuild-gcc-10-armel: &kbuild-gcc-10-armel-job
     <<: *kbuild-job
     image: kernelci/staging-gcc-10:arm64-kselftest-kernelci
-    params:
+    params: &kbuild-gcc-10-armel-params
       arch: arm
       compiler: gcc-10
       cross_compile: 'arm-linux-gnueabihf-'
       defconfig: multi_v7_defconfig
+
+  kbuild-gcc-10-armel-android-mainline: &kbuild-gcc-10-armel-android-mainline-job
+    <<: *kbuild-gcc-10-armel-job
+    rules:
+      tree:
+      - 'android'
+
+  kbuild-gcc-10-armel-android-mainline-imx: &kbuild-gcc-10-armel-android-mainline-imx-job
+    <<: *kbuild-gcc-10-armel-android-mainline-job
+    params:
+      <<: *kbuild-gcc-10-armel-params
+      defconfig: imx_v6_v7_defconfig
 
   kbuild-gcc-10-i386:
     <<: *kbuild-job
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -467,6 +467,12 @@ scheduler:
   - job: kbuild-gcc-10-armel
     <<: *build-k8s-all
 
+  - job: kbuild-gcc-10-armel-android-mainline
+    <<: *build-k8s-all
+
+  - job: kbuild-gcc-10-armel-android-mainline-imx
+    <<: *build-k8s-all
+
   - job: kbuild-gcc-10-i386
     <<: *build-k8s-all
 


config/pipeline.yaml
```diff
@@ -101,6 +101,22 @@ storage:
 
 runtimes:
 
+  lab-kontron:
```
a-wai, line 104: The current convention is to have LAVA labs definitions prefixed with `lava-`

config/pipeline.yaml
```diff
@@ -101,6 +101,22 @@ storage:
...
+    lab_type: legacy.lava_rest
+    url: 'https://lavalab.kontron.com/'
+    filters:
+      - passlist:
+          plan:
+            - baseline
```
a-wai, line 110: Those aren't actually used in the current system, and not needed: we have `rules:` for filtering on kernel trees/branches/versions and other small things, but scheduler entries directly associate jobs with runtimes, so there's no filtering needed here.

config/pipeline.yaml
```diff
@@ -101,6 +101,22 @@ storage:
...
+
+  lab-pengutronix:
+    lab_type: legacy.lava_rest
```
a-wai, line 113: No sure the new system implements this lab_type, @nuclearcat ?

config/pipeline.yaml
```diff
@@ -342,9 +361,91 @@ platforms:
 
   shell:
 
+  kontron-kswitch-d10-mmt-6g-2gs:
+    mach: at91
+    class: arm-dtb
+    dtb: 'lan966x-kontron-kswitch-d10-mmt-6g-2gs.dtb'
+    boot_method: uboot
```
a-wai, line 368:
```suggestion
    <<: *armel-device
    mach: microchip
    dtb: dtbs/lan966x-kontron-kswitch-d10-mmt-6g-2gs.dtb
```

`armel-device` defines the necessary things for those devices using u-boot. We haven't yet formalized what `mach` should define in the new system, but using the vendor (as in, device-tree subfolder name) makes sense to me. Other opinions are welcome.

config/pipeline.yaml
```diff
@@ -342,9 +361,91 @@ platforms:
...
+  imx6dl-riotboard:
+    mach: imx
+    class: arm-dtb
+    boot_method: barebox
+    filters:
+      - blocklist:
+          defconfig:
+            - 'imx_v4_v5_defconfig'
+            - 'multi_v5_defconfig'
+            - 'allmodconfig'
```
a-wai, line 385:
```suggestion
    rules:
      defconfig:
        - '!imx_v4_v5_defconfig'
        - '!multi_v5_defconfig'
        - '!allmodconfig'
```

That's the same filters, only translated to the new `rules:` parameters.

config/pipeline.yaml
```diff
@@ -342,9 +361,91 @@ platforms:
...
+
+  imx6qp-wandboard-revd1:
+    mach: imx
+    class: arm-dtb
+    boot_method: barebox
+    flags: ['fastboot']
```
a-wai, line 391: This field isn't handled in the new system.

config/pipeline.yaml
```diff
@@ -342,9 +361,91 @@ platforms:
...
+    mach: mvebu
+    class: arm-dtb
+    boot_method: barebox
+    filters:
+      - blocklist:
+          defconfig:
+            - 'allmodconfig'
+            - 'allnoconfig'
+            - 'tinyconfig'
+            - '-chromebook'
```
a-wai, line 409: This one will be tricky: I assume it means "exclude any defconfig containing `-chromebook`", for which we don't have an equivalent yet. Maybe having the following would do:
```
            fragments:
              - '!arm64-chromebook'
```

Not sure it would be usefull though, as the name implies we only have a fragment for arm64 chromebooks, but no other `arm*-chromebook` exists.

config/pipeline.yaml
```diff
@@ -342,9 +361,91 @@ platforms:
...
 
 scheduler:
 
+  - job: baseline-armel-kontron-android_mainline
+    event:
+      channel: node
+      name: android-mainline
```
a-wai, line 416: You must actually define a `kbuild` job (e.g. `kbuild-gcc-10-armel-android-mainline`), add a corresponding scheduler entry, and use that job's name here.

config/pipeline.yaml
```diff
@@ -552,3 +653,21 @@ build_configs:
...
     variants: *build-variants
+
+  android_mainline:
+    tree: android
+    branch: 'android-mainline'
+    build_environment: gcc-10
+    arch: armel
+    defconfig: multi_v7_defconfig
+    filters:
+      - regex: { defconfig: 'multi_v7_defconfig' }
```
a-wai, line 664:
```suggestion
    variants: *build-variants
```

No need to define additional parameters here, this section is only used to define tree/branch combinations we want to monitor. The important bit (defconfig) should be defined in the `kbuild` job.

config/pipeline.yaml
```diff
@@ -552,3 +653,21 @@ build_configs:
...
+
+  android_mainline_imx:
+    tree: android
+    branch: 'android-mainline'
+    build_environment: gcc-10
+    arch: armel
+    defconfig: imx_v6_v7_defconfig
+    filters:
+      - regex: { defconfig: 'imx_v6_v7_defconfig' }
```
a-wai, line 673: This can go away, as explained above this will be covered by the `android_mainline` config. A separate `kbuild` job should be created using this defconfig.

a-wai (changes requested):
This PR definitely demonstrates how documenting things becomes more and more urgent, as you're basically hitting the "only a few know" barrier, which is definitely on us (full-time contributors) to blame.

There are also stuff in there we never really discussed (like the `mach` parameter) so expect maybe a few contradictory opinions on that.

Finally, we have an (unwritten, again) rule of listing devices/runtimes/jobs/etc by alphabetical order (even if it's not fully enforced), can you please try to follow that principle in the future?

config/pipeline.yaml
pawiecz: It doesn't - that's a `lab_type` from the legacy system. New one relies on `runtime` names which would be just `lava` in this case.

config/pipeline.yaml
```diff
@@ -101,6 +101,22 @@ storage:
...
+            - baseline
+
+  lab-pengutronix:
+    lab_type: legacy.lava_rest
+    url: 'https://hekla.openlab.pengutronix.de/'
+    filters:
+      - passlist:
+          plan:
+            - baseline
+
```
pawiecz, line 119: I realize tests for Android trees were run in those two labs only. Let's focus on getting just Android builds in the scope of this PR and move lab enablement to a separate one - it will require [very similar change](https://github.com/kernelci/kernelci-pipeline/pull/474) and some admin work (token management) on both (KCI, lab) sides.

config/pipeline.yaml
pawiecz, line 409: Let's move `platforms` definitions out of the scope of this PR and bring them back when new labs are enabled in the config.

pawiecz (changes requested):
Let's reduce the scope of this PR to include only: tree definition, `build_configs` adjustments and setting up `kbuild_*` jobs (moving `baseline_*` ones to a separate PR - together with labs and platforms definitions).

config/pipeline.yaml
helen-fornazier: right, but if I remove it here, where should I add the `rules:`? on the `baseline:` definition?

```yaml
  baseline: &baseline-job
    template: baseline.jinja2
    kind: test
    rules:
      tree:
      - 'android'
```

How can I define which tests should be run on the tree I'm adding?

config/pipeline.yaml
helen-fornazier: what should I add instead? Or just remove it and it should work?

config/pipeline.yaml
helen-fornazier: right, so what is the meaning of the `build_configs:` if they are not used? (At least my grep powers failed me)

config/pipeline.yaml
helen-fornazier: but to get Android builds, don't we need those labs?

config/pipeline.yaml
helen-fornazier: same question as https://github.com/kernelci/kernelci-pipeline/pull/502#discussion_r1546339697

helen-fornazier:
@a-wai @pawiecz PR updated (except for the PR separation, please check my questions), please review it when you have some time (and thanks a lot for the review)

config/pipeline.yaml
a-wai: Most of those can be "inherited" from the kbuild job: in the `scheduler` entries, you'll associate test jobs with a parent node/job, so test jobs will only run on trees the parent node can use.

In this case, you'll add those rules to the `kbuild-gcc-10-armel-android*` jobs you're defining; that way, those kbuild nodes will only run on the android tree, and baseline jobs depending on those will only run on this tree as well.

config/pipeline.yaml
a-wai: I'm afraid we don't have any template for barebox yet, so the final answer will have to wait for that to be implemented.

I think it should go under a `params:` section, where we usually stuff most non-standard things we need.

config/pipeline.yaml
a-wai: IIUC it *is* used but only by a specific service, @nuclearcat should have the details about that.

config/pipeline.yaml
```diff
@@ -342,9 +369,86 @@ platforms:
...
+    boot_method: barebox
+    rules:
+      defconfig:
+        - '!imx_v4_v5_defconfig'
+        - '!multi_v5_defconfig'
+        - '!allmodconfig'
+
+  dove-cubox:
+    mach: mvebu
+    class: arm-dtb
```
a-wai, line 404:
```suggestion
    <<: *armel-device
    mach: marvell
    dtb: dtbs/filename.dtb
```

`class` isn't handled in the new system, and `mach` should match the vendor's DTS folder name.

Maybe creating an `armel-barebox-device` anchor would make sense here.

config/pipeline.yaml
```diff
@@ -342,9 +369,86 @@ platforms:
...
+    dtb: dtbs/lan966x-kontron-kswitch-d10-mmt-6g-2gs.dtb
+
+  kontron-kswitch-d10-mmt-8g:
+    <<: *armel-device
+    mach: microchip
+    dtb: dtbs/lan966x-kontron-kswitch-d10-mmt-8g.dtb
+
+  imx6dl-riotboard:
+    mach: imx
+    class: arm-dtb
```
a-wai, line 384:
```suggestion
    <<: *armel-device
    mach: nxp
    dtb: dtbs/filename.dtb
```

config/pipeline.yaml
```diff
@@ -49,6 +49,22 @@ _anchors:
...
+      compiler: gcc-10
+      defconfig: multi_v7_defconfig
+
+  kbuild-gcc-10-armel-android-mainline-imx: &kbuild-gcc-10-armel-android-mainline-imx-job
+    <<: *kbuild-job
+    image: kernelci/staging-gcc-10:armel-kselftest-kernelci
+    params: &kbuild-gcc-10-armel-imx-params
+      arch: armel
+      compiler: gcc-10
+      defconfig: imx_v6_v7_defconfig
```
a-wai, line 66:
```suggestion
    <<: *kbuild-gcc-10-armel-android-mainline-job
    params:
      <<: *kbuild-gcc-10-armel-params
      defconfig: imx_v6_v7_defconfig
```

Note: this will also inherit the rules when you add those, so no need to write them down twice ;)

a-wai (changes requested):

config/pipeline.yaml
helen-fornazier: well, I did as you mentioned for now

config/pipeline.yaml
helen-fornazier: I guess it is just a way to say to kernelci to monitor this tree (I guess)

config/pipeline.yaml
helen-fornazier: right, I through it would append and not override, awesome, thanks

config/pipeline.yaml
helen-fornazier: I added this change for now, please let me know if it is not correct

config/pipeline.yaml
helen-fornazier: @pawiecz @a-wai please check https://github.com/kernelci/kernelci-pipeline/pull/528

helen-fornazier: @pawiecz @a-wai  I also updated this PR to just add the kbuild jobs, please check

config/pipeline.yaml
```diff
@@ -57,6 +57,23 @@ _anchors:
       compiler: gcc-10
       defconfig: x86_64_defconfig
 
+  kbuild-gcc-10-armel-android-mainline: &kbuild-gcc-10-armel-android-mainline-job
+    <<: *kbuild-job
+    image: kernelci/staging-gcc-10:armel-kselftest-kernelci
+    params: &kbuild-gcc-10-armel-params
+      arch: armel
```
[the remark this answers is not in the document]
helen-fornazier: ` kbuild-gcc-10-armel` is using a arm64 image, is this correct? See https://github.com/kernelci/kernelci-pipeline/blob/main/config/pipeline.yaml#L203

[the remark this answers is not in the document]
helen-fornazier: in any case I just updated this PR with this change, please check

config/pipeline.yaml
```diff
@@ -57,6 +57,18 @@ _anchors:
...
+    rules:
+      tree:
+      - 'android'
+
+  kbuild-gcc-10-armel-android-mainline-imx: &kbuild-gcc-10-armel-android-mainline-imx-job
+    <<: *kbuild-gcc-10-armel-android-mainline-job
+    params:
+      <<: *kbuild-gcc-10-armel-params
+      defconfig: imx_v6_v7_defconfig
+
```
a-wai, line 71: Please move those to the `jobs` section (`_anchors` is only used for, well, anchors, and not parsed as actual pipeline items), right below `kbuild-gcc-10-armel` would be good.

a-wai (changes requested):
Only a small change left which I didn't spot before, otherwise should be good to go :)

After it's done feel free to drop the `staging-skip` label and undraft this PR

config/pipeline.yaml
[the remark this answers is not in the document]
a-wai: Yes, this is not an issue, the arm64 image also contains cross-compilers for armv7

config/pipeline.yaml
helen-fornazier: done, please check

helen-fornazier:
> Only a small change left which I didn't spot before, otherwise should be good to go :)
> 
> After it's done feel free to drop the `staging-skip` label and undraft this PR

looks like I don't have permissions to undraft and drop labels, I can only edit the commit title

a-wai (approved):
LGTM, thanks!

a-wai:
Looks like I spoke too soon ;) Can you please add entries to the `scheduler:` section, so the builds are actually executed? (those should be quite straightforward)

helen-fornazier:
> Looks like I spoke too soon ;) Can you please add entries to the `scheduler:` section, so the builds are actually executed? (those should be quite straightforward)

should I just add the following? Or is there anything else?

```yaml
  - job: kbuild-gcc-10-armel-android-mainline
    <<: *build-k8s-all
    
  - job: kbuild-gcc-10-armel-android-mainline-imx
    <<: *build-k8s-all
    ```

**Comment by helen-fornazier** [comment]
@a-wai I added a new commit to this PR adding entries to `scheduler:`, please check

File: config/pipeline.yaml [hunk 17]
```diff
@@ -467,6 +467,22 @@ scheduler:
   - job: kbuild-gcc-10-armel
     <<: *build-k8s-all
 
+  - job: kbuild-gcc-10-armel-android-mainline
+    <<: *build-k8s-all
+    rules:
+      tree:
+      - 'android'
+      branch:
+        - 'android-mainline'
```
a-wai, line 476: Those aren't needed (already enforced by the job config), same for the ones below

helen-fornazier: PR updated, please check

a-wai:
Indeed, I'm happily hitting that "Merge" button then :)

PR #503, opened 2024-03-26
src/regression_tracker: rework regression detection

hardboprobot:
**DEPENDS ON**: https://github.com/kernelci/kernelci-core/pull/2476

Take into account "active" and "inactive" regressions when creating them and when processing new passed or failed nodes.

When a node passes, it checks if it "inactivates" an existing "active" regression. When a node fails, it checks if it needs to create a new regression or update an existing "active" one.

### DIFF
--- a/src/regression_tracker.py
+++ b/src/regression_tracker.py
@@ -90,65 +90,126 @@ class RegressionTracker(Service):
             'error_msg': error['error_msg'],
         }
         regression['artifacts'] = self._collect_logs(failed_node)
-        resp = self._api_helper.submit_regression(regression)
-        reg = json.loads(resp.text)
-        self.log.info(f"Regression submitted: {reg['id']}")
+        return regression
 
-    def _detect_regression(self, fail_node):
-        """Detects if <fail_node> (a failed job) produces a regression,
-        ie. if the previous job run with the same parameters passed"""
-        search_params = {
-            'name': fail_node['name'],
-            'group': fail_node['group'],
-            'path': fail_node['path'],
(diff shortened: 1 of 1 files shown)


src/regression_tracker.py
```diff
@@ -72,6 +72,7 @@ def _create_regression(self, failed_node, last_pass_node):
         """Method to create a regression"""
         regression = {}
         regression['kind'] = 'regression'
+        regression['result'] = 'fail'
```
JenySadadia, line 75: As we are always creating regression nodes with `result=fail`, how about setting the field by default to `fail`?

src/regression_tracker.py
```diff
@@ -88,67 +89,135 @@ def _create_regression(self, failed_node, last_pass_node):
             'failed_kernel_version': failed_node['data'].get('kernel_revision'),   # noqa
             'error_code': error['error_code'],
             'error_msg': error['error_msg'],
+            'node_sequence': [],
```
JenySadadia, line 92: The same comment as above.

src/regression_tracker.py
hardboprobot: Absolutely right, fixed in https://github.com/kernelci/kernelci-core/pull/2476, I'll remove the line here.

src/regression_tracker.py
hardboprobot: I agree, fixed in https://github.com/kernelci/kernelci-core/pull/2476, I'll remove this line

nuclearcat:
Tested on staging

PR #514, opened 2024-04-01
result_summary: speed up the script with ThreadPoolExecutor to fetch logs

helen-fornazier:
Fetching logs is the bottleneck of the script.
Fetch logs from each job in parallel with ThreadPoolExecutor.


src/result_summary.py
```diff
@@ -185,6 +186,17 @@ def _get_log(self, url, snippet_lines=0):
             return '\n'.join(lines[snippet_lines:])
         return text
 
+    def _get_logs(self, node):
```
hardboprobot, line 189: Can you add a brief docstring to this function describing what it does and/or how it's meant to be used?

hardboprobot (changes requested):
Fantastic! Thanks for the contribution. Just a small request about documentation (`_get_logs`).

Additionally, can you update `config/result_summary_templates/CHANGELOG`, even if this is a performance improvement that doesn't affect the user interface?

src/result_summary.py
helen-fornazier: Done

helen-fornazier:
> Fantastic! Thanks for the contribution. Just a small request about documentation (`_get_logs`).
> 
> Additionally, can you update `config/result_summary_templates/CHANGELOG`, even if this is a performance improvement that doesn't affect the user interface?

Done, please check

hardboprobot (approved):

PR #517, opened 2024-04-03
send_kcidb: submit build and test nodes

JenySadadia:
Related to https://github.com/kernelci/kcidb/issues/338

Fetch build and test nodes for the current checkout node. Create KCIDB schema compatible nodes from it and
send them along with checkout node.

### DIFF
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -44,6 +44,7 @@ class KCIDBBridge(Service):
             self._api_helper.unsubscribe_filters(context['sub_id'])
 
     def _send_revision(self, client, revision):
+        self.log.debug(f"DEBUG: sending revision: {revision}")
         if kcidb.io.SCHEMA.is_valid(revision):
             return client.submit(revision)
         self.log.error("Aborting, invalid data")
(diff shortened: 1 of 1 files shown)


hardboprobot (commented):
From my uninformed pov, I wonder if this approach of listening only to checkout nodes and then walking down the node tree ends up being more complicated than simply listening to all kind of nodes and then filter/process/translate them according to a set of rules. I feel like that way could lend itself to simpler code down the road, where the processing/translation logic is clearly separated in a rule-based data processing stage, which could make it easier to maintain in the future if new changes and exceptions are required. Just an idea. Is there a reason why this top-down approach was used instead?

Apart from that, if it works then it's fine by me.

JenySadadia:
Thanks for the review.
Yes, handling one completed node at a time would make the logic a bit simpler.

I only see one issue with that approach.
Let's assume we submit nodes one by one. At first, the service submitted a build and a test node successfully. After some time the service crashed for whatever reason. Meanwhile if `checkout` node transits to `done` state, that event will be missed by the service and the checkout data will never reach KCIDB.
Hence, KCIDB will never be able to link those previously submitted nodes to a checkout. I am not sure how KCIDB deals with such data but those previously submitted nodes will always miss the link with the revision/checkout they were generated for.

JenySadadia:
> On the other hand, wouldn't it be possible to listen for checkout nodes when they're created and available rather than when they're `done`? Sorry, I don't remember the full details about the states of `checkout` nodes. What I mean is that, if I understood correctly, this waits until the `checkout` node is `done`, ie. when all its children have finished (again, I may be wrong here), while the node may have been created some time ago. Why not send it to KCIDB when it's created rather than wait until the `done` state?

Yes, that's doable. We can listen to `available` checkout nodes and submit it to KCIDB and then update its `result` when it goes to `done` state.

If we decide to go with this approach, another thing that came to my mind is that we will need to traverse to the parent `kbuild` node for every child test node e.g. `dmesg` or a child of it for `build_id` field.
Also, we would be submitting `dummy_build` node multiple times for test nodes hanging directly from `checkout` as KCIDB doesn't provide any method to check if the node already exists there.

Apart from that, yes, we can go with this approach as well.

JenySadadia:
Tested on staging.
Data submitted to KCIDB: https://kcidb.kernelci.org/d/revision/revision?orgId=1&var-datasource=playground&var-git_commit_hash=a019c154299cad412f6b958a0f5d2a5d89391400&var-patchset_hash=&var-origin=kernelci_api&var-build_architecture=All&var-build_config_name=All&var-test_path=

pawiecz:
@JenySadadia Should this PR be kept open with #550 merged or has it been superseded and is no longer relevant?

JenySadadia:
Yes, this can be closed now. Thanks for taking a look.

PR #520, opened 2024-04-03
config: chromeos: remove/disable meaningless Tast tests

a-wai:
Several Tast tests can only fail in the context of KernelCI:
* `video.PlatformDecoding.v4l2_state*_vp9_0_svc` do not actually exist, causing the whole test job to fail
* `platform.DLCService*` and `platform.Memd` rely on features only present in the downstream Chrom{e,ium}OS kernel (see b/247467814 and b/244479619 for those having access to Google's issue tracker)
* `kernel.ConfigVerify.chromeos` relies on downstream-only config options such as `CONFIG_SECURITY_CHROMIUMOS` and other similar ones, and therefore can only fail when testing upstream kernels

HEVC tests are also expected to either fail or be skipped as ChromeOS [doesn't handle hardware decoding](https://chromium.googlesource.com/chromiumos/overlays/board-overlays/+/accf64c6d3df95945c42ab6740dedddea556f218) for this codec. This might change in the future, though, so we only drop the scheduler entries for those tests in order to easily reinstate them in the future.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -77,7 +77,6 @@ _anchors:
     params:
       tests:
         - video.PlatformDecoding.v4l2_stateful_vp9_0_level5_*
-        - video.PlatformDecoding.v4l2_stateful_vp9_0_svc
 
   tast-decoder-v4l2-sl-av1: &tast-decoder-v4l2-sl-av1-job
     <<: *tast-job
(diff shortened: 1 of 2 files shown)


laura-nao:
Ack on removing `video.PlatformDecoding.v4l2_stateful_vp9_0_svc` as it's not available on R120 and on disabling the HEVC Tast test on `trogdor` and `cherry` for now. Thanks!

nuclearcat:
Tested on staging, merging

PR #521, opened 2024-04-04
config: chromeos: remove/disable more failing Tast tests

a-wai:
As a follow-up to #520 this PR ensures more always-failing tests are disabled. This is a separate PR as this part requires https://github.com/kernelci/kernelci-core/pull/2490 to actually give the expected results.

It contains 2 parts/commits:
* exclude tests failing on all platforms and all kernel versions
* exclude *some* tests when testing kernels < 6.7 on qcom/trogdor devices (those tests pass with 6.7+)

The 2nd part is a bit more involved as it required duplicating test jobs and setting the proper rules on each variant so only the appropriate one is executed depending on the version of the kernel being tested.

Depends on https://github.com/kernelci/kernelci-core/pull/2490 (without it `excluded_tests` will only be ignored, but otherwise that should be "mostly harmless").

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -44,6 +44,10 @@ _anchors:
       tests:
         - video.ChromeStackDecoder.*
         - video.ChromeStackDecoderVerification.*
+      excluded_tests:
+        # Those always fail on all platforms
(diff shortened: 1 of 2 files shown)


nuclearcat:
Tested on staging, merging

PR #528, opened 2024-04-10
Draft: Enable android tree and add kontron and pengutronix runtime

helen-fornazier:
reference: https://github.com/kernelci/kernelci-pipeline/pull/502#discussion_r1540684977 https://github.com/kernelci/kernelci-pipeline/issues/483

Add runtime for kontron and pengutronix labs to be used by android trees

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -286,6 +286,9 @@ trees:
   mediatek:
     url: 'https://git.kernel.org/pub/scm/linux/kernel/git/mediatek/linux.git'
 
+  android:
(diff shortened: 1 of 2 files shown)


nuclearcat:
Do we have formal OK from both labs that they want to be enabled on staging instance of new KernelCI?
As they will need to provide new tokens for access and etc.

a-wai:
Adding `staging-skip` as this affects the docker-compose file and we need more data (such as a token) for those new labs

padovan:
Imcomplete work, closing it.

PR #530, opened 2024-04-11
tests: Add basic yaml validation

nuclearcat:
Add yaml load to figure out earlier issues with yaml


nuclearcat:
Merging, as we need to use it ASAP

PR #531, opened 2024-04-11
config: exclude android tree from kbuild jobs

a-wai:
Only Android-specific kbuild jobs should run for this tree, let's not overload our system with unneeded builds.

Take this opportunity to limit mediatek kbuilds to 6.1+ as that's the earliest version that has upstream support for at least one of our devices.


a-wai:
@helen-fornazier Feel free to correct me if I'm wrong and we shouldn't have such broad filtering

a-wai:
Tested on staging: https://staging.kernelci.org:9000/viewer?search=parent%3D661d23a7a29b490c6784d2c4

PR #532, opened 2024-04-12
 Enable baseline-nfs test on all the Chromebooks

laura-nao:
Enable the baseline-nfs test on all the supported Chromebooks, with both the default and the chromeos kernel configurations.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -357,6 +357,22 @@ jobs:
     template: baseline.jinja2
     kind: test
 
+  baseline-nfs-arm64-mediatek: &baseline-nfs-arm64-job
+    template: baseline.jinja2
+    kind: test
+    params: &baseline-nfs-arm64-job-params
+      boot_commands: nfs
+      nfsroot: http://storage.kernelci.org/images/rootfs/debian/bookworm/20240313.0/arm64
+
+  baseline-nfs-arm64-qualcomm: *baseline-nfs-arm64-job
+
+  baseline-nfs-x86-amd: &baseline-nfs-x86-job
+    <<: *baseline-nfs-arm64-job
+    params: &baseline-nfs-x86-job-params
+      <<: *baseline-nfs-arm64-job-params
+      nfsroot: http://storage.kernelci.org/images/rootfs/debian/bookworm/20240313.0/amd64
+
+  baseline-nfs-x86-intel: *baseline-nfs-x86-job
   baseline-arm64-qualcomm: *baseline-job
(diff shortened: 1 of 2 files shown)


config/runtime/baseline-nfs.jinja2
```diff
@@ -0,0 +1,8 @@
+{% set test_method = 'baseline' %}
+{% set boot_commands = 'nfs' %}
```
a-wai, line 2: IIUC that line is the only difference with `baseline.jinja2`. It might be worth setting it as a job param and reuse the existing template.

Also, please note that will work only for ChromeBooks (at least for now) as only the  `depthcharge` boot template handles NFS boot with an arbitrary rootfs.

config/jobs-chromeos.yaml
```diff
@@ -357,6 +357,16 @@ jobs:
     template: baseline.jinja2
     kind: test
 
+  baseline-nfs-arm64-mediatek: &baseline-nfs-job
+    template: baseline-nfs.jinja2
+    kind: test
+    params: &baseline-nfs-job-params
+      nfsroot: http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240313.0/{arch}
```
a-wai, line 364: This will likely fail for x86, as `arch` is `x86_64` in the platform configs, whereas the corresponding subfolder here is `amd64`...

This is a mess, I know, see for another example of where it messes up: https://github.com/kernelci/kernelci-core/issues/2500

I'm considering adding something like `deb_arch`, `kernel_arch` and/or `kci_arch` to the platform configs (which would be automatically determined from `arch`, so we don't have to add more noise to the configs), that might solve a few things, but for now I'd suggest hardcoding the arch here.

laura-nao: I missed that, thanks for the heads up! I hardcoded the arch for now.

a-wai (changes requested):
Job ran as expected, but it looks like this rootfs is missing the `dmesg.sh` script: https://lava.collabora.dev/scheduler/job/13356994

laura-nao:
> Job ran as expected, but it looks like this rootfs is missing the `dmesg.sh` script: https://lava.collabora.dev/scheduler/job/13356994

@a-wai I accidentally picked the `-kselftest` variant of the nfsrootfs, sorry about that. Should be fixed now.

a-wai (approved):
Looks like it's working on both [arm64](https://lava.collabora.dev/scheduler/job/13375448) and [x86](https://lava.collabora.dev/scheduler/job/13375241) chromebooks :)

PR #535, opened 2024-04-12
regression_tracker: bugfix, failed test with no prior runs

hardboprobot:
Handle the case of a failed test run when it's the first occurence of that test case. Consider it "not a regression" for now, since we're defining a regression as a "breaking point" between a success and a failure.

I suggest we leave this running on staging for a while to see if the issue happens again, although it may be hard to catch. ~I'll try to test it locally next Monday.~ -> Tested locally already: OK


JenySadadia (approved):
LGTM.

PR #536, opened 2024-04-12
config/pipeline.yaml: Only build arm and arm64 for mediatek tree

nfraprado:
Commits that go through the mediatek tree only target arm and arm64 platforms. Build only for those architectures to avoid wasting resources.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -34,6 +34,7 @@ _anchors:
     rules:
       tree:
       - '!android'
+      - '!mediatek'
 
   kbuild-clang-17-x86: &kbuild-clang-17-x86-job
     <<: *kbuild-job
@@ -208,6 +209,9 @@ jobs:
 
   kbuild-gcc-10-arm64:
     <<: *kbuild-gcc-10-arm64-job
+    rules:
+      tree:
+        - '!android'
 
   kbuild-gcc-10-arm64-dtbscheck:
     <<: *kbuild-gcc-10-arm64-job
@@ -215,6 +219,9 @@ jobs:
     params:
       <<: *kbuild-gcc-10-arm64-params
       dtbs_check: true
+    rules:
+      tree:
+        - '!android'
 
   kbuild-gcc-10-arm: &kbuild-gcc-10-arm-job
     <<: *kbuild-job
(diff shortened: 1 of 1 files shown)


a-wai:
With #531 now merged, can you please add `'!android'` to all entries? The new rules override the default ones, but those jobs still shouldn't run on the android tree.

nfraprado:
@a-wai since the build jobs that should run on mediatek were the minority I opted to filter out the mediatek tree by default and override the filter only in the jobs that should run on that tree, like you did for android. It would be nice to be able to remove an entry from the list (because the way it is it'll be easy to forget to update those overrides when the base filter is updated...), but I don't think that's possible in YAML.

config/pipeline.yaml
```diff
@@ -224,6 +231,9 @@ jobs:
       compiler: gcc-10
       cross_compile: 'arm-linux-gnueabihf-'
       defconfig: multi_v7_defconfig
+    rules:
+      tree:
+        - '!android'
```
a-wai, line 236: Just double checking: do you really intend to build arm32 kernels from the mediatek tree?

nfraprado: Yep, there are arm32 mediatek boards, and they still get patches landed through that tree.

a-wai: OK, thanks for confirming!

a-wai (approved):
LGTM, let's see how it goes on staging (need to wait for a new commit to this tree, though)

a-wai:
@nfraprado Can you please rebase?

nfraprado:
@a-wai rebased.

JenySadadia:
I see this is already approved, I added `staging-skip` to this as it's conflicting with https://github.com/kernelci/kernelci-pipeline/pull/550 on staging.
Let me know if you still need to test it.

nuclearcat:
Do we still need this PR?

nfraprado:
I just looked at the results on staging and I don't even see the mediatek tree being checked out. This tree isn't a priority as far as I'm concerned, by the way, so we can just leave it as is. The point of the PR was just to avoid wasting resources with unnecessary builds, but if we're not even building on this tree currrently, then we can just close this.

nuclearcat:
Thanks!

PR #542, opened 2024-04-15
src/timeout: set `checkout` result

JenySadadia:
For `TIMEOUT` mode, set `checkout` node result to `fail` if its state is `running` as it means code checkout is still going on and node timed-out. Set it to `pass` if its state is any other than `running`.
Set `checkout` node result to `pass` if mode is `DONE` as it means once `checkout` has been in `available` or `closing` state and it could successfully complete source code checkout.


nuclearcat:
We will need to document this behavior eventually for all node states transitions

PR #544, opened 2024-04-16
Improve architecture naming and format strings

a-wai:
This PR addresses 2 issues:
* the naming of 32-bits ARM architecture (`armel`) can be confusing to some and has different meanings for different systems; rename to `arm` so it makes more sense
* replace hardcoded system-dependent architecture strings with the corresponding properties (`crosarch`, `debarch`, `karch`)

Depends on https://github.com/kernelci/kernelci-core/pull/2504


a-wai:
Tested on staging: https://staging.kernelci.org:9000/viewer?search=parent%3D661fc4ef4d0bca045f7b318b

nuclearcat:
Thanks!

PR #552, opened 2024-04-17
src/timeout: set timed-out `checkout` result

JenySadadia:
Set timed-out `checkout` node result to `incomplete` while in `running` state. As it denotes that the node timed-out while checkout was still going on.


JenySadadia:
Replaced with https://github.com/kernelci/kernelci-pipeline/pull/553

PR #553, opened 2024-04-18
Handle incomplete/failed checkout

JenySadadia:
Fixes https://github.com/kernelci/kernelci-pipeline/issues/554
Depends on https://github.com/kernelci/kernelci-core/pull/2507

- Tarball updates source code repo and creates tarball. If update repo operation fails even with the second attempt, it means it failed to checkout souce code. Hence, update `checkout` node with state `done` state and result `fail`.
- Set timed-out `checkout` node result to `incomplete` while in `running` state. As it denotes that the node timed-out while checkout was still going on.

### DIFF
--- a/src/tarball.py
+++ b/src/tarball.py
@@ -136,6 +136,22 @@ git archive --format=tar --prefix={prefix}/ HEAD | gzip > {tarball_path}
             err_msg = json.loads(err.response.content).get("detail", [])
             self.log.error(err_msg)
 
+    def _update_failed_checkout_node(self, checkout_node, error_code, error_msg):
+        node = checkout_node.copy()
+        node.update({
+            'state': 'done',
+            'result': 'fail',
+        })
+        if 'data' not in node:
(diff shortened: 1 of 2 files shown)


nuclearcat (changes requested):
Maybe we can add error_code as well (probably need to introduce new one) to indicate reason why node is incomplete. It will help to troubleshoot incomplete issues.

JenySadadia:
> Maybe we can add error_code as well (probably need to introduce new one) to indicate reason why node is incomplete. It will help to troubleshoot incomplete issues.

Yes, definitely. I'll update the PR.

JenySadadia:
Timeout worked. Error info is not set as the updated PR just got deployed on staging.
https://staging.kernelci.org:9000/viewer?node_id=6621c39d5f2e02af29643459

nuclearcat:
I think its fine to merge, as its not crashing on staging, and if any corner case will appear, we will follow up

PR #562, opened 2024-04-25
lava_callback: Don't upload empty files to Azure

nuclearcat:
There is no use for lot of empty files on Azure,
that only complicate cleanup.


src/lava_callback.py
```diff
@@ -45,6 +45,10 @@ def _upload_log(log_parser, job_node, storage):
         # open gzip in explicit text mode to avoid platform-dependent line endings
         with gzip.open(os.path.join(tmp_dir, 'lava_log.txt.gz'), 'wt') as f:
             log_parser.get_text_log(f)
+            # get current position in file
+            pos = f.tell()
+            if pos == 0:
+                return None
```
a-wai, line 51:
```suggestion
            if f.tell() == 0:
                return None
```

No need for an intermediary variable here.

nuclearcat: Thanks, will update now

PR #570, opened 2024-04-26
src: lava_callback: add device ID to node data

a-wai:
It can be useful to know the exact device on which a job ran, without having to open the LAVA job page. This is done by querying the device ID from the callback data and appending it to the node data.

Depends on https://github.com/kernelci/kernelci-core/pull/2519
Fixes https://github.com/kernelci/kernelci-core/issues/2502

### DIFF
--- a/src/lava_callback.py
+++ b/src/lava_callback.py
@@ -115,7 +115,10 @@ def async_job_submit(api_helper, node_id, job_callback):
     job_node['result'] = job_result
     job_node['state'] = 'done'
     if device_id:
+        print(f"DEBUG: Habemus device: {device_id}")
         job_node['data']['device'] = device_id
+    else:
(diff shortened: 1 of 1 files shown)


a-wai:
Added a commit saving raw callback data as a node artifact.

nuclearcat:
Tested https://staging.kernelci.org:9000/viewer?node_id=662fc21f1efae6e502c6eefd

a-wai:
Thanks for merging @nuclearcat but I'm afraid it doesn't work properly (callback data is there, but the device ID isn't found) -- I was about to change it to "draft" just seconds after you merged ;)

Plus, this PR includes a debug commit, so I'll likely open a partial revert PR tomorrow after I figure out what's wrong here...

PR #571, opened 2024-04-26
jobs-chromeos.yaml: Disable module compression for every kernel version

nfraprado:
Commit d4bbe942098b ("kbuild: remove CONFIG_MODULE_COMPRESS"), introduced in kernel v5.13, substituted CONFIG_MODULE_COMPRESS=n for CONFIG_MODULE_COMPRESS_NONE=y as the way to disable module compression. Since module compression causes "Invalid ELF header magic: != ELF" errors during boot on the ChromeOS base config, add the missing config to disable module compression on kernels > v5.13 as well.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -14,6 +14,7 @@ _anchors:
       fragments:
         - arm64-chromebook
         - CONFIG_MODULE_COMPRESS=n
+        - CONFIG_MODULE_COMPRESS_NONE=y
     rules: &kbuild-gcc-10-arm64-chromeos-rules
       tree:
       - '!android'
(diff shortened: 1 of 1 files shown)


nfraprado:
This is one example of `Invalid ELF header magic: != ELF` errors currently happening on baseline-nfs jobs across chromebooks: https://lava.collabora.dev/scheduler/job/13529286

I couldn't see them when building locally with the same config & kernel version: https://lava.collabora.dev/scheduler/job/13535233
Though it does seem modules are failing to load since no network adapter is found.

By just adding the config in this PR, modules now load (like the r8152 network adapter driver) and boot succeeds: https://lava.collabora.dev/scheduler/job/13535191

a-wai (approved):

PR #580, opened 2024-04-29
result_summary_templates: fix error 'node' is undefined

helen-fornazier:
The object is named test and not node, so s/node/test


hardboprobot (approved):
eek! Good catch! Behold, the dangers of copying and pasting.

PR #583, opened 2024-04-30
Compress callback data and temporarily disable its upload

a-wai:
This PR disables callback data for now as it contains secrets we shouldn't leak.

While at it, also enable compression for this data and drop the "log" part as it's already uploaded separately. This is useless while the feature is disabled, but it'll be ready when we re-enable it.


a-wai:
Testing on staging, see `callback_data` from https://staging.kernelci.org:9000/viewer?node_id=66339a0158f5a43d7c38b52c

nuclearcat (approved):
LGTM, verified on staging

PR #587, opened 2024-04-30
result_summary: add error classification

helen-fornazier:
Classify errors according to patterns in the logs.

~~TODO: classify based on the concatenated logs instead of per log file~~
~~TODO: remove traces_config global var from utils~~

### DIFF
--- a/src/result_summary/utils.py
+++ b/src/result_summary/utils.py
@@ -8,8 +8,13 @@
 import gzip
 import re
 import requests
+import yaml
+from typing import Any, Dict
 
 import result_summary
+
+
+CONFIG_TRACES_FILE_PATH = './config/traces_config.yaml'
 
 
 def split_query_params(query_string):
@@ -101,6 +106,16 @@ def iterate_node_find(service, params):
     return nodes
 
 
+def get_err_category(trace: str, traces_config: Dict) -> Dict[str, Any]:
+    """Given a trace and a traces config, return its category"""
+    # sourcery skip: raise-specific-error
+    for category in traces_config["categories"]:
+        p = "|".join(category["patterns"])
+        if re.findall(p, trace or ""):
+            return category
+    raise Exception(f"No category found")
+
+
 def get_log(url, snippet_lines=0):
     """Fetches a text log given its url.
 
@@ -146,10 +161,10 @@ def get_logs(node):
 
     This method iterates over a node's 'artifacts', if present, to find
     log files. For each identified log file, it obtains the content by
-    calling the `_get_log` method with a number of context lines (10
-    last lines by default). If the content is not empty, it then stores
-    this log data in a dictionary, which includes both the URL of the
-    log and its text content.
+    calling the `_get_log` method.
+    If the content is not empty, it then stores this log data in a
+    dictionary, which includes both the URL of the log and its text
+    content.
 
     If the node log points to an empty file, the dict will contain an
     entry for the log with an empty value.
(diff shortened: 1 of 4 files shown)


src/result_summary/utils.py
```diff
@@ -8,10 +8,18 @@
 import gzip
 import re
 import requests
+import yaml
+from typing import Any, Dict
 
 import result_summary
 
 
+CONFIG_TRACES_FILE_PATH = './config/traces_config.yaml'
```
hardboprobot, line 17: I'd rather not have this hardcoded, but we can change that later.

src/result_summary/utils.py
```diff
@@ -175,8 +193,9 @@ def get_logs(node):
                 log_fields[artifact] = value
         for log_name, url in log_fields.items():
             text = get_log(url, snippet_lines=snippet_lines)
+            category = get_err_category(text)
```
hardboprobot, line 196: I'd prefer if we could do this as an independent stage in the `post_process_node()` process rather than have it intertwined in a function that's supposed to do only one thing (to get logs), but I understand we'd need to either re-download the log files or save the whole log for each node instead of just the snippets, so I think we'll have to do it like this, at least for now.

hardboprobot (commented):
Good job so far!

src/result_summary/utils.py
helen-fornazier: any suggestions where I can place this?

src/result_summary/utils.py
helen-fornazier: done, I moved to `post_process_node()`, please check.

src/result_summary/utils.py
helen-fornazier: I moved the global trace_config and this hardcoded var to the post processing function, not ideal but better than a global thing I guess, please check

hardboprobot: Ideally, to be parameterized either via a command line option or a parameter in a yaml file. The problem is that, while we're using a yaml file for configuration, it's meant to be used at the top level (result_summary.py), and the functions in utils.py are supposed to be generic and independent, so leaking and propagating the top-level configs to here wouldn't be a good idea. So we can leave it as it is: I'd rather have this visible at the top of the file than buried in a function, though.

src/result_summary/utils.py
```diff
@@ -215,4 +226,26 @@ def find_node_logs(node, api):
...
+    def concatenate_logs(logs):
+        concatenated_logs = ''
+        for log in logs:
+            concatenated_logs += logs[log]['text']
+        return concatenated_logs
+
     node['logs'] = find_node_logs(node, api)
+    concatenated_logs = concatenate_logs(node['logs'])
+
+    CONFIG_TRACES_FILE_PATH = './config/traces_config.yaml'
```
hardboprobot, line 244: I left a comment about this: keep the hardcoded file but move this back to the top of the file instead of here: more visible and obvious.

src/result_summary/utils.py
```diff
@@ -215,4 +226,26 @@ def find_node_logs(node, api):
...
+    with open(CONFIG_TRACES_FILE_PATH) as f:
+        traces_config: Dict[str, Any] = yaml.load(f, Loader=yaml.FullLoader)
+    node['category'] = get_err_category(concatenated_logs, traces_config)
+
+    # Only get the last 10 lines of the log
+    snippet_lines = -10
```
hardboprobot, line 250: Small nit, but from a caller perspective it'd be clearer to make this number positive here (number of snippet lines) and then, inside the function do `logs[log]['text'] = '\n'.join(lines[-snippet_lines:])` instead.

hardboprobot (commented):
Two minor details and we're good to go.

hardboprobot (approved):

PR #588, opened 2024-05-01
config: chromeos: Enable cpufreq kselftest

crazoes:
Enable cpufreq kselftest on all the trees and branches.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -279,6 +279,15 @@ jobs:
     kind: test
     image: kernelci/staging-kernelci
 
+  kselftest-cpufreq:
+    template: kselftest.jinja2
+    kind: test
+    params:
+      nfsroot: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240313.0/{debarch}'
+      collections: cpufreq
+      job_timeout: 10
+    kcidb_test_suite: kselftest.cpufreq
+
   kselftest-dt:
     template: kselftest.jinja2
     kind: test
--- a/config/scheduler-chromeos.yaml
+++ b/config/scheduler-chromeos.yaml
@@ -187,6 +187,18 @@ scheduler:
       result: pass
     platforms: *mediatek-platforms
 
+  - job: kselftest-cpufreq
+    <<: *test-job-x86-intel
+
+  - job: kselftest-cpufreq
+    <<: *test-job-x86-amd
+
+  - job: kselftest-cpufreq
+    <<: *test-job-arm64-qualcomm
+
+  - job: kselftest-cpufreq
+    <<: *test-job-arm64-mediatek
+
   - job: tast-basic-arm64-mediatek
     <<: *test-job-chromeos-mediatek
 


crazoes:
@laura-nao @nuclearcat I am not sure how to enable this test for multiple platforms. For now I have selected `test-job-x86-intel` but I think this test can be enabled on arm platforms as well

laura-nao:
> @laura-nao @nuclearcat I am not sure how to enable this test for multiple platforms. For now I have selected `test-job-x86-intel` but I think this test can be enabled on arm platforms as well

To run the test on other platforms, you can just add a new job entry with the same name - e.g. to run the test on AMD Chromebooks:

```
  - job: kselftest-cpufreq
    <<: *test-job-x86-amd
```

Same applies for ARM64 Chromebooks (you can use the `test-job-arm64-mediatek` and `test-job-arm64-qualcomm` anchors for that).

As a side note - I see the cpufreq kselftest doesn't have KTAP output, so I'm not sure how the result parser will handle that (some results from staging should come back soon to confirm that). Regardless of whether the parser can handle non-KTAP results, it's probably worth looking into making the test KTAP compliant too.

config/jobs-chromeos.yaml
```diff
@@ -435,6 +435,14 @@ jobs:
...
         - for-kernelci
 
+  kselftest-cpufreq:
+    template: kselftest.jinja2
+    kind: test
+    params:
+      nfsroot: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240313.0/{debarch}'
+      collections: cpufreq
+      job_timeout: 10
+
```
JenySadadia, line 445: Could you please add `kcidb_test_suite` here? It is a mapping of KCI test with KCIDB test suite. We would require it for submitting results to KCIDB.
```suggestion
  kselftest-cpufreq:
    template: kselftest.jinja2
    kind: test
    params:
      nfsroot: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240313.0/{debarch}'
      collections: cpufreq
      job_timeout: 10
    kcidb_test_suite: kselftest.cpufreq
```

crazoes: Sure, I'll make this change in v2

crazoes:
@JenySadadia @nuclearcat I made the changes as requested and this should be good to merge in staging if everything looks good to you.

config/jobs-chromeos.yaml
```diff
@@ -435,6 +435,15 @@ jobs:
...
 
+  kselftest-cpufreq:
+    template: kselftest.jinja2
+    kind: test
+    params:
+      nfsroot: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240313.0/{debarch}'
+      collections: cpufreq
+      job_timeout: 10
+    kcidb_test_suite: kselftest.cpufreq
+
```
a-wai, line 446: This would be better off in `pipeline.yaml` as it definitely isn't chrome{os,book}-specific (and we already have `kselftest-dt` defined there).

LGTM otherwise, thanks!

crazoes: Thanks @a-wai, I've forced pushed the changes.

nuclearcat:
Tested on staging, thanks!

PR #590, opened 2024-05-02
tarball: Implement checkout for specific commit

nuclearcat:
We often need not ToT, but specific commit, implement this.

### DIFF
--- a/src/tarball.py
+++ b/src/tarball.py
@@ -58,6 +58,21 @@ git archive --format=tar --prefix={prefix}/ HEAD | gzip > {tarball_path}
             if config.tree.name == tree and config.branch == branch:
                 return config
 
+    def _find_build_commit(self, node):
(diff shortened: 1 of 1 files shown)


src/tarball.py
```diff
@@ -58,6 +58,21 @@ def _find_build_config(self, node):
             if config.tree.name == tree and config.branch == branch:
                 return config
 
+    def _find_build_commit(self, node):
+        revision = node['data'].get('kernel_revision')
```
a-wai, line 62: Is it 100% guaranteed that the node data has `kernel_revision`?

nuclearcat: Yes, otherwise checkout node doesn't make sense

a-wai: Alright then (I'm not familiar with those early pipeline stages)

a-wai (approved):

PR #595, opened 2024-05-07
Add all supported fluster tests to run on all the boards

musamaanjum:
Please look at the commit description for individual descriptions of patches.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -378,7 +378,8 @@ _anchors:
       tree:
         - mainline
         - next
-        - kernelci
+        - collabora-chromeos-kernel
+        - media
 
 jobs:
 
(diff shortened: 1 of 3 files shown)


config/jobs-chromeos.yaml
```diff
@@ -379,6 +379,9 @@ _anchors:
         - mainline
         - next
         - kernelci
+        - collabora-chromeos-kernel_for-kernelci
+        - media_master
+        - media_fixes
```
a-wai, line 384:
```suggestion
        - collabora-chromeos-kernel
        - media
```

Only trees (and/or branches) can be added there, build configs cannot (yet)

config/scheduler-chromeos.yaml
```diff
@@ -389,3 +389,53 @@ scheduler:
...
+  - job: v4l2-decoder-conformance-av1-chromium-10bit
+    <<: *test-job-arm64-mediatek
+    platforms:
+      - mt8195-cherry-tomato-r2
+
+  - job: v4l2-decoder-conformance-h264
+    <<: *test-job-arm64-mediatek
+    platforms:
+      - sc7180-trogdor-kingoftown
+      - sc7180-trogdor-lazor-limozeen
```
a-wai, line 402: This won't work as `test-job-arm64-mediatek` builds the kernel for mediatek devices only; please create additional scheduler entries for qualcomm devices.

a-wai (changes requested):

config/jobs-chromeos.yaml
musamaanjum: I see. I'll update you. Is the master branch being selected by default right now?

config/scheduler-chromeos.yaml
musamaanjum: I'll update it in next change.

config/scheduler-chromeos.yaml
```diff
@@ -389,3 +389,73 @@ scheduler:
...
+    platforms:
+      - mt8186-corsola-steelix-sku131072
+      - mt8192-asurada-spherion-r0
+      - mt8195-cherry-tomato-r2
+
+  - job: v4l2-decoder-conformance-h264
+    <<: *test-job-arm64-qualcomm
+    platforms:
+      - sc7180-trogdor-kingoftown
+      - sc7180-trogdor-lazor-limozeen
```
a-wai, line 437:
```suggestion
```

This is already included in `test-job-arm64-qualcomm` (same for the following entries)

config/scheduler-chromeos.yaml
```diff
@@ -389,3 +389,73 @@ scheduler:
...
+    platforms:
+      - mt8195-cherry-tomato-r2
+
+  - job: v4l2-decoder-conformance-h264
+    <<: *test-job-arm64-mediatek
+    platforms:
+      - mt8183-kukui-jacuzzi-juniper-sku16
+      - mt8186-corsola-steelix-sku131072
+      - mt8192-asurada-spherion-r0
+      - mt8195-cherry-tomato-r2
```
a-wai, line 404:
```suggestion
```

Already configured in `test-job-arm64-mediatek`

a-wai (changes requested):

config/jobs-chromeos.yaml
a-wai: No, there is no currently no way to add rules for specific combinations of trees/branches. The rules mean basically `run job if $tree in $allowed_trees and $branch in $allowed_branches`.

You can just list the trees, it will only use the branches configured in build_configs:
* `mainline` will use `master`
* `next` will use `master`
* `kernelci` will use `staging-mainline`, `staging-next` and `staging-stable`; 
* `collabora-chromeos-kernel` will use `for-kernelci`
* `media` will use `master` and `fixes`

One thing you could do is the following, if you really want/need to add branch-based rules:

```yaml
    tree:
      - mainline
      - next
      - collabora-chromeos-kernel
      - media
    branch:
      - master
      - fixes
      - for-kernelci
```

This won't prevent from running on e.g. `media/for-kernelci` if it were to exist some day, but is a bit more restrictive than using tree-based rules only. Please note this isn't needed at all for now as we don't have any tree/branch combination that wouldn't match.

Note: wasn't `kernelci` supposed to be removed from the trees list?

a-wai: Created https://github.com/kernelci/kernelci-core/issues/2548 for the feature request

config/scheduler-chromeos.yaml
musamaanjum: Done

musamaanjum:
@a-wai Thank you for reviewing and suggesting changes as well.

nuclearcat:
Thanks Usama, looks like results are present on next kernel tests: https://staging.kernelci.org:9000/viewer?search=parent%3D663b348103781dd7347c72c3
Merging

PR #596, opened 2024-05-07
trigger: Add exclude/include tree list for trigger

nuclearcat:
As we need to restrict list of running kernels on staging, we need to add option allowing that.
Also it will be good to exclude staging kernels from production kernel list.

So in case of staging we need to run kernels only from tree "kernelci"
and sometimes something else, for example "mediatek".
Option will look like:

--trees kernelci,mediatek
or
--trees kernelci

On production we need to exclude trees kernelci and buggytree:
--trees !kernelci,buggytree
or just kernelci:
--trees !kernelci

Purpose of this option is that our compiling capacity is limited,
and right now staging and production both compiling very large set
of kernels, we need to reduce this amount to drop costs.

### DIFF
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -33,7 +33,15 @@ class Trigger(Service):
     def _log_revision(self, message, build_config, head_commit):
         self.log.info(f"{message:32s} {build_config.name:32s} {head_commit}")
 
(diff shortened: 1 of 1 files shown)


nuclearcat:
Tested on staging, will be applied on upcoming production update

PR #601, opened 2024-05-09
src: update return values of `APIHelper.receive_event_node`

JenySadadia:
Depends on https://github.com/kernelci/kernelci-core/pull/2550

`APIHelper.receive_event_node` method is used to receive node data from PubSub event. The method has been updated to return `is_hierarchy` flag as well which represents events related to node hierarchy.
Update pipeline services using the method accordingly.


JenySadadia:
Tested OK on staging.

PR #602, opened 2024-05-13
config/pipeline: enable android branches from legacy

helen-fornazier:
Enable all android branches from the legacy system


nuclearcat (approved):
Can see example builds at: https://staging.kernelci.org:9000/viewer?search=parent%3D664c54ab3844ef9988987bd6

PR #603, opened 2024-05-14
config: platforms-chromeos: use CrOS R124 files

a-wai:
ChromeBooks were upgraded with a new image based on ChromiumOS R124, so we must use those files now.


nuclearcat:
Thanks!

PR #605, opened 2024-05-17
config: Use media-stage instead of media-tree

musamaanjum:


nuclearcat (approved):
Works,  thanks https://staging.kernelci.org:9000/viewer?node_id=664773d0c990379397bfbd9e

PR #607, opened 2024-05-22
pipeline.yaml: Update RISC-V configs

nuclearcat:
1)rv32 defconfig doesn't exist, remove
2)nommu_k210_defconfig have modules disabled


nuclearcat:
Tested on staging, results is better now (no build failures)

PR #615, opened 2024-05-27
lava_callback.py: Remove error_code/error_msg on lava-callback

nuclearcat:
Sometimes due congestion node might be set to timeout, but then result might arrive late and we need to use it properly.

### DIFF
--- a/src/lava_callback.py
+++ b/src/lava_callback.py
@@ -121,6 +121,9 @@ def async_job_submit(api_helper, node_id, job_callback):
     # failed LAVA job should have result set to 'incomplete'
     job_node['result'] = job_result
     job_node['state'] = 'done'
+    if job_node.get('error_code') == 'node_timeout':
+        job_node['error_code'] = None
+        job_node['error_msg'] = None
     if device_id:
         job_node['data']['device'] = device_id
     hierarchy = job_callback.get_hierarchy(results, job_node)


src/lava_callback.py
```diff
@@ -121,6 +121,8 @@ def async_job_submit(api_helper, node_id, job_callback):
     # failed LAVA job should have result set to 'incomplete'
     job_node['result'] = job_result
     job_node['state'] = 'done'
+    job_node['error_code'] = None
+    job_node['error_msg'] = None
```
JenySadadia, line 125: We should verify `job['result']` before re-setting error metadata.
If the result is set to `incomplete` with `error_code = node_timeout`, and if the current `job_result` is different from what is already set, then only the error metadata should be reset.

nuclearcat: Maybe it is better to verify and delete just node_timeout error in this case, and keep any other?
As scheduler assumed this node job disappeared and set timeout, but long lost job returned home, so we can erase this particular error.

JenySadadia: Yes, that's true.
My point is we need to check current node result for re-setting error related fields.

nuclearcat: Fixed, can you check if it is fine for you? On staging results looks better now on overload situations.

JenySadadia: Yes, looks good to me.

JenySadadia (approved):
LGTM. Thanks.

PR #616, opened 2024-05-28
lava_callback: Sometimes we get totally invalid log file uploaded

nuclearcat:
Most likely problems lays in threading of flask, and possibly callbacks are getting mixed. This commit attempts to introduce several countermeasures against that.


nuclearcat:
Tested on staging

PR #617, opened 2024-05-28
config: chromeos: enable watchdog reset test on Chromebooks

laura-nao:
Add a basic test to verify watchdog reset functionality. Enable the test on all ARM64 and AMD x86_64 Chromebooks. For Intel Chromebooks, enable the test only on octopus, as ACPI PM Timer on the other devices has been disabled in coreboot.

Requires: https://github.com/kernelci/kernelci-core/pull/2562

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -369,6 +369,13 @@ _anchors:
         - collabora-chromeos-kernel
         - media
 
+  watchdog-reset: &watchdog-reset-job
+    template: watchdog-reset.jinja2
+    kind: test
+    params: &watchdog-reset-job-params
+      bl_message: 'coreboot-'
+      wdt_dev: 'watchdog0'
+
 jobs:
 
   baseline-arm64-mediatek: &baseline-job
@@ -617,3 +624,8 @@ jobs:
       decoders:
         - 'GStreamer-VP9-V4L2-Gst1.0'
         - 'GStreamer-VP9-V4L2SL-Gst1.0'
+
+  watchdog-reset-arm64-mediatek: *watchdog-reset-job
+  watchdog-reset-arm64-qualcomm: *watchdog-reset-job
+  watchdog-reset-x86-amd: *watchdog-reset-job
+  watchdog-reset-x86-intel: *watchdog-reset-job
--- a/config/runtime/watchdog-reset.jinja2
+++ b/config/runtime/watchdog-reset.jinja2
@@ -1,0 +1,3 @@
+{%- set test_method = 'watchdog-reset' %}
+{%- set base_template = 'base/' + runtime + '.jinja2' %}
+{%- extends base_template %}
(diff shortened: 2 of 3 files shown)


config/jobs-chromeos.yaml
```diff
@@ -369,6 +369,21 @@ _anchors:
         - collabora-chromeos-kernel
         - media
 
+  watchdog-reset-arm64: &watchdog-reset-arm64-job
```
a-wai, line 372:
```suggestion
  watchdog-reset: &watchdog-reset-job
```

This is in no way architecture-dependent, so we can keep the naming arch-agnostic as well (same for the params) and get rid of `watchdog-reset-x86`.

config/runtime/watchdog-reset.jinja2
```diff
@@ -0,0 +1,4 @@
+{%- set boot_commands = 'ramdisk' %}
+{%- set test_method = 'watchdog-reset' %}
+{%- set base_template = 'base/' + runtime + '.jinja2' %}
+{%- extends base_template %}
```
a-wai, line 4: This is fine for now, but I'd suggest factoring this into a single generic test template at some point: with job params such as `test_method` (and potentially `boot_commands` as well) this shouldn't be too difficult and would allow us to reduce the number of templates (at least `kselftest`, `sleep` and `v4l2-decoder-conformance` would benefit from it, in addition to this one, maybe even `baseline` can be made to work that way).

config/runtime/watchdog-reset.jinja2
```diff
@@ -0,0 +1,4 @@
+{%- set boot_commands = 'ramdisk' %}
```
a-wai, line 1: Unneeded, if `boot_commands` isn't set it defaults to `ramdisk`

a-wai (changes requested):

config/jobs-chromeos.yaml
laura-nao: Right - will do.

config/runtime/watchdog-reset.jinja2
laura-nao: Makes sense. I can tackle this as a separate PR after this gets merged.

a-wai (approved):
LGTM, looks like it can be tested on staging

laura-nao:
Looking at the results on staging, it looks like the `timeout` attribute might not be always available, or the configuration for certain platforms is incorrect (wrong wdt device selected) - e.g. https://lava.collabora.dev/scheduler/job/14092849#L2552

I'll have a look at the failures and fix the test configuration accordingly.

laura-nao:
> Looking at the results on staging, it looks like the `timeout` attribute might not be always available, or the configuration for certain platforms is incorrect (wrong wdt device selected) - e.g. https://lava.collabora.dev/scheduler/job/14092849#L2552
> 
> I'll have a look at the failures and fix the test configuration accordingly.

I forgot to add the required watchdog options to the x86-board and arm64-chromebook fragments, hence the failed tests on staging.

I updated https://github.com/kernelci/kernelci-core/pull/2562 to add the missing kernel configs.

laura-nao:
Checked the results on staging again, all the driver dependencies seem now satisfied.

Made a small adjustment on the job template (see: https://github.com/kernelci/kernelci-core/pull/2562 ), other than that results seem ok to me.

PR #618, opened 2024-05-28
Enable CIP lab

JenySadadia:
Addresses https://github.com/kernelci/kernelci-project/issues/361

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -223,6 +223,7 @@ jobs:
   baseline-x86: *baseline-job
   baseline-x86-baylibre: *baseline-job
   baseline-x86-qualcomm: *baseline-job
+  baseline-x86-cip: *baseline-job
 
   kbuild-gcc-10-arc-haps_hs_smp_defconfig:
     <<: *kbuild-job
@@ -683,6 +684,17 @@ scheduler:
     runtime:
       type: lava
       name: lava-qualcomm
+    platforms:
+      - qemu
+
+  - job: baseline-x86-cip
+    event:
+      channel: node
+      name: kbuild-gcc-10-x86
+      result: pass
+    runtime:
+      type: lava
+      name: lava-cip
     platforms:
       - qemu
 
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -194,6 +194,14 @@ runtimes:
         token: kernelci-lab-qualcomm
         url: https://staging.kernelci.org:9100
 
+  lava-cip:
+    lab_type: lava
+    url: 'https://lava.ciplatform.org/'
+    notify:
+      callback:
+        token: kernel-ci-new-api
+        url: https://staging.kernelci.org:9100
+
   shell:
     lab_type: shell
 
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -90,6 +90,7 @@ services:
       - 'lava-broonie'
       - 'lava-baylibre'
       - 'lava-qualcomm'
+      - 'lava-cip'
     extra_hosts:
       - "host.docker.internal:host-gateway"
 


JenySadadia:
Hello @patersonc 
I enabled CIP in the new system and tried to submit `baseline` job for `x86`.
Here is the job link: https://lava.ciplatform.org/scheduler/job/1139445
There are the same failures in the jobs submitted by legacy system as well. See https://lava.ciplatform.org/scheduler/job/1139379#bottom.
Is this the issue from the lab side or do you expect KernelCI to handle this job for CIP differently?

patersonc:
> Hello @patersonc I enabled CIP in the new system and tried to submit `baseline` job for `x86`. Here is the job link: https://lava.ciplatform.org/scheduler/job/1139445 There are the same failures in the jobs submitted by legacy system as well. See https://lava.ciplatform.org/scheduler/job/1139379#bottom. Is this the issue from the lab side or do you expect KernelCI to handle this job for CIP differently?

Hi @JenySadadia  
It looks like that job is using docker.
We're using lava-docker (based on KernelCI's [lava-docker](https://github.com/kernelci/lava-docker)) for our labs, and I'm guessing that docker-in-docker isn't supported.

patersonc:
Ah, maybe it is supported and we need to manually enable it:
`use_docker:			Permit to use docker commands in slave`

patersonc:
> Ah, maybe it is supported and we need to manually enable it: `use_docker: Permit to use docker commands in slave`

Yep, that fixed it. Here's the same job, now run in a lab with `use_docker`: https://lava.ciplatform.org/scheduler/job/1139581
I only have access to enable this setting in two of our five labs though - I'll need to get the other maintainers to update settings their end.

So technically this MR is working (as well as the legacy system). I'll let you decide if you want to wait for us to get dind support enabled in all of our labs before merging or not.

patersonc:
Hi @JenySadadia 

> I'll wait until this is fixed in all the CIP labs.

This should be okay now. Thanks.

> Meanwhile, what are other tests you want to enable for CIP in the new system?

Is there a list to choose from somewhere?  
What would be most useful to KernelCI?

JenySadadia:
> Is there a list to choose from somewhere? What would be most useful to KernelCI?

Here are the tests the legacy system runs with CIP lab:
https://github.com/kernelci/kernelci-core/blob/main/config/core/runtime-configs.yaml#L38

But not all the jobs are enabled with the new system yet. Here is the tests enabled atm:
https://github.com/kernelci/kernelci-pipeline/blob/main/config/pipeline.yaml#L193

As per the latest announcement from @padovan, we would be enabling limited number of tests in the new system as as we transit through legacy system sunset.
So it depends on what are the important tests for CIP atm and then we can discuss further.

pawiecz (approved):

patersonc:
Hello,

Looking at https://linux.kernelci.org/test/job/cip/branch/linux-5.10.y-cip/kernel/v5.10.216-cip47/, it doesn't seem like the list of tests being run matches up to the list in https://github.com/kernelci/kernelci-core/blob/main/config/core/runtime-configs.yaml#L38.

Am I missing something?

patersonc:
Ah yes, thanks Jeny.

If possible can we keep the same tests as currently run on the CIP trees? If it would help, we can run more tests in lab-cip rather than only run a subset; we have capacity.

padovan:
@patersonc if you want to carry over all the tests that is fine.  We have documentation on how to enable the tests.

nuclearcat:
Merging, i suggest if there is follow-up work planned, to open it as issue, so it wont be blocking other PR.
Thanks!

JenySadadia:
Hello @patersonc 
Here is the documentation on enabling new kernel tree and tests in KernelCI https://docs.kernelci.org/api_pipeline/pipeline/developer-documentation/

PR #621, opened 2024-05-29
lava_callback.py: Sanitize lava log data

nuclearcat:
As we use this data in reports, lets remove all
non-printable characters as they confuse grafana, browsers and others.

### DIFF
--- a/src/lava_callback.py
+++ b/src/lava_callback.py
@@ -75,6 +75,12 @@ def _upload_log(log_parser, job_node, storage):
             data = log_parser.get_text()
             if not data or len(data) == 0:
                 return None
+            # Delete NULL characters from log data
+            data = data.replace('\x00', '')
+            # Sanitize log data from non-printable characters (except newline)
+            # replace them with '?', original still exists in cb data
+            data = ''.join([c if c.isprintable() or c == '\n' else
+                            '?' for c in data])
             f.write(data)
         src = os.path.join(tmp_dir, 'lava_log.txt.gz')
         return _upload_file(storage, job_node, src, 'log.txt.gz')


nuclearcat:
Unfortunately there is way more than just null characters.
For example take a look at: https://staging.kernelci.org:9000/viewer?node_id=6656ec0f8b971c830cc01c07
(This one not filtered yet)

JenySadadia:
> Unfortunately there is way more than just null characters. For example take a look at: https://staging.kernelci.org:9000/viewer?node_id=6656ec0f8b971c830cc01c07 (This one not filtered yet)

Yes, but do you think replacing them with `?` would solve the issue?
Although this fix is not related to just KCIDB, from the KCIDB integration PoV, I'll be still getting validation error.

nuclearcat:
I guess kcidb want to see only printable characters, isn't it?
What exactly is issue of kcidb validation is not that?

JenySadadia:
> One issue is you are replacing new line char so it tries to accommodate everything in the log field now.

Check the log file https://staging.kernelci.org:9000/viewer?node_id=6657cdda2b005d102de7e3a1. Now, it just has one line with all the logs.

nuclearcat:
I don't think i can truncate also length of lines, it might truncate some useful data, if some tests will generate very long lines for example.
Then we should do additional filtering in kcidb bridge according to grafana requirements.

JenySadadia:
Can we just drop all the `\x00` chars from the log file if it won't cause any issues?

JenySadadia (approved):
Looks good on staging.
Verified log files.

nuclearcat:
Thanks, merging

PR #624, opened 2024-05-30
config: jobs-chromeos: Add lab-setup fragment

nfraprado:
Add the lab-setup fragment to the chromebook builds, which contains the architecture independent kernel configs needed to run tests on the platform. Notably this disables IP autoconfig by the kernel.

The result of this change is that the 12 seconds boot delay and the consequent deferred probe pending warnings will no longer happen on any platform. Particularly on mt8186-corsola-steelix-sku131072 (due to a different network adapter being used) on which it was still happening.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -12,6 +12,7 @@ _anchors:
       defconfig: 'cros://chromeos-{krev}/{crosarch}/chromiumos-{flavour}.flavour.config'
       flavour: '{crosarch}-generic'
       fragments:
+        - lab-setup
         - arm64-chromebook
         - CONFIG_MODULE_COMPRESS=n
         - CONFIG_MODULE_COMPRESS_NONE=y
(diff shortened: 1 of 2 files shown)


a-wai (approved):

a-wai:
Shouldn't this fragment be added to all kbuilds which are tested on chromebooks? (so pretty much all of the x86 and arm64 builds, except the android ones)

nfraprado:
@a-wai Yes, I was missing `kbuild-gcc-10-x86`. That's the only other kbuild that has either `arm64-chromebook` or `x86-board`, which are the rules for Chromebooks

 In reality this fragment should also be added for non-Chromebooks, but I'd prefer to enable for Chromebooks first and see some results before adding it to others.

a-wai:
Perfect, thanks!

a-wai (approved):

nuclearcat:
Can you review results on staging if they are as expected?

PR #625, opened 2024-05-31
config/runtime/kunit.jinja2: fix result map

JenySadadia:
Addresses one of the issues from https://github.com/kernelci/kernelci-api/issues/536

Fix result map for skipped tests. Initially, API didn't have `skip` available node result in the schema. That's why it was 
mapped to `None` result. But now API has `skip` result to denote skipped tests.
Fix the result mapping accordingly.


pawiecz (approved):
LGTM

PR #626, opened 2024-06-03
config: platforms-chromeos: Add serial delay for some MediaTek platforms

nfraprado:
Add test_character_delay to the Spherion and Tomato platforms to workaround the fact that they're sometimes unable to process serial input fast enough, resulting in mangled commands and consequently flaky test results, as described in
https://github.com/kernelci/kernelci-project/issues/366.

The right place to do this change would be in the device-type template as described in LAVA's documentation [1]. This overriding in KernelCI is meant only as a temporary workaround to verify whether this fixes the issue. If it does, then we'll do it in LAVA upstream instead.

[1] https://docs.lavasoftware.org/lava/debugging.html#differences-in-input-speeds
Signed-off-by: Nícolas F. R. A. Prado <nfraprado@collabora.com>

### DIFF
--- a/config/platforms-chromeos.yaml
+++ b/config/platforms-chromeos.yaml
@@ -118,6 +118,8 @@ platforms:
       flash_kernel:
         url: https://storage.chromeos.kernelci.org/images/rootfs/chromeos/chromiumos-corsola/20240514.0/arm64
         image: 'Image'
+    context:
+      test_character_delay: 10
     rules:
       <<: *arm64-chromebook-device-rules
       min_version:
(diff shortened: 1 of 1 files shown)


nuclearcat:
Is it works fine and improve things? If it is tested, i can merge it.

laura-nao:
@nfraprado I just caught an instance of `input overrun` on `corsola` as well, causing a false regression being reported in the watchdog-reset test: https://lava.collabora.dev/scheduler/job/14402618#L2265 (KernelCI node: https://staging.kernelci.org:9000/viewer?node_id=6670b75ed0b7694b3998a7ea)

It'd be great to have the delay added for `corsola` too in this PR, so we can confirm that it fixes the watchdog-reset issue. Alternatively, I can open a separate PR to tackle this.

nfraprado:
@laura-nao good catch! I've added corsola to the PR now.

PR #628, opened 2024-06-04
Re-structure test hierarchy with `kind=job` for test suites

JenySadadia:
Depends on https://github.com/kernelci/kernelci-core/pull/2571

### DIFF
--- a/config/runtime/kunit.jinja2
+++ b/config/runtime/kunit.jinja2
@@ -36,11 +36,15 @@ class Job(BaseJob):
                 'node': {
                     'name': test_case['name'],
                     'result': RESULT_MAP[test_case['status']],
+                    'kind': 'test'
                 },
                 'child_nodes': [],
             })
         for sub_group in group.get('sub_groups', []):
             child_nodes.append(self._parse_results(sub_group))
+
+        node['kind'] = 'job' if child_nodes else 'test'
+
         return {
             'node': node,
             'child_nodes': child_nodes,
@@ -130,6 +134,7 @@ cd {src_path}
             'child_nodes': [
                 {
                     'node': {
+                        'kind': 'job' if child_nodes else 'test' ,
                         'name': name,
                         'result': result,
                     },
--- a/config/runtime/kunit.jinja2
+++ b/config/runtime/kunit.jinja2
@@ -152,5 +152,4 @@ cd {src_path}
         api_helper = kernelci.api.helper.APIHelper(self._api)
         api_helper.submit_results(result, self._node)
 
-        return node
 {% endblock %}
--- a/config/runtime/kunit.jinja2
+++ b/config/runtime/kunit.jinja2
@@ -12,6 +12,7 @@ import subprocess
 {%- block python_local_imports %}
 {{ super() }}
 import kernelci.api.helper
+import kernelci.runtime
 {%- endblock %}
 
 {%- block python_globals %}
@@ -44,6 +45,8 @@ class Job(BaseJob):
             child_nodes.append(self._parse_results(sub_group))
 
         node['kind'] = 'job' if child_nodes else 'test'
+        if node['kind'] == 'job':
+            node['result'] = kernelci.runtime.evaluate_test_suite_result(child_nodes)
 
         return {
             'node': node,
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -55,7 +55,7 @@ _anchors:
 
   tast: &tast-job
     template: tast.jinja2
-    kind: test
+    kind: job
     rules: *min-5_4-rules
     kcidb_test_suite: tast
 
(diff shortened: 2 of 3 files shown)


config/runtime/kunit.jinja2
```diff
@@ -152,5 +152,4 @@ cd {src_path}
         api_helper = kernelci.api.helper.APIHelper(self._api)
         api_helper.submit_results(result, self._node)
 
-        return node
```
pawiecz, line 150: It looks like this bit should have been adjusted to `return self._node` in https://github.com/kernelci/kernelci-pipeline/commit/33fa931f20a1fceb2e278989e4c4c13da75e02d3.

I'd suggest fixing it here to comply with superclass function signature (i.e. returning updated Node) but I'm not 100% sure if this case applies here :thinking: (it made sense back in [`42c2ceb`](https://github.com/kernelci/kernelci-pipeline/commit/42c2cebe45d4abdf0f3be8981b7b722a0ad239fd#diff-4e48889b22e74380181ae303dd5372a80219e386747e5830863a06a26ee6519cL84))

pawiecz (changes requested):

config/runtime/kunit.jinja2
JenySadadia: Thanks for digging in to find relevant commits. Your suggestion makes sense but the returned value is not being used anywhere.
The `main` function, once the job is generated would look like this:
```
def main(args):
    # Sometimes container DNS is failing if k8s node is still booting
    # This is a workaround:
    # test www.google.com until it works or retry 10 times
    for _ in range(30):
        try:
            socket.gethostbyname('www.google.com')
            break
        except socket.gaierror:
            pass
        # sleep 1 second
        time.sleep(1)
    else:
        print("DNS resolution failed, exiting", file=sys.stderr)
        sys.exit(1)

    job = Job(api_config=API_CONFIG_YAML, node=NODE, workspace=WORKSPACE)
    try:
        results = job.run()
        job.submit(results)
    except requests.exceptions.HTTPError as ex:
        print(ex, file=sys.stderr)
        detail = ex.response.json().get('detail')
        if detail:
            print(detail, file=sys.stderr)
        results = None
    except Exception:
        print(traceback.format_exc())
        results = None

    return results


if __name__ == '__main__':
    results = main(sys.argv)
    sys.exit(1 if results is None else 0)
```
`job.submit` internall calls `job._submit` and its returned value is not being used.

JenySadadia:
> Child job node results of `exec` are missing somehow. I need to fix it before we can merge this.

The issue is fixed:
https://staging.kernelci.org:9000/viewer?search=parent%3D66728229ac160d93e7dcc6cc

config/runtime/kunit.jinja2
```diff
@@ -26,6 +26,17 @@ ARCH = '{{ arch }}'
...
+    def _evaluate_job_node_result(self, child_nodes):
+        result = None
+        result_list = [child['node']['result'] for child in child_nodes]
+        if 'fail' in result_list:
+            result = 'fail'
+        elif all(result == 'pass' for result in result_list) or 'pass' in result_list:
+            result = 'pass'
+        elif all(result == 'skip' for result in result_list) or 'skip' in result_list:
+            result = 'skip'
+        return result
```
pawiecz, line 38: This job result evaluation procedure is the same as the one [propsed for LAVA](https://github.com/kernelci/kernelci-core/blob/928b1f91b6714078affc1c0873083e7a75184b16/kernelci/runtime/lava.py#L167) - would it be possible to extract it to a common package and reuse it instead of duplicating?

pawiecz (changes requested):

config/runtime/kunit.jinja2
JenySadadia: Yes, I can try that.

JenySadadia: Updated the code accordingly. 
Moved the method to `kernelci.runtime`.

JenySadadia: Tested OK on staging https://staging.kernelci.org:9000/viewer?node_id=6673de03638913d5a6abf4bd

config/runtime/kunit.jinja2
JenySadadia: What do you think about the above comment @pawiecz ?

pawiecz: I agree that it makes no sense to forcefully keep the return value if it's not used anymore.

I'd suggest checking if the superclass method is called the same way (discarding returned Node) and align them so that such a mismatch doesn't bite us in the long run (e.g. during debugging)

config/runtime/kunit.jinja2
pawiecz: Looks good in https://github.com/kernelci/kernelci-core/pull/2571/files#diff-9ebe59937a7e94546a071f481a86fd3cd0ae41b4534c8cfe858eb4cf75f35b89R242 :+1: 

This method is now safe to be dropped from this patch

config/runtime/kunit.jinja2
JenySadadia: Thanks for the suggestion. Created https://github.com/kernelci/kernelci-core/pull/2582

pawiecz (approved):
LGTM

PR #632, opened 2024-06-07
src/send_kcidb: send tests only if KCIDB test mapping exists

JenySadadia:
All test suite definitions must have `kcidb_test_suite` property i.e. KCIDB test suite mapping.
Only send tests for those the mapping is found.

### DIFF
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -199,6 +199,10 @@ class KCIDBBridge(Service):
                     suite_name = self._jobs[sub_path].kcidb_test_suite
                     if suite_name:
                         new_path.append(suite_name)
+                    else:
+                        self.log.error(f"KCIDB test suite mapping not found for \
+the test: {sub_path}")
+                        return None
                 else:
                     new_path.append(sub_path)
             # Handle path such as ['tast-ui-x86-intel', 'tast', 'os-release'] converted
@@ -357,6 +361,9 @@ in {test_node['data'].get('runtime')}",
         test_node, build_node = self._parse_test_node(
             origin, node
         )
+        if not test_node['path']:
+            self.log.info(f"Not sending test as path information is missing: {test_node['id']}")
+            return
         parsed_test_node.append(test_node)
         if build_node:
             parsed_build_node.append(build_node)


src/send_kcidb.py
```diff
@@ -331,12 +331,12 @@ def _parse_test_node(self, origin, test_node):
         if artifacts:
             parsed_test_node['output_files'] = self._get_output_files(
                 artifacts=artifacts,
-                exclude_properties=('lava_log', 'test_log')
+                exclude_properties=('lava_log', 'job_txt')
```
pawiecz, line 334: Is there any other type of job that logs to `test_log` or `job_txt`?

Would it make sense to align `kunit` to log to `test_log`?

config/jobs-chromeos.yaml
```diff
@@ -379,7 +379,7 @@ _anchors:
     params: &watchdog-reset-job-params
       bl_message: 'coreboot-'
       wdt_dev: 'watchdog0'
-    kcidb_test_suite: kernelci_watchdog_reset
+    # kcidb_test_suite: kernelci_watchdog_reset
```
pawiecz, line 382: Should this property be dropped entirely from the config or will it be still necessary in the future?

JenySadadia: It's just a testing commit which will be discarded later. Please check the commit message.

src/send_kcidb.py
JenySadadia: `kunit` nodes already have `test_log` file which has different content than job logs.

JenySadadia: I think that's a binary file. I need to check.

JenySadadia:
Tested OK on staging:
```
06/11/2024 05:38:45 AM UTC [INFO] Submitting node to KCIDB: 66679ab43f57810c99cb27e1
06/11/2024 05:38:45 AM UTC [ERROR] KCIDB test suite mapping not found for the test: watchdog-reset-x86-intel
06/11/2024 05:38:45 AM UTC [INFO] Not sending test as path information is missing: kernelci_api:66679ab43f57810c99cb27e1

```

config/jobs-chromeos.yaml
JenySadadia: Deleted the commit as verification is completed.

src/send_kcidb.py
JenySadadia: The editor finds issues with `test_log` file formatting when I try to open it. There are some un-readable chars when I opened it with VSCode.
Could you please check and confirm which file should I use for log excerpt?

pawiecz: As @nfraprado noted [on Slack](https://kernelci.slack.com/archives/C015N3UF7C1/p1718194687818789?thread_ts=1718186120.571499&cid=C015N3UF7C1) `test_log` would be the one to get an excerpt from because:

> [it] contains the KTAP output of the tests (including potential error messages and kernel error logs that are helpful to diagnose the error)

JenySadadia: Thanks for confirming. 
I'll update the code accordingly. I'll drop this commit from the PR as it's not relevant and open up a new one.

PR #633, opened 2024-06-10
src/send_kcidb: add error info to build nodes

JenySadadia:
Add error metadata fields such as `error_code` and `error_msg` to `misc` field for build nodes.


pawiecz (approved):
LGTM

PR #634, opened 2024-06-10
result_summary presets: add watchdog-reset presets for mainline/next

laura-nao:
Add monitor and summary presets to track the results from the watchdog reset test on the mainline and next trees.


config/result-summary.yaml
```diff
@@ -750,3 +750,61 @@ summary-v4l2-decoder-conformance-failures:
...
+    regression:
+      - data.error_code: null
+        group__re: watchdog-reset
+        repos:
+          - tree: mainline
+          - tree: next
+
+monitor-watchdog-reset-failures-mainline-next:
+  metadata:
+    action: summary
```
hardboprobot, line 772: I think this should be

```suggestion
    action: monitor
```

rather?

laura-nao: Definitely, thanks for the heads up! fixed now.

hardboprobot (approved):

PR #635, opened 2024-06-11
src/send_kcidb: get error metadata for failed/incomplete tests

JenySadadia:
Tweak condition to get error metadata for test nodes. It should get error info for incomplete nodes as well and not just failed nodes.


pawiecz (approved):
LGTM

PR #637, opened 2024-06-12
regression_tracker: add a new data field to regressions (node type)

r-c-n:
Store the type (kind) of node that caused the regression. This will come in handy when processing the output of nodes.

If later we decide we won't need this we can drop it.

### DIFF
--- a/src/regression_tracker.py
+++ b/src/regression_tracker.py
@@ -83,6 +83,7 @@ class RegressionTracker(Service):
         regression['state'] = 'done'
         error = self._collect_errors(failed_node)
         regression['data'] = {
+            'fail_node_kind': failed_node['kind'],
             'fail_node': failed_node['id'],
             'pass_node': last_pass_node['id'],
             'arch': failed_node['data'].get('arch'),


src/regression_tracker.py
```diff
@@ -83,6 +83,7 @@ def _create_regression(self, failed_node, last_pass_node):
         regression['state'] = 'done'
         error = self._collect_errors(failed_node)
         regression['data'] = {
+            'fail_node_kind': failed_node['kind'],
```
JenySadadia, line 86: Is this to distinguish if the regression is for a job node or a test node?

hardboprobot: The idea is to be able to tell which kind of node caused the regression (test, job, kbuild, etc) and have that data readily available for queries.

JenySadadia: Okay. Shouldn't this field be added [here](https://github.com/kernelci/kernelci-core/blob/main/kernelci/api/models.py#L506) first?

hardboprobot: Eventually, if we make it permanent. Actually, I don't even need to have this merged to test it (it's already been tested in staging), so if I find it's not necessary I can just drop the PR.

nuclearcat:
Do we merge this PR or we drop it?

r-c-n:
We have been living without it, and for some reason the regressions in production don't show this field, so I guess we don't need it. If someone asks for it in the future we can enable this back.

PR #638, opened 2024-06-13
src/send_kcidb: fix parsing of KUnit log file

JenySadadia:
Handle both compressed(gzip) and plain text log files for getting log excerpt.


JenySadadia:
Tested OK on KCIDB dashboard:
```
https://kcidb.kernelci.org/d/test/test?orgId=1&var-datasource=playground&var-origin=kernelci_api&var-build_architecture=All&var-build_config_name=All&var-id=kernelci_api:666ad3231ac2b8278507bdd9
```

PR #639, opened 2024-06-13
KCIDB submission: parse job nodes

JenySadadia:
Depends on https://github.com/kernelci/kernelci-pipeline/pull/628

Parse newly added job node and its child tests for KCIDB submission.


JenySadadia:
Tested OK on staging.

PR #641, opened 2024-06-13
regression_tracker: workaround: match node paths programmatically

hardboprobot:
Don't use 'path' as an api search parameter. The use of lists as query parameters (path is a list) is undefined. Instead, do the filtering in code.

**NOT TESTED YET**, waiting to see it running on staging

### DIFF
--- a/src/regression_tracker.py
+++ b/src/regression_tracker.py
@@ -107,9 +107,14 @@ class RegressionTracker(Service):
         TODO: Move this to core helpers.
 
         """
+        # Workaround: Don't use 'path' as a search parameter (we can't
+        # use lists as query parameter values). Instead, do the
+        # filtering in python code
+        path = search_params.pop('path')
         nodes = self._api.node.find(search_params)
         if not nodes:
             return None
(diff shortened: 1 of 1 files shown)


JenySadadia:
As discussed on Slack, for the time being, we can enable support for querying nodes with array elements https://www.mongodb.com/docs/manual/tutorial/query-arrays/#match-an-array

JenySadadia:
> > As discussed on Slack, for the time being, we can enable support for querying nodes with array elements https://www.mongodb.com/docs/manual/tutorial/query-arrays/#match-an-array
> 
> But that isn't as simple as flicking a switch, isn't it? That'd need changes in https://github.com/kernelci/kernelci-api/blob/16e81dccea653599f6b6c821f75d4b2b972675e2/api/db.py#L98 and other related functions, right?

That's true. I won't be straight-forward. We can consider this option or node path schema change as you had proposed earlier with '/' as a long-term solution.
We don't need to implement any of this atm if the workaround you proposed is enough for now.

nuclearcat:
Is it tested? Can we merge it?

PR #643, opened 2024-06-14
tests/validate_yaml: add validation for KCIDB mapping

JenySadadia:
To submit KernelCI generated data to KCIDB, it is required to have a mapping for all the job definitions with
`kcidb_test_suite` property.
Add validation to ensure all the jobs have a mapping present to avoid missing data submission. This check is to notify test authors trying to enable tests in maestro to include the required property for the mapping in their definition.

### DIFF
--- a/tests/validate_yaml.py
+++ b/tests/validate_yaml.py
@@ -15,7 +15,16 @@ def validate_yaml():
         if file.endswith('.yaml'):
             with open('config/' + file, 'r') as stream:
                 try:
-                    yaml.safe_load(stream)
+                    data = yaml.safe_load(stream)
+                    jobs = data.get('jobs')
+                    if not jobs:
+                        continue
+                    for name, definition in jobs.items():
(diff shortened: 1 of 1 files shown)


pawiecz (approved):
LGTM

These checks could also be added to [`kci config validate` tool](https://github.com/kernelci/kernelci-core/blob/main/kernelci/cli/config.py#L35) (which checks [section order](https://github.com/kernelci/kernelci-core/blob/e46365e03bad73b39905f6376e815bc538522555/kernelci/config/__init__.py#L64)) but that's a future task

JenySadadia:
> LGTM
> 
> These checks could also be added to [`kci config validate` tool](https://github.com/kernelci/kernelci-core/blob/main/kernelci/cli/config.py#L35) (which checks [section order](https://github.com/kernelci/kernelci-core/blob/e46365e03bad73b39905f6376e815bc538522555/kernelci/config/__init__.py#L64)) but that's a future task

Yes, that's a good idea.

PR #644, opened 2024-06-14
KCIDB submission: skip `setup` tests submission

JenySadadia:
Depends on https://github.com/kernelci/kernelci-pipeline/pull/639

`setup` test suite has been introduced to store test results for environment setup checks before running an actual test suite.
KCIDB doesn't require `setup` test suite result as long as the main test job result is submitted.


JenySadadia:
Tested OK on staging.

PR #646, opened 2024-06-17
Update KCIDB status mapping

JenySadadia:
Depends on https://github.com/kernelci/kernelci-pipeline/pull/639

Update evaluation of KCIDB status from KCI result.


JenySadadia:
Tested OK on staging.

PR #654, opened 2024-06-20
pipeline.yaml: Add kcidebug fragment

nuclearcat:
Add useful low-overhead debug option to kernel,
and test on most x86 boards we have available,
with minimal baseline tests.


nuclearcat:
Tested on staging

PR #656, opened 2024-06-21
result_summary/utils: ignore failures on log retrieval

hardboprobot:
Make the script continue running if there was an error fetching a test log.


musamaanjum (approved):
LGTM

PR #658, opened 2024-06-24
Fix links after docs page migration

pawiecz:
Documentation has been migrated to the "docs.*" subdomain.

Thanks Minas Hambardzumyan for reporting the issue!


JenySadadia (approved):
Thanks for the fix.

PR #659, opened 2024-06-26
configs: update gcc-10 to gcc-12

nuclearcat:
As we upgrade compiler images, we need update gcc version


nuclearcat:
Tested on staging
https://staging.kernelci.org:9000/viewer?search=parent%3D667bcdfae38b4895e541175e

PR #660, opened 2024-06-26
config: remove qemu jobs from lab-qualcomm

mwasilew:
QEMU jobs use container pulled from hub.docker.com. After the lab move pulling from this registry is no longer possible at Qualcomm. This patch disables QEMU jobs from Qualcomm lab.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -803,17 +803,6 @@ scheduler:
       name: lava-collabora
     platforms: *qualcomm-platforms
 
-  - job: baseline-x86-qualcomm
-    event:
-      channel: node
-      name: kbuild-gcc-12-x86
-      result: pass
-    runtime:
-      type: lava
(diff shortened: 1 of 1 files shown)


pawiecz:
@mwasilew would this still be an issue for any Docker registry or for hub.docker.com only?

Other registry could be provided as a task param [in the template](https://github.com/kernelci/kernelci-core/blob/404c850547101bebca29fffb22cabf7de07387ba/config/runtime/boot/qemu.jinja2#L23C14-L23C27).

mwasilew:
@pawiecz it's just dockerhub. I'm able to pull from AWS or github for example. I don't know how to change the template, so I propose to disable the job for now. Having a parameter would be the way to go forward for sure.

nuclearcat:
@mwasilew can you rebase PR and i will merge it?

mwasilew:
@nuclearcat done.

PR #661, opened 2024-06-27
validate_yaml.py: Improve pipeline validation

nuclearcat:
Add validation that scheduler entries have matching job entry, this is critical validation, and job entries have at least one entry in the scheduler.
Fix one entry detected by this validation

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -549,6 +549,21 @@ jobs:
 
 trees:
 
+  broonie-misc:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/broonie/misc.git"
+
+  broonie-regmap:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/broonie/regmap.git"
+
+  broonie-regulator:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/broonie/regulator.git"
+
+  broonie-sound:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/broonie/sound.git"
+
+  broonie-spi:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/broonie/spi.git"
+
   kernelci:
     url: "https://github.com/kernelci/linux.git"
 
@@ -982,6 +997,51 @@ build_variants:
 
 
 build_configs:
+
+  broonie-misc:
+    tree: broonie-misc
+    branch: 'for-kernelci'
+    variants: *build-variants
+
+  broonie-regmap:
+    tree: broonie-regmap
+    branch: 'for-next'
+    variants: *build-variants
+
+  broonie-regmap-fixes:
+    tree: broonie-regmap
+    branch: 'for-linus'
+    variants: *build-variants
+
+  broonie-regulator:
+    tree: broonie-regulator
+    branch: 'for-next'
+    variants: *build-variants
+
+  broonie-regulator-fixes:
+    tree: broonie-regulator
+    branch: 'for-linus'
+    variants: *build-variants
+
+  broonie-sound:
+    tree: broonie-sound
+    branch: 'for-next'
+    variants: *build-variants
+
+  broonie-sound-fixes:
+    tree: broonie-sound
+    branch: 'for-linus'
+    variants: *build-variants
+
+  broonie-spi:
+    tree: broonie-spi
+    branch: 'for-next'
+    variants: *build-variants
+
+  broonie-spi-fixes:
+    tree: broonie-spi
+    branch: 'for-linus'
+    variants: *build-variants
 
   kernelci_staging-mainline:
     tree: kernelci
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -893,9 +893,6 @@ scheduler:
   - job: kbuild-gcc-12-riscv-nommu_k210_defconfig
     <<: *build-k8s-all
 
-  - job: kbuild-gcc-12-riscv-rv32_defconfig
-    <<: *build-k8s-all
-
   - job: kbuild-gcc-12-arc-haps_hs_smp_defconfig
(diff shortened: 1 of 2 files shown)


tests/validate_yaml.py
```diff
@@ -7,27 +7,82 @@
...
+            if not definition.get('kcidb_test_suite'):
+                raise yaml.YAMLError(
+                    f"KCIDB test suite mapping not found for job: {name}'"
+                )
+
+def validate_scheduler_jobs(data):
+    '''
+    Each entry in scheduler have a job, that should be defined in jobs
+    '''
+    sch_entries = data.get('scheduler')
```
pawiecz, line 42: nit: `s/sch_entries/schedules/`?

btw what happened to code suggestions in the review? I'm no longer able to add ready-to-apply comments :'(

tests/validate_yaml.py
```diff
@@ -7,27 +7,82 @@
 import yaml
 import sys
 
-def validate_yaml():
+def recursive_merge(d1, d2):
```
pawiecz, line 10: We already have got a [similar procedure in `kernelci.config`](https://github.com/kernelci/kernelci-core/blob/2c249d6c9389e3c8f129afead12108421322b633/kernelci/config/__init__.py#L53).

Do you think it would be possible to refactor both these validators to use a common code base instead of separate check implementations?

config/pipeline.yaml
```diff
@@ -983,6 +998,51 @@ build_variants:
 
 build_configs:
 
+  broonie-misc:
```
pawiecz, line 1001: Monitor service will pick up changes in these trees and create events for them but they won't be processed further. 

Should this commit also include changes to `scheduler` section (adding kbuild jobs schedules)?

pawiecz (changes requested):

tests/validate_yaml.py
nuclearcat: Yes, this is why i keep it as simple functions.
But tbh i dont want to touch kernelci.config overengineered stuff or waste time trying to integrate with it.
Maybe one day we will have budget to rework it.

config/pipeline.yaml
nuclearcat: I am adding it step by step, observing possible errors and adding validator logic for such missing changes, like scheduler section.

nuclearcat: Btw there is enough default kbuild entries in scheduler that will trigger builds on any trees except android.
https://staging.kernelci.org:9000/viewer?search=parent%3D667d30c13e82bd70839e9584

tests/validate_yaml.py
pawiecz: I'm afraid time/energy saving could only work short-term in this case - long-term there will be two implementations to maintain which should also be considered as an additional cost and an error-prone conditions (especially given the fact that this is data validation code).

What I meant was that extracted data merge code from the linked `kernelci.config` module to a separate procedure could also be reused here.

config/pipeline.yaml
pawiecz: Right, all that include [`kbuild-job`](https://github.com/kernelci/kernelci-pipeline/blob/9c96cd9ef70ec0decbd6dd96e71e3510d295b5e7/config/pipeline.yaml#L32) anchor.

tests/validate_yaml.py
nuclearcat: At this particular case I'm confident i don't want to add dependency on kernelci-core. At least until we clean it up from legacy code and over-engineered parts.
If i add such dependency, it adds one big hell of dependencies and additional procedures instead of very basic verification.
That means i need to fetch kernelci-core repository, install it, with all dependencies it want, just for trivial yaml script?

If we go over proper way, such trivial functions should be made as separate library. And for example we can take a look to https://pypi.org/project/deepmerge/ which looks like does the job, instead of reinventing the wheel here and in kernelci-core. But again, we dont have time allocated for such research.

Additionally this merge function already expected be totally different from kernelci-core, as i'm planning to add more validation during merge as well, which kernelci-core likely wont have.
For example we might have job kbuild-arm64-test in pipeline.yaml, and someone by mistake might miss and add same job definition in job-chromeos.yaml. This will create very bad situation as both jobs will merge in not very predictable way, and to deal with it, i might customize further merge script to detect if there is same 2nd level definition keys have conflicts and etc.

pawiecz: I understand your point about limiting introduced dependencies, especially for simple tasks.

I don't fully understand why all these checks are implemented separately from the tool that seems to have been designed especially for this use case: [`kci config validate`](https://github.com/kernelci/kernelci-core/blob/404c850547101bebca29fffb22cabf7de07387ba/kernelci/cli/config.py#L35)

Maybe at some point these validators [will get merged](https://github.com/kernelci/kernelci-pipeline/pull/643#issuecomment-2167794407) and [run in a container](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container) including [all the necessary tools](https://hub.docker.com/r/kernelci/kernelci)

(marking this thread as resolved because the point where the discussion exceeded potential ROI has been reached)

tests/validate_yaml.py
```diff
@@ -7,27 +7,82 @@
...
+    Check if all jobs are used in scheduler
+    '''
+    sch_entries = data.get('scheduler')
+    jobs = data.get('jobs')
+    sch_jobs = [entry.get('job') for entry in sch_entries]
+    for job in jobs.keys():
+        if job not in sch_jobs:
+            print(f"Warning: Job {job} is not used in scheduler")
+
+def validate_yaml(dir='config/'):
```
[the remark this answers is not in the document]
nuclearcat: done

tests/validate_yaml.py
nuclearcat: done

PR #663, opened 2024-06-27
src/send_kcidb: replace dots in test names with underscore

JenySadadia:
KCIDB uses '.' in node path to represent hierarchy. Some tests in KCI will have '.' in its name itself such as
`AV1-TEST-VECTORS-GStreamer-AV1-V4L2SL-Gst1.0`. Replace dots with `_` in test names to avoid misinterpreting hierarchy on KCIDB side.

### DIFF
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -179,16 +179,6 @@ class KCIDBBridge(Service):
 
         return [parsed_build_node]
 
-    def _replace_restricted_chars(self, path, pattern, replace_char='_'):
-        # Replace restricted characters with "_" to match the allowed pattern
-        new_path = ""
-        for char in path:
-            if not re.match(pattern, char):
-                new_path += replace_char
-            else:
-                new_path += char
-        return new_path
-
     def _parse_node_path(self, path, is_checkout_child):
         """Parse and create KCIDB schema compatible node path
         Convert node path list to dot-separated string. Use unified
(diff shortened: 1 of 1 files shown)


src/send_kcidb.py
```diff
@@ -225,6 +225,10 @@ def _parse_node_path(self, path, is_checkout_child):
             if len(new_path) >= 2:
                 if new_path[0] == new_path[1]:
                     new_path = new_path[1:]
+
+            # Replace dots in test names with '_'
+            new_path = [path.replace('.', '_') for path in new_path]
```
hardboprobot, line 230: Isn't there a function that does something just like this but for the whole path? `_replace_restricted_chars()`, used a few lines above. If possible, path character translations should be done in one place only.

The proper solution should be:

1. Prepare all the path parts in a list (done before line 228)
2. Do all the necessary transliteration of forbidden characters in every item of the path list. **Note**: but you don't need a custom function for this.
3. Join the path with dots

So, something like `path_str = '.'.join([re.sub(restricted_chars_regex, '_', path) for path in new_path])` should do. You can do it in a loop if you think it will look clearer.

hardboprobot (changes requested):
See below for suggestions

src/send_kcidb.py
JenySadadia: Yes, I agree.
I'd move this translation to a single method.

src/send_kcidb.py
```diff
@@ -181,13 +181,7 @@ def _parse_build_node(self, origin, node):
...
     def _replace_restricted_chars(self, path, pattern, replace_char='_'):
         # Replace restricted characters with "_" to match the allowed pattern
-        new_path = ""
-        for char in path:
-            if not re.match(pattern, char):
-                new_path += replace_char
-            else:
-                new_path += char
-        return new_path
+        return re.sub(pattern, replace_char, path)
```
hardboprobot, line 184: Having this function as a wrapper for a single common library function doesn't add much, the intention is clearer if you just use the library function.

hardboprobot (changes requested):

src/send_kcidb.py
JenySadadia: Right, fixed it.
Fixed another issue related to replacing chars in KCIDB test suite name as well.

src/send_kcidb.py
```diff
@@ -225,10 +216,15 @@ def _parse_node_path(self, path, is_checkout_child):
             if len(new_path) >= 2:
                 if new_path[0] == new_path[1]:
                     new_path = new_path[1:]
+
+            # Allowed pattern for test name is ^[a-zA-Z0-9_-]*$'
+            # Replace restricted characters with '_'
+            # KCIDB test suite name should be kept as-is
+            new_path = [re.sub(r'[^a-zA-Z0-9_-]', "_", path)
```
hardboprobot, line 223: Easier and clearer: rather than making an exception for `kcidb_suite_name` here, do the character translation only where it's needed. That's in the `else` branch in line 221:

``` python
for sub_path in parsed_path:
    if sub_path in self._jobs:
        if self._jobs[sub_path].kcidb_test_suite:
            new_path.append(self._jobs[sub_path].kcidb_test_suite)
        else:
            self.log.error("KCIDB test suite mapping not found for "
                           f"the test: {sub_path}")
            return None
    else:
        ##### CHARACTER SUBSTITUTION HERE #####
        new_path.append(sub_path)
```

hardboprobot (changes requested):

src/send_kcidb.py
JenySadadia: Thanks.
Done.

hardboprobot (approved):

JenySadadia:
Based on the discussion last week, @laura-nao is going to fix the test names issue in the fluster test parser.
That would be a better solution than handling such cases on the KCIDB bridge side.
Hence, closing this PR.

PR #667, opened 2024-07-02
Enable Android builds in different configurations with gcc-12 and clang-17 both

musamaanjum:
Android UM builds are already enabled. This PR enables android builds for arm, arm64, x86, i386, riscv with gcc-12 and clang-17.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -1620,6 +1620,10 @@ build_configs:
     tree: android
     branch: 'android15-6.6'
 
+  android15-6.6-lts:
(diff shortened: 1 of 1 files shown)


musamaanjum:
Confirmed results on arm and arm64. I've not found results for other architectures yet.
[Uploading kbuilds.html.txt…]()

There was typo in config name. That has been fixed.

nuclearcat:
Looks like working: https://staging.kernelci.org:9000/viewer?search=parent%3D6686cc205dfebcce60dc316b

PR #668, opened 2024-07-03
aks: Add kubernetes kcidb deployment

nuclearcat:
We need file that will manage deployment of kcidb bridge in kubernetes production deployment.


nuclearcat:
Tested on production

PR #670, opened 2024-07-03
config: chromeos: Add 'nfs' suffix to KCIDB suite name for baseline-nfs

laura-nao:
The baseline test is currently run with both ramdisk and nfs rootfs. To distinguish baseline-nfs tests in KCIDB, add an 'nfs' suffix to the KCIDB test suite name.


JenySadadia (approved):
Tested OK on staging:
https://kcidb.kernelci.org/d/test/test?orgId=1&var-datasource=playground&var-origin=maestro&var-build_architecture=All&var-build_config_name=All&var-id=maestro:668645655e5e08b0d58111e9

PR #674, opened 2024-07-04
regression_tracker: bugfix: catch empty search condition

hardboprobot:
Fix _get_last_matching_node(), after the previous change there was an unhandled scenario where nodes may be empty but the function wouldn't return None immediately.


JenySadadia (approved):
LGTM.

PR #676, opened 2024-07-04
scheduler-chromeos.yaml: Temporarily disable non-essential tast tests

nuclearcat:
As per discussion, we disable temporary tast tests which unlikely will be reviewed.


nuclearcat:
As per staging works as intended.

PR #679, opened 2024-07-05
docs: Add setup.md initial version

nuclearcat:
I am adding initial version to document important environment variables that are not documented anywhere.
This is initial version only, that will be extended in near future.

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -1,0 +1,63 @@
+---
+title: "Pipeline setup"
+date: 2024-07-05
+description: "KernelCI Pipeline self-hosted setup"
+weight: 2
+---
+
+GitHub repository:
+[`kernelci-pipeline`](https://github.com/kernelci/kernelci-pipeline.git)
+
+The KernelCI Pipeline is a set of tools and scripts to automate the
+testing of Linux kernels on real hardware. In some cases, like
+development, or running your own compiling farm and lab, you may want
+to run the KernelCI Pipeline on your own infrastructure. This document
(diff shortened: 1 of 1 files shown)


JenySadadia (commented):
I think this should go to https://docs.kernelci.org/api_pipeline/api/local-instance/#setting-up-a-pipeline-instance.

nuclearcat:
I think we will rather move part of that doc to pipeline, because documentation about each repository must belong to it.
If there is any documentation file that describe multiple repositories, it must be in kernelci-project.
It is counterintuitive to search documentation about how to set up pipeline instance - in api instance.
I suggest in that manual, we will put link to next chapter, so user click and go to pipeline instance setup, separate documentation.
We have to think how to rearrange everything.

JenySadadia:
> I think we will rather move part of that doc to pipeline, because documentation about each repository must belong to it. If there is any documentation file that describe multiple repositories, it must be in kernelci-project. It is counterintuitive to search documentation about how to set up pipeline instance - in api instance. I suggest in that manual, we will put link to next chapter, so user click and go to pipeline instance setup, separate documentation. We have to think how to rearrange everything.

I think the idea was to centralize meastro's documentation and keep docs inside the repo as minimal as possible.
That's why we reduced content of `README.md` file for both the repo and moved everything possible to kernelci.org.
As we already have docs for setting up pipeline instance, it makes sense to me to move all the related docs to that location.

nuclearcat:
Will be moved to API instance docs, but i believe it is wrong place to put docs how to setup pipeline instance.

PR #681, opened 2024-07-05
callback_url: Update config and README

nuclearcat:
As we are moving callback URL to environment variable, updating config and README accordingly.


nuclearcat:
Tested on staging

PR #684, opened 2024-07-08
scheduler.py: If event have jobfilter, inject it to the node data

nuclearcat:
When someone generate artificial event with jobfilter, this is likely maintainer trying to repeat job. Treat this accordingly, and inject job filter to job node, so we will run only tests maintainer wants.

### DIFF
--- a/src/scheduler.py
+++ b/src/scheduler.py
@@ -189,6 +189,10 @@ class Scheduler(Service):
             event = self._api_helper.receive_event_data(sub_id)
             for job, runtime, platform, rules in self._sched.get_schedule(event):
                 input_node = self._api.node.get(event['id'])
+                jobfilter = event.get('jobfilter')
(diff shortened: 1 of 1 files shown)


JenySadadia (commented):
A minor type in the commit message:

"When someone generate artificial event with jobfilter, this is
likely maintainer trying to repeat job. **Treat** this accordingly,
and inject job filter to job node, so we will run only tests
maintainer wants."

nuclearcat:
> A minor type in the commit message:
> 
> "When someone generate artificial event with jobfilter, this is likely maintainer trying to repeat job. **Treat** this accordingly, and inject job filter to job node, so we will run only tests maintainer wants."

Oops, thanks, seems seems i have too many threats, hehe.

nuclearcat:
Tested on staging manually, by submitting custom event

PR #685, opened 2024-07-09
lava_callback: migrate to fastapi

nuclearcat:
It will be easier to maintain API and Pipeline, as both will be powered by FastAPI framework.


nuclearcat:
Works on staging

PR #686, opened 2024-07-09
doc: developer-documentation: Update documentation by adding more details

musamaanjum:
- Reorganize some things
    - Specify how to write different variants by removing old syntax
    - Give two separate templates for kbuild and test
    - Try to put more details for new contributors


doc/developer-documentation.md
```diff
@@ -7,8 +7,8 @@ weight: 4
 
 ## Enabling a new Kernel tree
 
-We can monitor different kernel trees in KernelCI.
-This manual describes how to enable them in [`kernelci-pipeline`](https://github.com/kernelci/kernelci-pipeline.git).
+We can monitor different kernel trees in KernelCI. The builds and test jobs are triggered whenever the spcified branches are updated.
```
nuclearcat, line 10: specified

doc/developer-documentation.md
```diff
@@ -17,105 +17,147 @@ In order to enable a new tree, we need to add an entry there.
...
+      defconfig: <defconfig_name>
+      fragments:
+       - <fragment_name>
+    rules:
+      min_version:
+        version: <kernel_version>
+        patchlevel: <kernel_patch_level>
+      tree:
+        - <tree_name>
+        - <!tree_name>
```
JenySadadia, line 55: I think it should be the below instead
```suggestion
        - !<tree_name>
```
Also, I don't think the same `<tree-name>` should be there for both the rules.

doc/developer-documentation.md
```diff
@@ -17,105 +17,147 @@ In order to enable a new tree, we need to add an entry there.
...
+      collections: <name_of_suite>
+      job_timeout: <timeout>
     kcidb_test_suite: <kcidb-mapping>
+    rules:
+      min_version:
+        version: <kernel_version>
+        patchlevel: <kernel_patch_level>
+      tree:
+        - <tree_name>
+        - <!tree_name>
```
JenySadadia, line 71: The same comment as above.

doc/developer-documentation.md
```diff
@@ -17,105 +17,147 @@ In order to enable a new tree, we need to add an entry there.
...
+  <name_of_variant1>:
+    tree: <tree_name>
+    branch: <branch_name1>
+```
+That's it! The tree is enabled now. All the jobs defined under `jobs` section of [config file](https://github.com/kernelci/kernelci-pipeline/blob/main/config/pipeline.yaml) would run on the specified branched for this tree.
 
 ### Schedule the job
 
-We also need a `scheduler` entry for the newly added job to specify pre-conditions for scheduling and define runtime and platforms for job submissions.
+We also need a `scheduler` entry for the newly added job to specify pre-conditions for scheduling, define runtime and platforms for job submissions.
```
JenySadadia, line 123:
```suggestion
We also need a `scheduler` entry for the newly added job to specify pre-conditions for scheduling, and defining runtime and platforms for job submissions.
```

doc/developer-documentation.md
```diff
@@ -17,105 +17,147 @@ In order to enable a new tree, we need to add an entry there.
...
+For a test that requires a successful completion of a build job such as `kbuild-gcc-10-arm64`, specify the event as follows:
+```yaml
+  event:
     channel: node
     name: kbuild-gcc-10-arm64
     result: pass
 ```
-- Choose a `runtime` for scheduling and running the job. Currently supported runtimes are shell, docker, lava, and kubernetes. Select a runtime type from `runtimes` section. Please note that `name` property is required for `lava` and `k8s` runtimes to specify which lab/k8s context should run the test as number of LAVA labs such as BayLibre, Collabora, Qualcomm and k8s contexts have been enabled in KernelCI.
-- `platforms` will include a list of device types on which the test should run. It should match entries defined in `platforms` section such as `qemu-x86`, `bcm2711-rpi-4-b`, and so on.
+Here, `node` refers to the database.
```
JenySadadia, line 159:
```suggestion
Here, `node` refers to the name of API PubSub channel where node events are published.
```

JenySadadia (changes requested):
A few suggestions to look at.

doc/developer-documentation.md
```diff
@@ -17,105 +17,147 @@ In order to enable a new tree, we need to add an entry there.
...
-    kernelci:
-        url: "https://github.com/kernelci/linux.git"
+  kernelci:
+    url: "https://github.com/kernelci/linux.git"
 ```
 
-After adding a `trees` entry, we need to define build configurations for it.
-In the same [configuration](https://github.com/kernelci/kernelci-pipeline/blob/main/config/pipeline.yaml) file, `build_configs` section is there to specify them.
-For example, we need to specify which branch to monitor of a particular tree and other build variants as well.
+This `tree-name` is used to refer to this tree from here onwards.
```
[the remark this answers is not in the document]
musamaanjum: I don't have solid stance on keeping it. But it does clarify a bit.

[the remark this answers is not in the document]
JenySadadia: Okay, if you want to keep it, it should be rephrased a bit.
For example, something like "We would use `<tree-name>` in the below sections to refer to the newly added tree." would make it more clear.

doc/developer-documentation.md
JenySadadia: The above suggestion has not been addressed yet to exclude `!` from `<tree-name>`.

JenySadadia: Also, it should be explained somewhere what `!` mark means in `rules`.

doc/developer-documentation.md
```diff
@@ -17,105 +17,147 @@ In order to enable a new tree, we need to add an entry there.
...
+      collections: <name_of_suite>
+      job_timeout: <timeout>
     kcidb_test_suite: <kcidb-mapping>
+    rules:
+      min_version:
+        version: <kernel_version>
+        patchlevel: <kernel_patch_level>
+      tree:
+        - <tree_name>
+        - <!tree_name2>
```
JenySadadia, line 71: In this example and other ones as well, some property markers use `-`(e.g. `<kbuild-job-name>`) whereas others use `_`(e.g. `<tree_name>`).
Please unify a pattern in the whole file.

doc/developer-documentation.md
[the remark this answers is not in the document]
musamaanjum: Sure.

doc/developer-documentation.md
musamaanjum: Unified now.

doc/developer-documentation.md
musamaanjum: I'd updated to 
```
      tree:
        - <tree_name>
        - <!tree_name2>
```

I want to give an example of how we can include a tree `<tree_name>` and ignore a tree `<!tree_name2>`. Technically is it correct that we can specify two trees like this? What do you suggest?

doc/developer-documentation.md
```diff
@@ -17,105 +17,147 @@ In order to enable a new tree, we need to add an entry there.
...
```
JenySadadia, line 71:
```suggestion
        - <tree_name>
        - !<tree_name2>
```

doc/developer-documentation.md
JenySadadia: Yes, you can definitely include two tree names with different rules. I am suggesting to keep `!` outside of `<tree-name2>` marker.
Please take a look at https://github.com/kernelci/kernelci-pipeline/pull/686#discussion_r1682244359 for suggested change.

JenySadadia:
@nuclearcat Can we deploy this to staging to verify?

doc/developer-documentation.md
```diff
@@ -17,105 +17,147 @@ In order to enable a new tree, we need to add an entry there.
...
-    kernelci:
-        url: "https://github.com/kernelci/linux.git"
+  kernelci:
+    url: "https://github.com/kernelci/linux.git"
 ```
 
-After adding a `trees` entry, we need to define build configurations for it.
-In the same [configuration](https://github.com/kernelci/kernelci-pipeline/blob/main/config/pipeline.yaml) file, `build_configs` section is there to specify them.
-For example, we need to specify which branch to monitor of a particular tree and other build variants as well.
+The <tree_name> will be used in the other sections to refer to the newly added tree.
```
JenySadadia, line 30:
```suggestion
The `<tree_name>` will be used in the other sections to refer to the newly added tree.
```

doc/developer-documentation.md
JenySadadia: This doesn't look like it's fixed yet. Maybe you haven't committed the change yet?

doc/developer-documentation.md
musamaanjum: Now I've realized that ! should be outside <>. Thanks for mentioning it again. Fixing.

doc/developer-documentation.md
musamaanjum: Updated now.

doc/developer-documentation.md
musamaanjum: Done

doc/developer-documentation.md
musamaanjum: Sure. Updated.

doc/developer-documentation.md
```diff
@@ -17,105 +17,147 @@ In order to enable a new tree, we need to add an entry there.
...
+    kind: <kind of job, either 'job' or 'test'>
+    params:
+      nfsroot: <rootfs_url>
+      collections: <name_of_suite>
+      job_timeout: <timeout>
     kcidb_test_suite: <kcidb-mapping>
+    rules:
+      min_version:
+        version: <kernel_version>
+        patchlevel: <kernel_patch_level>
```
JenySadadia, line 68:
```suggestion
    params:
      arch: <architecture>
      compiler: <compiler-name>
      cross_compile: <compiler-prefix>
      dtbs_check: <dtb-check-enabled>
      defconfig: <defconfig-name>
      fragments:
       - <fragment-name>
    rules:
      min_version:
        version: <kernel-version>
        patchlevel: <kernel-patch_level>
      tree:
        - <tree-name>
        - !<tree-name2>

  <test-job-name>:
    template: <template-name>
    kind: <kind of job, either 'job' or 'test'>
    params:
      nfsroot: <rootfs-url>
      collections: <name-of-suite>
      job_timeout: <timeout>
    kcidb_test_suite: <kcidb-mapping>
    rules:
      min_version:
        version: <kernel-version>
        patchlevel: <kernel-patch-level>
```

doc/developer-documentation.md
JenySadadia: It's not fixed yet. See my recent 2 comments.

musamaanjum: I'd fixed only trees. Thank you for suggesting. 

I'll take extra care in naming things in yaml from now on.

doc/developer-documentation.md
```diff
@@ -17,105 +17,146 @@ In order to enable a new tree, we need to add an entry there.
...
-- `kcidb_test_suite` is a field used to map KernelCI test suite name with
-KCIDB test. Build jobs (with `kind: kbuild`) do not require this field.
-
+- **`template`**: A `jinja2` template should be added to the [`config/runtime`](https://github.com/kernelci/kernelci-pipeline/tree/main/config/runtime) directory. This template will be used to generate the test definition.
+- **`kind`**: The `kind` field specifies the type of job. It should be `kbuild` for build jobs, `job` for a test suite, and `test` for a single test case.
+- **`image`**: The `image` field specifies the Docker image used for building and running the test. This field is optional. For example, LAVA test jobs use an image defined in the test definition template instead.
+- **`params`**: The `params` field includes parameters for building the kernel (for `kbuild` jobs) or running the test. These parameters can include architecture, compiler, defconfig options, job timeout, etc.
+- **`rules`**: The `rules` field defines job rules. If a test should be scheduled for a specific kernel tree, branch, or version, these rules can be specified here.
```
JenySadadia, line 78: One last thing.
Please add a line here describing what rules mean with and without `!`.

JenySadadia: Sorry, but it seems a bit unclear to me.
Maybe something like "The rules with `!` will exclude that specific condition for job scheduling e.g. in the above example, the scheduler won't schedule a job if the event is received for Kernel tree `tree-name2`"?
Feel free to rephrase it if required.

musamaanjum: Changed:
"The rules prefixed with `!` exclude the specified condition from job scheduling. For example, in the given scenario, the scheduler does not schedule a job if an event is received for the kernel tree `tree-name2`."

PR #687, opened 2024-07-10
config: pipeline: fix defconfigs in fragments

musamaanjum:


musamaanjum:
This PR will be tested with https://github.com/kernelci/kernelci-core/pull/2608 to find out if current method would work fine.

PR #688, opened 2024-07-10
kbuild.jinja2: support defconfig as list or str

nuclearcat:
As required in https://github.com/kernelci/kernelci-core/pull/2608 defconfig might be two types. Support it in jinja2 accordingly.


nuclearcat:
Tested on staging, for example: https://staging.kernelci.org:9000/viewer?node_id=668ef28b4a3f5a2f49659614
on build.log: "make x86_64_defconfig allmodconfig"

PR #690, opened 2024-07-12
k8s: Update deployment to uvicorn, as we use fastapi now

nuclearcat:


nuclearcat:
Tested on production, as a fix for current deployment issue

PR #691, opened 2024-07-12
lava_callback: Add API call to pipeline

nuclearcat:
Add call to create custom checkout node (manually specified commit).

### DIFF
--- a/src/lava_callback.py
+++ b/src/lava_callback.py
@@ -1,7 +1,8 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 #
-# Copyright (C) 2023 Collabora Limited
+# Copyright (C) 2024 Collabora Limited
 # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
+# Author: Denys Fedoryshchenko <denys.fedoryshchenko@collabora.com>
 
 import os
 import tempfile
@@ -12,12 +13,16 @@ import requests
 import toml
 import threading
 import uvicorn
-from fastapi import FastAPI, HTTPException, Request
-
+import jwt
+import logging
+from datetime import datetime, timedelta
+from fastapi import FastAPI, HTTPException, Request, Header
+from pydantic import BaseModel
 import kernelci.api.helper
 import kernelci.config
 import kernelci.runtime.lava
 import kernelci.storage
+import kernelci.config
 from concurrent.futures import ThreadPoolExecutor
 
 
@@ -26,9 +31,19 @@ CONFIGS = kernelci.config.load(
     SETTINGS.get('DEFAULT', {}).get('yaml_config', 'config')
 )
 SETTINGS_PREFIX = 'runtime'
+YAMLCFG = kernelci.config.load_yaml('config')
 
 app = FastAPI()
 executor = ThreadPoolExecutor(max_workers=16)
+
+
+class ManualCheckout(BaseModel):
+    nodeid: str
+    commit: str
+
+
+class JobRetry(BaseModel):
+    nodeid: str
 
 
 def _get_api_helper(api_config_name, api_token):
@@ -112,7 +127,7 @@ def async_job_submit(api_helper, node_id, job_callback):
     results = job_callback.get_results()
     job_node = api_helper.api.node.get(node_id)
     if not job_node:
-        print(f'Node {node_id} not found')
+        logging.error(f'Node {node_id} not found')
         return
     # TODO: Verify lab_name matches job node lab name
     # Also extract job_id and compare with node job_id (future)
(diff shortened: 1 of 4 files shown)


JenySadadia:
I have a bunch of questions for this implementation:
- Probably, I don't understand things correctly, but why do we need to introduce endpoints in pipeline rather than API server? That way we don't need to repeat certain checks like providing tokens and authenticating a user.
We can introduce `retry_job` endpoint in API and publish events on `node` channel with some kind of flag `retry: 1` to notify pipeline to schedule a retry. Or introduce a new PubSub channel e.g. `retry` to publish retry-specific events with required data.
And a pipeline service would be listening to that channel and schedule a retry accordingly.
- Why can't we introduce `kci-maintainer` tool in `kernelci-core` with other KCI commands?

nuclearcat:
1) API user management is intended to provide namespaces for different pipeline instances; it is not intended to share a single pipeline instance. 
I can assume on an abstraction layer it is a sort of custom ODBC-alike interface to MongoDB and Redis.
If we start implementing multiple users as one user group, we will have to implement granular access control and change the whole system, which will require much more effort.
An additional reason I prefer to avoid giving API access - the way fastapi-users is implemented - is that it's very inconvenient for users. There are too many confusing steps; a good example is Aritsu - you witnessed her hardships, and I had the same feedback from other users.
On Pipeline-API, we issue a simple JWT token for users, instantly ready to use, with trivial granular access control. If the situation requires, we can implement expiration, revocation, etc.
Technically, we provide them access to our specific pipeline instance, specific functions, such as compiling and test farm, not to the whole system.

2) At the current moment, the design approach is based on verbal feedback from @aritsu, that kernelci-core tools have too many dependencies, which often require Docker to run them, and it is not convenient for users.
The current tool is designed with minimum dependencies; I think it can run as-is on bare Debian and might be moved to a new repository kernelci-maestro (@padovan), with all other new generation tools that have minimum dependencies.
We might design a lightweight new kernelci library for them later to unify certain aspects, but definitely without overcomplicating things and dragging in every possible shiny library and lot of legacy as it is done in kernelci-core.

3) Regarding the retry entry, I have a proposal to add the following field:
job kind> retry: null or nodeid. If this is a retry job, then we specify the jobid that initiated this retry.
This will allow us to track retry chains.
Custom checkout cannot really be classified as a retry, but we might set some flag as well to track it, if needed.

tools/kci-maintainer
```diff
@@ -0,0 +1,96 @@
...
+
+import requests
+import json
+import os
+import argparse
+
+API_ADDR = ['https://staging.kernelci.org:9100/',
+            'https://kernelci-pipeline.westus3.cloudapp.azure.com/']
+API_DESC = ['staging', 'production']
+API_URL = API_ADDR[0]
```
pawiecz, line 20:
```suggestion
APIS = {
    'staging': 'https://staging.kernelci.org:9100/',
    'production': 'https://kernelci-pipeline.westus3.cloudapp.azure.com/',
}
DEFAULT_API = 'staging'
API_URL = APIS.get(DEFAULT_API)
```

tools/kci-maintainer
```diff
@@ -0,0 +1,96 @@
...
+        print('API token is required')
+        return
+
+    if args.jobretry:
+        if not args.nodeid:
+            print('Node ID is required')
+            return
+        send_jobretry_request(token, args.nodeid)
+
+    if args.checkout:
```
pawiecz, line 85: Is there a use case where `./kci-maintainer --jobretry -n foo --checkout -c bar` make sense?

If not these flags should be made mutually exclusive.

src/lava_callback.py
```diff
@@ -187,9 +202,194 @@ async def callback(node_id: str, request: Request):
...
+        return None
+    parent_node = api_helper.api.node.get(parent_id)
+    if not parent_node:
+        return None
+    if parent_node.get('kind') == kind:
+        return parent_node
+    return find_parent_kind(parent_node, api_helper, kind)
+
+
+@app.post('/api/jobretry')
```
pawiecz, line 251: This is a second set of APIs implemented in a pipeline service, bypassing the main API.

Does it do its job? It most likely works
Would I support such a shift in a solution architecture? It most likely works

src/lava_callback.py
```diff
@@ -187,9 +202,194 @@ async def callback(node_id: str, request: Request):
...
+    knode.pop('usergroups', None)
+
+    evnode = {'data': knode}
+    # Now we can submit custom kbuild node to the API(pub/sub)
+    api_helper.api.send_event('node', evnode)
+    logging.info(f"Job retry for node {data.nodeid} submitted")
+    return 'OK', 200
+
+
+@app.post('/api/checkout')
```
pawiecz, line 305: This additional API endpoint benefits only from being run with the pipeline service privileges - other than that it could be a `kci` tool subcommand.

I get its use case but I don't believe it solves the real issue here - Node permissions and access control, they will have to be addressed sooner or later.

pawiecz (commented):
I mark these as comments because I don't believe this input could spark any changes anymore

tools/kci-maintainer
```diff
@@ -0,0 +1,96 @@
...
+
+
+def main():
+    global API_URL
+    ap = argparse.ArgumentParser()
+    token = get_token()
+    if not token:
+        ap.add_argument('-t', '--token', required=True, help='API token')
+
+    ap.add_argument('--api', type=int, choices=[0, 1], default=0,
```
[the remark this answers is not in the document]
nuclearcat: I changed it a bit different way, but using dict as you suggested

tools/kci-maintainer
nuclearcat: done

tools/jwt_generator.py
```diff
@@ -0,0 +1,67 @@
...
+    args.permissions = args.permissions.replace(" ", "")
+
+    # Basic email validation
+    if '@' not in args.email:
+        print("Invalid email")
+        sys.exit(1)
+
+    # if comma separated permissions are provided, split them into a list
+    permissions = args.permissions.split(',')
+    jsdict = {}
```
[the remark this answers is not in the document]
nuclearcat: done

JenySadadia:
> 1. API user management is intended to provide namespaces for different pipeline instances; it is not intended to share a single pipeline instance.
>    I can assume on an abstraction layer it is a sort of custom ODBC-alike interface to MongoDB and Redis.
>    If we start implementing multiple users as one user group, we will have to implement granular access control and change the whole system, which will require much more effort.

I am sorry, but I still don't understand why to re-implement the same thing in pipeline if we already have user management in place in API.
The idea was to implement permissions, access, and user groups in API itself. I think it's worth the implement those things in API rather than in pipeline.
That will serve the purpose.

>    An additional reason I prefer to avoid giving API access - the way fastapi-users is implemented - is that it's very inconvenient for users. There are too many confusing steps; a good example is Aritsu - you witnessed her hardships, and I had the same feedback from other users.

Could you please elaborate what was the feedback from user management PoV?

>    Technically, we provide them access to our specific pipeline instance, specific functions, such as compiling and test farm, not to the whole system.

The same thing is possible by giving permissions to a group of users to a certain API endpoint.

> 2. At the current moment, the design approach is based on verbal feedback from @aritsu, that kernelci-core tools have too many dependencies, which often require Docker to run them, and it is not convenient for users.
>    The current tool is designed with minimum dependencies; I think it can run as-is on bare Debian and might be moved to a new repository kernelci-maestro (@padovan), with all other new generation tools that have minimum dependencies.

It looks like we are re-writing almost everything instead of improving and stablising the tools already developed for the new system.

nuclearcat:
> I am sorry, but I still don't understand why to re-implement the same thing in pipeline if we already have user management in place in API.
> The idea was to implement permissions, access, and user groups in API itself. I think it's worth the implement those things in API rather than in pipeline.
> That will serve the purpose.

API user management is providing separate namespace for multiple pipeline instances. This is how it is working now and this is how it is designed. If another user create node, owner of node will be that user, and our pipeline instance will ignore it.
I'm fine with it, how it works now, but for now.

I have significant concerns about modifying the API user management model at this stage:

While we certainly have the capability to implement various features like permissions and access groups, I believe it's not ideal to introduce major changes at this point. We're approaching a critical deadline where we need to demonstrate a relatively stable and functional system.

Implementing user groups and their proper integration with existing schema would require substantial time and effort. We've been discussing retry functionality for the past 1-2 months, and I've had to quickly implement a working solution to provide this feature for testing as soon as possible.

Given our current timeline, I'm unsure if anybody want to take risk and can successfully implement and thoroughly test these changes within a week. 
Perhaps we could discuss alternative approaches or a phased implementation plan that aligns better with our immediate goals and deadline.

Furthermore, altering the API user management model and introducing a new approach to node ownership at this stage carries some risks. These changes could potentially impact our existing functionality. Given our upcoming deadline, it might be prudent to consider whether making such significant changes is the best course of action. There's a possibility that it could affect even our basic features, which might not be ideal for the project's reputation at this critical juncture. I'm concerned about taking on this level of risk so close to our delivery date.

> Could you please elaborate what was the feedback from user management PoV?
It was conversation with Arisu few month ago, when we tried to get early instance working. Unfortunately Slack don't keep history long enough. You can ask Arisu for feedback, if there is still memories about that moment.
Another piece of feedback I received, which I've written as notes:
A) Why do we need email verification? Such measures are typically required for public services when users supply their email addresses themselves, and we need to verify if they are real. In our case, we are creating users through manual requests and already sending them emails with credentials. If their email is invalid, they won't receive this email.
B) We currently allow only administrators to reset passwords, which doesn't make sense. We should consider two options:

* Administrators handle everything for the user (including sending a new temporary password), OR
* Users should be able to perform full "self-service," e.g., request a password reset token themselves.

This revised version maintains the core content of your notes while improving readability and coherence. Let me know if you'd like any further adjustments or explanations.

> The same thing is possible by giving permissions to a group of users to a certain API endpoint.
Do we have any group of users functionality implemented? Do we have even any design drafts for it?

> It looks like we are re-writing almost everything instead of improving and stablising the tools already developed for the new system.
Sometimes a fresh start is necessary to align with new leadership's vision and current organizational goals. Rebuilding can address fundamental issues that may have hindered previous efforts. This approach can reinvigorate project and lead to more maintainable, efficient solutions tailored to present needs.

.github/workflows/main.yml
```diff
@@ -48,4 +48,24 @@ jobs:
       - name: Run basic yaml validation
         run: |
           python tests/validate_yaml.py
-        
+  on-fail:
```
pawiecz, line 51: I believe this change belongs to https://github.com/kernelci/kernelci-pipeline/pull/712 which was already reverted in https://github.com/kernelci/kernelci-pipeline/pull/725

nuclearcat: Removed, got them here by mistake

nuclearcat:
Merging as it is, development will continue

.github/workflows/main.yml
pawiecz: Removal of this workflow change also reverted all the other changes applied during review.

I choose to believe it might have been done by mistake. If that's not the case, it's a next-level review trolling :laughing:

pawiecz: Resolved in: https://github.com/kernelci/kernelci-pipeline/pull/734

PR #693, opened 2024-07-15
config: pipeline: Unblock android runs on lava-collabora

musamaanjum:


musamaanjum:
Verified that now we are getting the test runs for the android tree. This PR can be merged now.

PR #694, opened 2024-07-16
config: platforms-chromeos: Add serial delay for Jacuzzi

laura-nao:
Add `test_character_delay` to Jacuzzi Chromebooks, similar to other Mediatek Chromebooks. This workaround addresses the issue of not processing serial input quickly enough, which may cause output mangling and inconsistent test results.

This is a temporary workaround to identify the correct delay value. The proper place to define the delay is the LAVA device dictionaries.

Similar to: https://github.com/kernelci/kernelci-pipeline/pull/626

Instance of test failure caused by corrupt serial output, causing a false regression on this platform (`ttyS ttyS0: 1 input overrun(s)` is also reported in the log): https://lava.collabora.dev/scheduler/job/14838004#L6255

### DIFF
--- a/config/platforms-chromeos.yaml
+++ b/config/platforms-chromeos.yaml
@@ -103,6 +103,8 @@ platforms:
     base_name: jacuzzi
     mach: mediatek
     dtb: dtbs/mediatek/mt8183-kukui-jacuzzi-juniper-sku16.dtb
+    context:
+      test_character_delay: 10
     rules:
       <<: *arm64-chromebook-device-rules
       min_version:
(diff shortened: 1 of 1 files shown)


config/platforms-chromeos.yaml
```diff
@@ -103,6 +103,8 @@ platforms:
     base_name: jacuzzi
     mach: mediatek
     dtb: dtbs/mediatek/mt8183-kukui-jacuzzi-juniper-sku16.dtb
+    context:
+      test_character_delay: 10
```
a-wai, line 107: So this means this parameter is now added for all MTK chromebooks, right? In that case, can you please drop those lines for other ones? Setting it for jacuzzi means it will be part of `mediatek-chromebook-device`, and therefore picked up for other devices based on this definition.

a-wai (changes requested):

config/platforms-chromeos.yaml
laura-nao: Done - thanks for the heads up!

a-wai (approved):

PR #697, opened 2024-07-16
config: use the new `tree:branch` format for rules

a-wai:
For cases where we want a single branch to be allowed for a given tree, we can now use the `tree:branch` format in rules. Convert existing rules accordingly.

Depends on https://github.com/kernelci/kernelci-core/pull/2613


nuclearcat:
Merged as related and tested with https://github.com/kernelci/kernelci-core/pull/2613

PR #698, opened 2024-07-16
config: pipeline: fix improper use of "filters" attribute

a-wai:
The `filters` param was used in the legacy system but has been replaced by `rules`, with a different syntax.

For Android RISC-V builds, this was used to deny job execution on kernels < 4.19, so let's translate this condition with the rules format, and do a similar change for the `rt-tests`-based jobs.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -687,10 +687,10 @@ jobs:
       defconfig:
         - defconfig
         - allnoconfig
-      filters:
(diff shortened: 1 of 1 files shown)


musamaanjum (approved):

nuclearcat:
Merged as related and tested with https://github.com/kernelci/kernelci-core/pull/2613

PR #701, opened 2024-07-17
config: pipeline: Remove android's deprecated branches

musamaanjum:
It has been confirmed with Todd that we should remove the deprecated branches. Hence remove those branches.


nuclearcat (approved):
LGTM

Issue #650, opened 2024-06-18
Document KernelCI test hierarchy and result semantics


JenySadadia:
As a part of addressing issues mentioned in https://github.com/kernelci/kernelci-core/issues/2455, I have restructured the test hierarchy and results semantics in the maestro.

Document the details about how KernelCI is storing results and interpreting test hierarchy with the new approach.

- Add documentation to [Developer doc](https://github.com/kernelci/kernelci-pipeline/blob/main/doc/developer-documentation.md)

PR #703, opened 2024-07-18
doc/developer-documentation: test hierarchy and result evaluation

JenySadadia:
Closes https://github.com/kernelci/kernelci-pipeline/issues/650

Explain KernelCI test hierarchy structure and results semantics.


pawiecz (approved):
LGTM - I wish I had this type of support (these docs) when I first started Maestro journey

Final flowchart could use some tuning (e.g. condition block don't need `if`s in them - their shape already implies that) but it can be adjusted in future patches

JenySadadia:
> LGTM - I wish I had this type of support (these docs) when I first started Maestro journey
> 
> Final flowchart could use some tuning (e.g. condition block don't need `if`s in them - their shape already implies that) but it can be adjusted in future patches

That's true. I'll update the diagram.

PR #706, opened 2024-07-19
Enable builds of arnd, amlogic and ardb trees

musamaanjum:
(Added description to individual commmits)

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -111,6 +111,10 @@ _anchors:
     - sc7180-trogdor-kingoftown
     - sc7180-trogdor-lazor-limozeen
 
+  build-only-trees: &build-only-trees-rules
+    rules:
+      tree:
+      - 'arnd'
 
 api:
 
@@ -271,7 +275,7 @@ jobs:
   baseline-arm64-kcidebug-qualcomm: *baseline-job
   baseline-x86-mfd: *baseline-job
 
-  kbuild-gcc-12-arc-haps_hs_smp_defconfig:
+  kbuild-gcc-12-arc-haps_hs_smp_defconfig: &kbuild-gcc-12-arc-job
     <<: *kbuild-job
     image: kernelci/staging-gcc-12:arc-kselftest-kernelci
     params:
@@ -283,6 +287,11 @@ jobs:
       tree:
       - 'stable-rc'
       - 'kernelci'
+
+  # Default config and build only job
+  kbuild-gcc-12-arc-build-only:
+    <<: *kbuild-gcc-12-arc-job
+    rules: *build-only-trees-rules
 
   kbuild-clang-17-arm: &kbuild-clang-17-arm-job
     <<: *kbuild-job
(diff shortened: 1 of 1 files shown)


nuclearcat:
This PR have indentation errors that breaks staging, i added this case in validator in tests.

config/pipeline.yaml
```diff
@@ -341,6 +342,13 @@ jobs:
   kbuild-gcc-12-arm64:
     <<: *kbuild-gcc-12-arm64-job
 
+  # Default config and build only job
+  kbuild-gcc-12-arm64-build-only:
+    <<: *kbuild-gcc-12-arm64-job
+  rules:
```
nuclearcat, line 348: Indentation is not correct

musamaanjum:
It doesn't seem like the arnd:to-build (https://git.kernel.org/pub/scm/linux/kernel/git/arnd/playground.git/commit/?h=to-build) gets updated very often. It is a hassle to manually trigger builds for testing before merging the PR. How can it be resolved?

config/pipeline.yaml
```diff
@@ -341,6 +342,13 @@ jobs:
   kbuild-gcc-12-arm64:
     <<: *kbuild-gcc-12-arm64-job
 
+  # Default config and build only job
+  kbuild-gcc-12-arm64-build-only:
+    <<: *kbuild-gcc-12-arm64-job
+    rules:
+      tree:
```
a-wai, line 349: This makes sense overall, with #718 in mind it would be nice to have an anchor for "build-only" rules, for example:

```yaml
  build-only-trees: &build-only-trees-rules
    rules:
      tree:
      - 'arnd'
```

And then this job definition would end up as:
```yaml
  kbuild-gcc-12-arm64-build-only:
    <<: *kbuild-gcc-12-arm64-job
    rules: *build-only-trees-rules
```

a-wai:
Also, please write actual commit messages! You're not just adding a tree here (even then, the commit message should give an indication why this tree is useful and/or who requested it to be added).

The commit message should also explain why you create build-only jobs instead of re-using existing ones, for example.

Also, please add `Fixes #<issue-nr>` to the commit message itself when closing an issue.

config/pipeline.yaml
musamaanjum: I'd created *-build-only job for some architectures. Some architectures have different name jobs, but they are build only. I'd reused those jobs. To create a uniformity, I'll create build-only-trees-rules and *-build-only job separately.

musamaanjum:
@a-wai I've updated the PR and resolved the above concerns. Feel free to review again.

config/pipeline.yaml
a-wai: Now, the question is: do we actually want to have a build-only job for all architectures, and do we want to execute all those jobs for all those trees? I mean, I'd expect the `amlogic` tree to be irrelevant to all architectures except `arm*`.

musamaanjum: There could be a division that some trees should only be tested on their intended architectures. But it would be hard to get it right. From legacy what I've understood that we run tests for all architectures minimally on these trees without much thought to actual intended architecture.

I guess at this point, we should think from just a migration point of view and port the configurations same to same from legacy to new KernelCI.

musamaanjum:
Confirmed results of builds on ardb and amlogic as well.
[Uploading ardb.html.txt…]()

PR #707, opened 2024-07-19
config: pipeline: remove params

musamaanjum:
The parameters are only needed when they are changed or appeneded. Remvoe the parameters which aren't being modified.


musamaanjum:
@nuclearcat if you agree with this PR, it can be merged quickly.

nuclearcat:
Let's run one staging run and then we can merge

PR #708, opened 2024-07-22
validate_yaml.py: Jobs are required to have template parameter

nuclearcat:
Add more validation to config files of mandatory parameters.


nuclearcat:
Merging asap as we have breakage on staging related to this mistake

PR #710, opened 2024-07-22
doc/developer-documentation: document `tree:branch` rule

JenySadadia:
Add details about specifying scheduler job rule for a specific branch of a specific tree.

### DIFF
--- a/doc/developer-documentation.md
+++ b/doc/developer-documentation.md
@@ -53,6 +53,8 @@ jobs:
       tree:
         - <tree-name>
         - !<tree-name2>
+      branch:
+        - <branch-name>
 
   <test-job-name>:
     template: <template-name>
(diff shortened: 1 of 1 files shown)


doc/developer-documentation.md
```diff
@@ -69,13 +71,15 @@ jobs:
...
 - **`params`**: The `params` field includes parameters for building the kernel (for `kbuild` jobs) or running the test. These parameters can include architecture, compiler, defconfig options, job timeout, etc.
-- **`rules`**: The `rules` field defines job rules. If a test should be scheduled for a specific kernel tree, branch, or version, these rules can be specified here. The rules prefixed with `!` exclude the specified condition from job scheduling. For example, in the given scenario, the scheduler does not schedule a job if an event is received for the kernel tree `tree-name2`.
+- **`rules`**: The `rules` field defines job rules. If a test should be scheduled for a specific kernel tree, branch, or version, these rules can be specified here. The rules prefixed with `!` exclude the specified condition from job scheduling. For example, in the given scenario, the scheduler does not schedule a job if an event is received for the kernel tree `tree-name2`. It is also possible to schedule job for a specific branch of a tree. For example, a rule `tree:branch` can be mentioned under
+`tree` or `branch` section to define that.
```
a-wai, line 82:
```suggestion
`tree` or `branch` section to define that, still accepting the `!` prefix to ensure jobs can be scheduled on all branches from a given tree, except a few select ones.
Rules can be defined in relation to any attribute present in the parent node (such as `defconfig` or `fragments` -- or even `arch` for jobs having a parent node of the `kbuild` type).
`min_version` and `max_version` are special cases allowing to allow jobs to run only on given kernel versions.
```

This would probably deserve a (sub)paragraph of its own though, so feel free to disregard this suggestion and I'll follow up on this asap.

doc/developer-documentation.md
```diff
@@ -69,13 +71,15 @@ jobs:
       tree:
         - <tree-name>
         - !<tree-name2>
+        - <tree-name3:branch-name>
```
a-wai, line 74:
```suggestion
        - <tree-name3>:<branch-name>
```

I think it makes the syntax a bit clearer.

a-wai (changes requested):

doc/developer-documentation.md
JenySadadia: I'll add these lines.

doc/developer-documentation.md
```diff
@@ -69,13 +71,18 @@ jobs:
...
+- **`rules`**: The `rules` field defines job rules. If a test should be scheduled for a specific kernel tree, branch, or version, these rules can be specified here. The rules prefixed with `!` exclude the specified condition from job scheduling. For example, in the given scenario, the scheduler does not schedule a job if an event is received for the kernel tree `tree-name2`. It is also possible to schedule job for a specific branch of a tree. For example, a rule `tree:branch` can be mentioned under
+`tree` or `branch` section to define that, still accepting the `!` prefix to ensure jobs can be scheduled on all branches from a given tree, except a few select ones.
+Rules can be defined in relation to any attribute present in the parent node (such as `defconfig` or `fragments` or even `arch` for build jobs or
+test jobs such as `kunit`).
```
a-wai, line 84:
```suggestion
Rules can be defined in relation to any attribute present in the parent node (such as `defconfig` or `fragments`, or even `arch` for test jobs such as `kunit`).
```

`arch` can't be used for build jobs, as they're created from a checkout node, which doesn't have this attribute.

a-wai (changes requested):

doc/developer-documentation.md
JenySadadia: I see `params.arch` in a couple of build jobs such as https://github.com/kernelci/kernelci-pipeline/blob/main/config/pipeline.yaml#L51

a-wai: Indeed, but rules don't take params into account, only node data; moreover, having an `arch` rule for build jobs wouldn't make sense as this is indeed a job parameter, so we're sure it won't run for any other arch than the one defined in `params.arch`.

JenySadadia: Okay. Isn't the same applicable for `fragments` and `defconfig`?
Those are also `params` attributes.

a-wai: Yes, you're absolutely right! kbuild nodes can only use tree/branch/version rules (simply put, what is available from checkout nodes) indeed.

JenySadadia: Okay, thanks for clarifying.
Fixed the line.

a-wai (approved):
LGTM, thanks!

PR #712, opened 2024-07-22
workflows: Add label on CI check failures

nuclearcat:
Automatically add label so broken PR wont go to staging

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,4 +48,19 @@ jobs:
       - name: Run basic yaml validation
         run: |
           python tests/validate_yaml.py
-        
+  on-fail:
+    if: failure() && github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    needs: check
+    steps:
+      - name: Add label to PR
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const label = 'staging-skip';
+            github.rest.issues.addLabels({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              labels: [label]
+            });


.github/workflows/main.yml
```diff
@@ -48,4 +48,24 @@ jobs:
...
+    steps:
+      - name: Add label to PR
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const label = 'staging-skip';
+            const { context, github } = require('@actions/github');
+            const owner = context.repo.owner;
+            const repo = context.repo.repo;
+            const pull_number = context.payload.pull_request.number;
```
a-wai, line 64: I assume declaring those variables is needed due to the use of `await`? If that's right, then LGTM.

nuclearcat: On their README seems they declare values as const before await (but sometimes not), so i preferred to be on safe side.

a-wai (approved):

.github/workflows/main.yml
```diff
@@ -48,4 +48,24 @@ jobs:
...
+    if: failure()
+    runs-on: ubuntu-latest
+    needs: check
+    steps:
+      - name: Add label to PR
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const label = 'staging-skip';
+            const { context, github } = require('@actions/github');
```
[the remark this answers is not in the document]
nuclearcat: Not tested yet, will spin up on my work and try to test.
I think this declaration is not correct, it was taken from some old code.
Likely:
const github = require('@actions/github');
const context = github.context;
and even not sure about that, will have to test.

.github/workflows/main.yml
pawiecz, line 64: This action runs also on [direct pushes to `main`](https://github.com/kernelci/kernelci-pipeline/pull/712/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R9).

Adding a `staging-skip` label wouldn't make much sense in that context (there's no PR) but should this case be also covered somehow if any of the validation steps fails on such a push?

.github/workflows/main.yml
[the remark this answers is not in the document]
pawiecz: It turns out in v7 this line can be dropped entirely, see _successful_ run (in terms of expected results) in the `workflow-test` repo: https://github.com/pawiecz/workflow-test/actions/runs/10057355394/job/27798378215

.github/workflows/main.yml
```diff
@@ -48,4 +48,24 @@ jobs:
...
+            const owner = context.repo.owner;
+            const repo = context.repo.repo;
+            const pull_number = context.payload.pull_request.number;
+
+            await github.issues.addLabels({
```
pawiecz, line 66: Adding labels requires changing the repo settings: Actions -> General -> Workflow Permissions -> Read and Write permissions

pawiecz, line 66: also:
```suggestion
            await github.rest.issues.addLabels({
```

.github/workflows/main.yml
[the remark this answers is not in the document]
nuclearcat: Thanks, just updated, seems with v7 many things changed in syntax and definitions, also tested locally

.github/workflows/main.yml
nuclearcat: fixed and set only on pull_request

PR #715, opened 2024-07-23
config: pipeline: Port chrome-platform to new KernelCI

musamaanjum:
Enable chrome-platform tree and its 2 branches. Only builds are run over
    it over default configs on arm, arm64 and x86.
    
    The job for arm64 is being created as there is no other suitable job
    with only arm64-chromebook. (There is kbuild-gcc-12-arm64-chromebook-kcidebug
    which has kcidebug fragment.) No need to create job for x86 as
    kbuild-gcc-12-x86 covers it. Specify chrome-platform tree in
    kbuild-gcc-12-arm-multi_v7_defconfig.
                                           
Closes https://github.com/kernelci/kernelci-project/issues/422

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -426,6 +426,16 @@ jobs:
       fragments:
        - CONFIG_RANDOMIZE_BASE=y
 
+  kbuild-gcc-12-arm64-chrome:
+    <<: *kbuild-gcc-12-arm64-job
+    params:
+      <<: *kbuild-gcc-12-arm64-params
+      fragments:
+       - 'arm64-chromebook'
+    rules:
+      tree:
+      - 'chrome-platform'
+
   kbuild-gcc-12-arm64-mfd:
     <<: *kbuild-gcc-12-arm64-job
     rules:
@@ -499,6 +509,7 @@ jobs:
       tree:
       - 'stable-rc'
       - 'kernelci'
+      - 'chrome-platform'
 
   kbuild-gcc-12-arm-mfd:
     <<: *kbuild-gcc-12-arm-job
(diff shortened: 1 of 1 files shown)


a-wai (changes requested):
LGTM, but can you please give more context (preferably in the commit message), like why are we adding this tree, what jobs are expected to run on it (build checks, basic tests, chromebooks tests?), etc...

config/pipeline.yaml
```diff
@@ -813,6 +824,16 @@ jobs:
       tree:
       - 'android'
 
+  kbuild-gcc-12-x86-chrome:
+    <<: *kbuild-gcc-12-x86-job
+    params:
+      <<: *kbuild-gcc-12-x86-params
+      fragments:
+        - x86-board
```
a-wai, line 832: Would having the `lab-setup` fragment be a problem? If the answer is no, then there's no need for this job, it will run on `kbuild-gcc-12-x86` and `kbuild-gcc-12-x86-chromeos` (which has 2 more options, but nothing really significant).

Similarly, those build jobs would run for arm64 as well with `kbuild-gcc-12-arm64-chromeos`.

Overall, IMHO adding new jobs is unnecessary here, so the only changes needed would be:
* add the tree and build configs
* allow this tree for the arm build as you did above

musamaanjum: Removed kbuild-gcc-12-x86-chromeos. kbuild-gcc-12-arm64-chromeos cannot be removed as there is only kbuild-gcc-12-arm64-kcidebug job which could be used if it didn't have kcidebug fragment.

musamaanjum: Edited above comment.

I'm confused. Let me share what I've understood.
`kbuild-gcc-12-arm64-chromeos` is an anchor in jobs-chromeos.yaml. `kbuild-gcc-12-arm64-chromebook` is the job using the anchor `kbuild-gcc-12-arm64-chromeos` which builds all trees other than android. Thus should I just remove `kbuild-gcc-12-arm64-chrome` from `pipeline.yaml`?

musamaanjum:
Verified results: 
[platform.html.txt](https://github.com/user-attachments/files/16388596/platform.html.txt)

config/pipeline.yaml
musamaanjum: @a-wai what do you think?

Issue #714, opened 2024-07-23
Add stable-rc 6.9 kernel version


crazoes:
Currently, we have 5.4, 5.10, 5.15, 6.1 and 6.6 stable-rc branches added to the new KernelCI.
We would like to add 6.9 kernel version as well since the template for sending automated stable-rc reports is almost ready.

Details :- 

Tree: stable-rc
Branch: linux-6.9.y
Link: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git/log/?h=linux-6.9.y

PR #716, opened 2024-07-23
config: pipeline: enable 6.9.y stable-rc

musamaanjum:
The request for this addition can be found: https://github.com/kernelci/kernelci-pipeline/issues/714


a-wai:
Thanks! 6.10.y stable-rc would make sense as well, since 6.10 has been released recently.

PR #717, opened 2024-07-23
config: scheduler-chromeos: enable codec Tast tests on corsola

a-wai:
This device's hardware decoder is able to process H.264, VP8 and VP9.


nuclearcat:
Tested on staging

PR #719, opened 2024-07-23
config: jobs-chromeos: set large timeout for `tast-decoder-chromestack`

a-wai:
With recent CrOS versions, those jobs can run for a veeery long time (> 2hrs). Increase the job timeout to 3 hrs so we can finally get meaningful results.

Depends on https://github.com/kernelci/kernelci-core/pull/2621


nuclearcat:
LGTM

PR #728, opened 2024-07-24
config: pipeline: de-duplicate entries using anchors

a-wai:
There are multiple instances of nearly-identical job/scheduler entries (and subsets of those, such as runtime selection for scheduler entries).

This change aims at reducing redundancies by using YAML anchors everywhere it can be relevant, and similarly removes a few anchors used only once.

Other minor fixes (node names, int vs. string value) are also included in this PR as a separate commit.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -29,6 +29,10 @@ _anchors:
     runtime:
       name: k8s-all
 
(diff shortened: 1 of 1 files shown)


a-wai:
Merged as part of #729

PR #731, opened 2024-07-26
pipeline.yaml: Disable docker kunit

nuclearcat:
Docker type can't really run on production (because pipeline runs in k8s itself), and even in staging it is DinD and makes unnecessary high load on each run. Disable, but dont remove, as later we might use such type for some lightweight staging jobs.


nuclearcat:
Verified on staging

PR #734, opened 2024-07-29
pipeline/api: tools: Add changes that was lost due mistake

nuclearcat:
@pawiecz had several suggestions that i amended, but by mistake i amended workflow patch, which i dropped later. Returning them back.


pawiecz:
Reference to the previous PR: https://github.com/kernelci/kernelci-pipeline/pull/691

pawiecz (approved):
LGTM

PR #737, opened 2024-07-30
config: pipeline: fix rt-tests scheduling

musamaanjum:
The rt-tests were wrongly set to trigger based on chromebook jobs which didn't had preempt-rt fragment. Fix it by specifying the correct event which is the name of preempt-rt kbuild job names. Run each of 4 enabled rt-test on each variant of kernel. The kbuild variant are 2 arm64, 2 x86 and 1 arm.

Schedule the tests on a few devices instead of scheduling on all of the devices. Select devices from different vendors for each architecture.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -1825,6 +1825,143 @@ scheduler:
     platforms:
       - qemu-x86
 
+  - job: rt-tests-cyclicdeadline
+    event: &kbuild-gcc-12-arm64-preempt_rt-node-event
+      channel: node
+      name: kbuild-gcc-12-arm64-preempt_rt
+      result: pass
+    runtime: *runtime-lava-collabora
+    platforms: &arm64-preempt_rt-platforms
+      - bcm2711-rpi-4-b
+      - meson-g12b-a311d-khadas-vim3
+      - rk3399-gru-kevin
+      - rk3399-rock-pi-4b
+      - rk3588-rock-5b
+      - sun50i-h6-pine-h64
+
+  - job: rt-tests-cyclicdeadline
+    event: &kbuild-gcc-12-arm64-preempt_rt_chromebook-node-event
+      <<: *kbuild-gcc-12-arm64-preempt_rt-node-event
+      name: kbuild-gcc-12-arm64-preempt_rt_chromebook
+    runtime: *runtime-lava-collabora
+    platforms: &arm64-preempt_rt_chromebook-platforms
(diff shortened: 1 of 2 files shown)


config/pipeline.yaml
```diff
@@ -1825,6 +1825,135 @@ scheduler:
     platforms:
       - qemu-x86
 
+  - job: rt-tests-cyclicdeadline
+    event: &kbuild-gcc-12-arm64-preempt_rt-node-event
+      channel: node
+      name: kbuild-gcc-12-arm64-preempt_rt
+      result: pass
+    runtime: *runtime-lava-collabora
+    platforms: &kbuild-gcc-12-arm64-preempt_rt-platforms
```
a-wai, line 1834:
```suggestion
    platforms: &arm64-preempt_rt-platforms
```

Please trim other "platforms" anchors in a similar way

config/pipeline.yaml
```diff
@@ -1825,6 +1825,135 @@ scheduler:
...
+      - mt8183-kukui-jacuzzi-juniper-sku16
+      - mt8186-corsola-steelix-sku131072
+      - sc7180-trogdor-kingoftown
```
a-wai, line 1837: Do those really work when not using a chromebook kbuild?

config/pipeline.yaml
```diff
@@ -1825,6 +1825,135 @@ scheduler:
...
+    event: &kbuild-gcc-12-x86-preempt_rt-node-event
+      channel: node
+      name: kbuild-gcc-12-x86-preempt_rt
+      result: pass
+    runtime: *runtime-lava-collabora
+    platforms: &kbuild-gcc-12-x86-preempt_rt-platforms
+      - acer-cb317-1h-c3z6-dedede
+      - acer-cbv514-1h-34uz-brya
+      - acer-R721T-grunt
+      - acer-cp514-3wh-r0qs-guybrush
```
a-wai, line 1859: Same question about not using a chromebook kbuild for those.

a-wai (changes requested):

config/pipeline.yaml
musamaanjum: I don't think so. I've changed the platforms without chromebook fragment.

config/pipeline.yaml
musamaanjum: Fixed now.

a-wai (approved):
LGTM, a few nits but I'll address those when rebasing #728 depending on when we merge things...

PR #738, opened 2024-07-31
Enable efi, clk and peterz trees

musamaanjum:
(see individual commits' description for information)

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -1670,6 +1670,9 @@ scheduler:
 
   - job: kbuild-gcc-12-riscv
     <<: *build-k8s-all
+    rules:
+      tree:
+        - '!efi'
 
   - job: kbuild-gcc-12-riscv-android-defconfig
     <<: *build-k8s-all
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -117,6 +117,8 @@ _anchors:
       - 'amlogic'
       - 'ardb'
       - 'arnd'
+      - 'clk'
+      - 'peterz'
 
 api:
 
@@ -1228,6 +1230,9 @@ trees:
   chrome-platform:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/chrome-platform/linux.git"
 
+  clk:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/clk/linux.git"
+
   kernelci:
     url: "https://github.com/kernelci/linux.git"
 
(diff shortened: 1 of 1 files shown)


config/pipeline.yaml
```diff
@@ -1913,6 +1934,14 @@ build_configs:
     branch: 'for-linus'
     variants: *build-variants
 
+  clk:
+    tree: clk
+    branch: 'clk-next'
+
```
a-wai, line 1940:
```suggestion
    variants: *build-variants

```

For consistency (to be added to other build config entries as well).

config/pipeline.yaml
```diff
@@ -117,6 +117,18 @@ _anchors:
       - 'amlogic'
       - 'ardb'
       - 'arnd'
+      - 'clk'
+      - 'peterz'
+
+  build-only-trees: &build-only-trees-rules-minimal
```
a-wai, line 123:
```suggestion
  build-only-trees-minimal: &build-only-trees-minimal-rules
```

The entry has the same name as the previous one (`build-only-trees`) which should be avoided. Moreover, then anchor name should end with the "kind", in this case `-rules`.

I'm not a fan of having that many duplicated entries. If it's really necessary to have 2 sets of build-only trees, I'd rather have those as separate entries; although it feels better to have a single set and simply exclude `efi` adding a simple rule to scheduler entries where it shouldn't be used.

a-wai (changes requested):

config/pipeline.yaml
musamaanjum: @a-wai I've not been specifying variants for newly added trees from the time when it was indicated that these variants are legacy way of specifying them and we should avoid them. Why should variant be specified for these trees?

config/pipeline.yaml
musamaanjum: I'd not thought of it like this. I've updated as you have asked.

a-wai (approved):

config/pipeline.yaml
```diff
@@ -1664,6 +1676,9 @@ scheduler:
 
   - job: kbuild-gcc-12-riscv-build-only
     <<: *build-k8s-all
+    rules:
```
a-wai, line 1679: Actually I assume we also want to apply this rule to `kbuild-gcc-12-riscv` as well, right? This job [got triggered](https://staging.kernelci.org:9000/viewer?node_id=66b20bee2fbd8114fc0b7438) on `efi` (worked fine to exclude ARC/MIPS builds though, and `kbuild-gcc-12-riscv-build-only` as well)

a-wai (changes requested):

config/pipeline.yaml
musamaanjum: Done

a-wai (approved):

PR #739, opened 2024-07-31
lava_callback.py: Extend /checkout endpoint functionality

nuclearcat:
Before user can use reference nodeid and custom commitid to submit custom checkout node.
Now we allow also user to submit tree information and jobfilter, but all this data will be validated in config, e.g. user can submit tree that only exist in pipeline configs, job names that exist in jobs: and etc.


nuclearcat:
Tested on staging

Issue #682, opened 2024-07-05
Implement purge mechanism


JenySadadia:
As we have enabled log dumping to files for pipeline services, we need a purge mechanism to delete log files at a specific interval.
Otherwise, the machine running maestro will complain about storage after a while.

Related to https://github.com/kernelci/kernelci-project/issues/334

PR #740, opened 2024-08-01
Use timed rotation logging

JenySadadia:
Fixes https://github.com/kernelci/kernelci-pipeline/issues/682

Use timed rotating file handler to optimize storage space for log files instead of using a regular file handler.
It will create a new log file with `<service-name>.log` convention and keep one backup file for older logs than
7 days. It will delete older backup log files on rollover if exists.
In case of `logs` directory is not mounted due to a different environment such as `k8s`, the service would crash with
`FileNotFoundError`. Enable only console logging when file logging fails to fix the issue.

### DIFF
--- a/config/logger.conf
+++ b/config/logger.conf
@@ -22,7 +22,7 @@ keys=root,
      trigger
 
 [handlers]
-keys=consoleHandler, fileHandler
+keys=consoleHandler, timedRotatingHandler
 
 [formatters]
 keys=defaultFormatter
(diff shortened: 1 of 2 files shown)


nuclearcat (approved):
Amazing, exactly what we need. Thank you Jeny, again, for saving the day. I might roll tomorrow (Friday) production, but... might be bad idea, i will explain why, it might wait until Monday.

JenySadadia:
Thank you Denys :)
Atm I have set the rotation duration to 7 days. I guess we should set at least 15 days.
What should we keep from storage PoV?

nuclearcat:
I think this logs wont take much space, so 15 days should be fine.

JenySadadia:
> I think this logs wont take much space, so 15 days should be fine.

Thanks. Updated the config to keep 15 days.

nuclearcat:
Seems working on staging fine, will test on k8s deploy today. Merging.

PR #742, opened 2024-08-02
Auto restart services on failure

JenySadadia:
Restart pipeline container when it exists with an error using `restart` policy.


JenySadadia:
@nuclearcat Do we need some kind of notification when a service restarts?

nuclearcat:
I think no need for now.
In worst case i can set something in logger we have, and we can see if service restarts in docker ps or docker-compose ps

nuclearcat:
Looks like it improved things on staging!

Issue #636, opened 2024-06-11
Report test configuration details in the nodes


laura-nao:
This issue is to initiate an open discussion regarding the possibility of providing users with more comprehensive information about the specific configurations of test suites through the KernelCI nodes.

Tests may be executed with different parameters across different platforms (depending on factors such as hardware configuration). Thus, it's important to provide details on the configuration/parameters used on a specific test run. This enables users to know how the test was conducted on each platform and to easily identify if a test failure correlates with the specific configuration in use.

This may become particularly relevant with the addition of support for running kselftests with parameters (see https://github.com/Linaro/test-definitions/pull/511).

Some possible solutions:
- Consider adding extra node fields to report the command used for the test and/or the parameters applied (where applicable).
- Ensure that test suites running with different parameters are assigned distinct names or IDs. This may be challenging to enforce as a rule, and users may still need to refer to the YAML files for detailed test information.

Some test suites can be harder to describe than others, so I feel like the challenge is to find a way to effectively describe the test configuration that works for different test suites (maybe a link to the test job template used is enough?).

Any thoughts on this?

musamaanjum:
CI systems like 0-day and Syzbot are successful as they provide the entire information to reproduce the failures. Anybody should be able to use the email/dashboard data to investigate and fix the issue. We should try to follow the same motto that the entire information must be present in bug reports/dashboards. We can add information in a step by step process.

One thing that I've noticed recently in legacy KernelCI is that the information regarding used rootfs isn't easily available. So its very difficult to fix issue as you don't have exact reproducer environment.

PR #743, opened 2024-08-02
Enable suspend and hibernation test

crazoes:
Enable suspend and hibernation test. This patch depends on the https://github.com/kernelci/kernelci-pipeline/pull/743

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -910,6 +910,28 @@ jobs:
       job_timeout: 10
     kcidb_test_suite: kselftest.cpufreq
 
+  kselftest-cpufreq-hibernate:
+    <<: *kselftest-job
+    params:
+      <<: *kselftest-params
+      collections: cpufreq
+      env: 'KSELFTEST_MAIN_SH_ARGS="-t hibernate_rtc"'
+    rules:
+      tree:
+      - 'kernelci:staging-next'
+    kcidb_test_suite: kselftest.cpufreq.hibernate
+
+  kselftest-cpufreq-suspend:
+    <<: *kselftest-job
+    params:
+      <<: *kselftest-params
+      collections: cpufreq
+      env: 'KSELFTEST_MAIN_SH_ARGS="-t suspend_rtc"'
+    rules:
+      tree:
+      - 'kernelci:staging-next'
+    kcidb_test_suite: kselftest.cpufreq.suspend
+
   kselftest-dmabuf-heaps:
     <<: *kselftest-job
     params:
--- a/config/scheduler-chromeos.yaml
+++ b/config/scheduler-chromeos.yaml
@@ -216,6 +216,30 @@ scheduler:
   - job: kselftest-cpufreq
     <<: *test-job-arm64-mediatek
 
+  - job: kselftest-cpufreq-hibernate
+    <<: *test-job-x86-intel
+
+  - job: kselftest-cpufreq-hibernate
+    <<: *test-job-x86-amd
+
+  - job: kselftest-cpufreq-hibernate
+    <<: *test-job-arm64-qualcomm
+
+  - job: kselftest-cpufreq-hibernate
+    <<: *test-job-arm64-mediatek
+
+  - job: kselftest-cpufreq-suspend
+    <<: *test-job-x86-intel
+
+  - job: kselftest-cpufreq-suspend
+    <<: *test-job-x86-amd
+
+  - job: kselftest-cpufreq-suspend
+    <<: *test-job-arm64-qualcomm
+
+  - job: kselftest-cpufreq-suspend
+    <<: *test-job-arm64-mediatek
+
   - job: kselftest-dmabuf-heaps
     <<: *test-job-x86-intel
 


config/pipeline.yaml
```diff
@@ -920,6 +920,26 @@ jobs:
...
 
+  kselftest-cpufreq-suspend:
+    template: kselftest.jinja2
+    kind: job
+    params:
+      nfsroot: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240313.0/{debarch}'
+      collections: cpufreq
+      job_timeout: 10
+      env: 'KSELFTEST_MAIN_SH_ARGS="-t suspend_rtc"'
+    kcidb_test_suite: kselftest.cpufreq
```
laura-nao, line 931: I'm not sure if we need to define a different `kcidb_test_suite` here, as this is the same cpufreq kselftest but run with different parameters. @JenySadadia is there an existing convention for this? Users need a way to differentiate between the same test executed with different parameters once the results land in KCIDB.
(This goes back to the discussion raised [here](https://github.com/kernelci/kernelci-pipeline/issues/636))

laura-nao:
Hi @crazoes - were you planning to add only the test definition at this stage, or also to run the test on some devices? If it’s the latter, you’ll need to include the test in the `scheduler:` section (either in `pipeline.yaml` or `scheduler-chromeos.yaml`), similar to what was done in https://github.com/kernelci/kernelci-pipeline/pull/588

crazoes:
@laura-nao I just wanted to first see if this approach is correct so thought of only adding the test definition. But I can make changes in my v2 as per the reviews

config/pipeline.yaml
laura-nao: Sounds good to me, thanks @JenySadadia ! About the parameters passed to this test, is there any field we can add or re-use to encode this information? besides being useful for filtering results, it's also good information for manually reproducing the issues.

JenySadadia: One way is to store that information in node `data` field and later on assign that to `misc` field for KCIDB submission.

config/pipeline.yaml
```diff
@@ -920,6 +920,26 @@ jobs:
       job_timeout: 10
     kcidb_test_suite: kselftest.cpufreq
 
+  kselftest-cpufreq-suspend:
+    template: kselftest.jinja2
```
pawiecz, line 924: This job template no longer exists. It was replaced by [`generic.jinja2`](https://github.com/kernelci/kernelci-pipeline/blob/main/config/runtime/generic.jinja2) - see [the example](https://github.com/kernelci/kernelci-pipeline/blob/main/doc/developer-documentation.md#an-example-of-enabling-a-new-job) for more details.

crazoes: @pawiecz this is a bit confusing. I see quite different threads here https://github.com/kernelci/kernelci-core/pull/2626
too.
We just want to add parameter support and for that we needed kselftest.jinja2 ( see https://github.com/Linaro/test-definitions/pull/511 )

a-wai: You have to rebase as there have been significant changes, and make it similar to other kselftest job definitions: [link](https://github.com/kernelci/kernelci-pipeline/blob/main/config/pipeline.yaml#L898)

crazoes: @a-wai thanks, I just forced pushed the changes after rebasing my branch

config/pipeline.yaml
```diff
@@ -895,6 +895,30 @@ jobs:
       job_timeout: 10
     kcidb_test_suite: kselftest.cpufreq
 
+  kselftest-cpufreq-suspend: &kselftest-job
```
a-wai, line 898: No, please make use of the YAML anchors instead of redefining them (see the definition for `kselftest-dmabuf-heaps` just below). Your new definitions should have only the `collections` and `env` params explicitly defined, everything else should come from anchors.

a-wai (changes requested):

config/pipeline.yaml
crazoes: @a-wai thanks for reviewing it, I just force-pushed the changes as you requested.

a-wai: Thanks, looks good now

a-wai (approved):

a-wai:
There are still errors on the kci-core side, but this PR is working as expected: multiples instances of `kselftest-cpufreq-{hibernate,suspend}` in [this node](https://staging.kernelci.org:9000/viewer?search=parent%3D66c8015d90812a603fbd41fc)

crazoes:
@a-wai @pawiecz I updated this PR to add a rule for running this test only on kernelci/staging-next. Hopefully this will fix the failing test.

a-wai (approved):

crazoes:
If everything looks good then can we merged this now @nuclearcat @a-wai ?

config/pipeline.yaml
```diff
@@ -910,6 +910,28 @@ jobs:
       job_timeout: 10
     kcidb_test_suite: kselftest.cpufreq
 
+  kselftest-cpufreq-hibernate:
+    <<: *kselftest-job
+    params:
+      <<: *kselftest-params
+      collections: cpufreq
+      env: 'KSELFTEST_MAIN_SH_ARGS="-t hibernate_rtc"'
+    rules:
```
musamaanjum, line 919: Maybe we want to run this test on all the trees and not just kernelci. I may not have enough context for this. So what do you think?

nuclearcat:
Did we got results of tests on dashboard or in viewer?

crazoes:
@nuclearcat yes we did get them but lots of tests are failing. https://lava.collabora.dev/scheduler/alljobs?length=25&search=suspend#table

It's not really the problem related to the tests or environment but mostly related to the kernel I think.

nuclearcat:
Thanks! I think soon we will start looking into such problems more.

PR #744, opened 2024-08-06
feature(api): Add patchset endpoint

nuclearcat:
Patchset initial implementation to apply patch to existing checkout node.


nuclearcat:
Merging, as it is tested on staging, and going to final round of fixing discovered bugs and changing small details

PR #747, opened 2024-08-07
config: fix remanining sorting issues

a-wai:
A few entries were still left improperly ordered, both in the main pipeline config and in ChromeOS jobs definitions, as reported by `kci config validate`. Let's fix those.


nuclearcat:
Staging works fine

PR #750, opened 2024-08-08
fix(scheduler): Don't crash if runtime.generate failed

nuclearcat:
```
[ERROR] Traceback (most recent call last):
  File "/home/kernelci/pipeline/base.py", line 69, in run
    status = self._run(context)
             ^^^^^^^^^^^^^^^^^^
  File "/home/kernelci/./pipeline/scheduler.py", line 197, in _run
    self._run_job(job, runtime, platform, input_node)
  File "/home/kernelci/./pipeline/scheduler.py", line 120, in _run_job
    data = runtime.generate(job, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kernelci/runtime/lava.py", line 303, in generate
    return yaml.dump(yaml.load(rendered, Loader=yaml.CLoader))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/yaml/__init__.py", line 81, in load
    return loader.get_single_data()
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/yaml/constructor.py", line 49, in get_single_data
    node = self.get_single_node()
           ^^^^^^^^^^^^^^^^^^^^^^
```
Produce a bit more error information, and dont crash whole scheduler process.

### DIFF
--- a/src/scheduler.py
+++ b/src/scheduler.py
@@ -117,7 +117,29 @@ class Scheduler(Service):
         }
         extra_args.update(job.config.params)
         params = job.platform_config.format_params(params, extra_args)
-        data = runtime.generate(job, params)
+        # we experience sometimes that the job is not created properly
+        # due exception in the runtime.generate method
+        try:
+            data = runtime.generate(job, params)
+        except Exception as e:
+            self.log.error(' '.join([
+                node['id'],
+                runtime.config.name,
+                platform.name,
+                job_config.name,
+                str(e),
+            ]))
+            node['state'] = 'done'
+            node['result'] = 'incomplete'
+            node['data']['error_code'] = 'job_generation_error'
+            node['data']['error_msg'] = str(e)
+            try:
+                self._api.node.update(node)
+            except requests.exceptions.HTTPError as err:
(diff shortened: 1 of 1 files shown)


nuclearcat:
> ooc, what's the point of creating a `Node` for a job that doesn't result in any actual action done? Can't this same condition be simply notified in the scheduler log?
For reports and alerting there is problem with job generation, and it is likely invalid configuration that need to be fixed.

nuclearcat:
> > For reports and alerting there is problem with job generation, and it is likely invalid configuration that need to be fixed.
> 
> Yes, my question was: why a `Node` for this rather than a log message or other kind of notification mechanism that's more suitable for, well, for notifications (an event, for instance)?

At moment when we took this decision we didnt had even permanent logging for staging. And we dont have (yet) logging for kubernetes instance and doesn't look like there is easy solution for that.

PR #751, opened 2024-08-08
src/send_kcidb: handle incomplete builds

JenySadadia:
As KCIDB only supports `pass` and `fail` status for builds and doesn't have a separate status for incomplete/missed build jobs, keep `valid` flag empty for incomplete builds.

### DIFF
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -166,6 +166,14 @@ class KCIDBBridge(Service):
             return data[-(16*1024):]
 
     def _parse_build_node(self, origin, node):
+        result = node.get('result')
+        result_map = {
+            'pass': True,
+            'fail': False,
+            'incomplete': None,
+        }
+        valid = result_map.get(result) if result else None
+
         parsed_build_node = {
             'checkout_id': f"{origin}:{node['parent']}",
             'id': f"{origin}:{node['id']}",
@@ -175,7 +183,7 @@ class KCIDBBridge(Service):
             'architecture': node['data'].get('arch'),
             'compiler': node['data'].get('compiler'),
             'config_name': node['data'].get('defconfig'),
-            'valid': node['result'] == 'pass',
+            'valid': valid,
             'misc': {
                 'platform': node['data'].get('platform'),
                 'runtime': node['data'].get('runtime'),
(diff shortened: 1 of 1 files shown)


JenySadadia:
Tested OK on staging:
https://kcidb.kernelci.org/d/build/build?orgId=1&var-datasource=playground&var-id=maestro:66b6fa324fe62d823df5d40b&var-origin=maestro&var-build_architecture=All&var-build_config_name=All

JenySadadia:
Added a commit for not sending timeout error metadata as it won't be reset in the second KCIDB submission if the job completes after some time with no failure.

src/send_kcidb.py
```diff
@@ -175,15 +176,16 @@ def _parse_build_node(self, origin, node):
             'architecture': node['data'].get('arch'),
             'compiler': node['data'].get('compiler'),
             'config_name': node['data'].get('defconfig'),
-            'valid': node['result'] == 'pass',
+            'valid': node['result'] if node['result'] in ('pass', 'fail') else None,
```
pawiecz, line 179: Previously `valid` contained a bool (`True`/`False`) and now it might be str ("pass"/"fail") or None - was that expected? Is it handled on the KCIDB side?

JenySadadia: You're right. Thanks a lot for catching this.
I'll fix it.

src/send_kcidb.py
```diff
@@ -166,6 +166,15 @@ def _get_log_excerpt(self, log_url):
             return data[-(16*1024):]
 
     def _parse_build_node(self, origin, node):
+        result = node.get('result')
+        result_map = {
+            'pass': True,
+            'fail': False,
+            'incomplete': None,
```
pawiecz, line 173: nit: `incomplete` would get `valid=None` even if it was absent in the map - no pressure to remove it though

pawiecz (approved):
LGTM

src/send_kcidb.py
JenySadadia: I kept it as KCIDB will have a different status for it at some point and then we can update this mapping.

JenySadadia:
Tested OK on staging.

PR #752, opened 2024-08-09
fix(lava_callback): Fix all return codes/content

nuclearcat:
We had leftovers from flask that was returning invalid data, when any API calls was sent. Now we return proper JSON.


nuclearcat:
tested on staging

PR #754, opened 2024-08-14
feat(k8s): Add regression tracker manifest

nuclearcat:
By some reason regression tracker was not added in kubernetes manifests, and recently discovered by Ricardo. Fixing that.


nuclearcat:
Tested on production, merging

PR #757, opened 2024-08-15
Rename early access configs

JenySadadia:
Related to https://github.com/kernelci/kernelci-core/pull/2638
- Rename configurations for `early-access` to `production`
- Remove `lava-collabora-early-access` lab

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -134,7 +134,7 @@ api:
   staging:
     url: https://staging.kernelci.org:9000
 
-  early-access:
+  production:
     url: https://kernelci-api.westus3.cloudapp.azure.com
 
   k8s-host:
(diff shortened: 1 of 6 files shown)


config/pipeline.yaml
```diff
@@ -130,10 +130,10 @@ storage:
     port: 8022
     base_url: http://172.17.0.1:8002/
 
-  early-access-azure: &azure-files
+  production-azure: &azure-files
     storage_type: azure
     base_url: https://kciapistagingstorage1.file.core.windows.net/
-    share: early-access
+    share: production
```
JenySadadia, line 136: @nuclearcat Could you please confirm if updating `share` value won't break storage?

config/pipeline.yaml
```diff
@@ -212,14 +212,6 @@ runtimes:
...
         token: kernelci-api-token-staging
 
-  # ToDo: avoid creating a separate Runtime entry
-  # https://github.com/kernelci/kernelci-core/issues/2088
-  lava-collabora-early-access:
-    <<: *lava-collabora-staging
-    notify:
-      callback:
-        token: kernelci-api-token-early-access
-
```
JenySadadia, line 222: I couldn't find any job using this lab.
Is it safe to drop this lab? @nuclearcat

nuclearcat: seems yes, thanks for cleanup!

nuclearcat:
There is at least one file conflicting, can you fix pipeline.yaml so we can run over staging?

JenySadadia:
> There is at least one file conflicting, can you fix pipeline.yaml so we can run over staging?

Resolved the conflict and deployed on staging along with https://github.com/kernelci/kernelci-core/pull/2638.

pawiecz (approved):
LGTM

The same deployment policy I mentioned in [kernelci/kernelci-core#2638](https://github.com/kernelci/kernelci-core/pull/2638#pullrequestreview-2444859623) applies here

PR #760, opened 2024-08-20
feat(extract_compatible.py): Add new tool to extract "compatible" field

nuclearcat:
Add new tool for devices using DTB, to assist with compatible field.

Also update config files accordingly.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -1318,6 +1318,7 @@ platforms:
     <<: *arm-device
     mach: broadcom
     dtb: dtbs/bcm2836-rpi-2-b.dtb
+    compatible: ['raspberrypi,2-model-b', 'brcm,bcm2836']
 
   docker:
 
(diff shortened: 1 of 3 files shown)


config/pipeline.yaml
```diff
@@ -1318,33 +1318,38 @@ platforms:
     <<: *arm-device
     mach: broadcom
     dtb: dtbs/bcm2836-rpi-2-b.dtb
+    compatible: ['raspberrypi,2-model-b', 'brcm,bcm2836']
```
JenySadadia, line 1321: Why do these values need to be hard-coded?

nuclearcat:
> `compatible` property is missing for `bcm2711-rpi-4-b`.

yes, as it is missing also in dtb/dts for it

config/pipeline.yaml
nuclearcat: They are extracted once from dtb when we add platform.
Otherwise we will need to fetch dts file (or extract from dtb), parse it in pipeline service for each device, and all that for same value every time.
In rare cases they might be updated, but also it is better if we are aware about it.

JenySadadia: Thanks for clarifying.

PR #762, opened 2024-08-21
src/send_kcidb: load `compatible` field

JenySadadia:
Depends on https://github.com/kernelci/kernelci-core/pull/2644, https://github.com/kernelci/kernelci-pipeline/pull/760

Device tree compatible information has been added to platforms configuration in YAML files. 
Load the value if available and send it along with KCIDB test submission.


musamaanjum:
> @musamaanjum As per my understanding, the first value in `compatible` shows model information and then SoC families. Should I send the whole list or just `[1:]` ?

I think the whole list is needed. The list gives further information regarding the board.

PR #764, opened 2024-08-22
config/jobs-chromeos.yaml: update `kcidb_test_suite` for fluster tests

JenySadadia:
To distinguish different fluster tests on KCIDB dashboard, use syntax `fluster.v4l2.<media_framework>_<codec>` for KCIDB test suite mapping property.
This would make stats analysis possible such as total number of passed and failed tests for summary reports.


JenySadadia:
Tested OK on staging.

PR #769, opened 2024-08-27
config: pipeline: Enable remaining tests from rt-test repo

musamaanjum:
On request of rt-maintainer, enable the remaining tests. Keep the default timeout for now. As the test definitions are always present, only schedule these tests on all the different targets in the same manner as the previous rt-tests are being scheduled.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -1967,6 +1967,106 @@ scheduler:
     runtime: *lava-collabora-runtime
     platforms: *collabora-arm-preempt_rt-platforms
 
+  - job: rt-tests-pi-stress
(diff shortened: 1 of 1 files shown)


musamaanjum:
Related to https://github.com/kernelci/kernelci-project/issues/417

musamaanjum:
I've collected the results on https://github.com/kernelci/kernelci-project/issues/439 and fix issues under this ticket. Overall, 3/11 tests have issues in their executor script in https://github.com/Linaro/test-definitions.

musamaanjum:
@nuclearcat I've confirmed results for this PR as well.

PR #770, opened 2024-08-28
feat(lava_callback): Prepare prometheus compatible metrics

nuclearcat:
To monitor performance and possible failures we need metrics export endpoint available. This is initial implementation.


nuclearcat:
image
tested on staging

PR #772, opened 2024-09-02
src/send_kcidb: remove a log line for printing `compatible`

JenySadadia:
The log printing device tree compatible info was intended for development purpose. Hence, drop it.


JenySadadia:
Tested OK on staging.

PR #774, opened 2024-09-03
fix(lava_callback): Fix job name validation

nuclearcat:
Turned out job validation is not working, fixing it.


JenySadadia (approved):
Tested OK.

PR #775, opened 2024-09-03
kci-maintainer: Improve tool error readability

nuclearcat:
Often with HTTP error code we receive helpful body message, it will be good to show it.


JenySadadia (approved):
Tested OK locally.

PR #777, opened 2024-09-03
Fix `kci-maintainer` tool

JenySadadia:
- Fix getting token for `prepare_checkout` when token is not available in env variable
- Fix the URL for local pipeline API


nuclearcat:
Thanks!

PR #778, opened 2024-09-03
Extend `node.submitter`

JenySadadia:
Closes https://github.com/kernelci/kernelci-core/issues/2654

### DIFF
--- a/src/lava_callback.py
+++ b/src/lava_callback.py
@@ -506,6 +506,7 @@ async def checkout(data: ManualCheckout, request: Request,
             }
         },
         "timeout": checkout_timeout.isoformat(),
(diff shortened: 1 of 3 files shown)


JenySadadia:
@nuclearcat A couple of comments:
- I don't think we need to set `submitter` for job retry operation. As it should re-run existing job and submitter info would already be set on it
- I have used `user:<email>` convention for submitter field and didn't add any hash string as I guess only email would be enough to distinguish the maintainer

nuclearcat:
* For job retry agree, it might be ok, at least for now.

* For user:<email>, i am wondering if having *hash* might make easier to trace submission, or we just should introduce new field, like treehash, for easier tree tracking.

Issue #543, opened 2024-04-16
Add documentation for setting `checkout` result


JenySadadia:
Add documentation describing `checkout` node result and how it is being set atm.

Related to https://github.com/kernelci/kernelci-pipeline/pull/542

JenySadadia:
Similar issue https://github.com/kernelci/kernelci-pipeline/issues/422

PR #790, opened 2024-09-10
doc/developer-documentation: checkout result evaluation

JenySadadia:
Closes https://github.com/kernelci/kernelci-pipeline/issues/543

Explain `checkout` node result evaluation by pipeline services.

### DIFF
--- a/doc/developer-documentation.md
+++ b/doc/developer-documentation.md
@@ -268,6 +268,33 @@ It has two child nodes. One is a `setup` test suite as mentioned above, and anot
 `kselftest-cpufreq` (child job node) will be executed only after `setup` test suite's successful execution.
 
 
+## Checkout result evaluation
+
+Maestro pipeline creates a `checkout` node i.e., root node whenever a new kernel revision is detected.
+All the build and test jobs are created as child nodes of this root node.
+
+Below is the explanation of `checkout` node result evaluation based on different conditions.
+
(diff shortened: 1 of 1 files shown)


pawiecz (approved):
LGTM

I like the direction this patch (and many others you authored) set to explain Maestro as a finite-state machine. I think flowchart introduced by this PR could be improved to add a bit of context (e.g. putting services in subsections of Maestro patch flow) but this can be addressed in future patches.

JenySadadia:
> LGTM
> 
> I like the direction this patch (and many others you authored) set to explain Maestro as a finite-state machine. I think flowchart introduced by this PR could be improved to add a bit of context (e.g. putting services in subsections of Maestro patch flow) but this can be addressed in future patches.

I suppose you are suggesting something like what we have here https://docs.kernelci.org/maestro/pipeline/pipeline-details/. I think adding this information to that diagram would make things more complicated.

pawiecz:
> > LGTM
> > I like the direction this patch (and many others you authored) set to explain Maestro as a finite-state machine. I think flowchart introduced by this PR could be improved to add a bit of context (e.g. putting services in subsections of Maestro patch flow) but this can be addressed in future patches.
> 
> I suppose you are suggesting something like what we have here https://docs.kernelci.org/maestro/pipeline/pipeline-details/. I think adding this information to that diagram would make things more complicated.

I agree those would only make things less clear.
I had in mind this high-level overview: https://docs.kernelci.org/maestro/api/design/#pipeline-design but maybe that's redundant.

PR #793, opened 2024-09-11
fix(config): Fix "flasher" kernel for Corsola device

nuclearcat:
Corsola cannot use kernel supplied with ChromeOS build for installing modules. Use current 6.10-stable kernel for that


nuclearcat:
Tested manually at LAVA: https://lava.collabora.dev/scheduler/job/15624695

PR #800, opened 2024-09-16
src/send_kcidb: send `runtime` to `misc` field

JenySadadia:
Send runtime information to `misc` field to be used for post-processing.


JenySadadia:
Tested OK on staging:
https://kcidb.kernelci.org/d/build/build?orgId=1&var-datasource=playground&var-id=maestro:66f0b10cbca8c035cac2c1b2&var-origin=maestro&var-build_architecture=All&var-build_config_name=All&var-test_path=boot

PR #801, opened 2024-09-17
feat(errors): Improve error output

nuclearcat:
Add additional messages to help distinct between two places where particular error code might be generated.


nuclearcat:
tested on staging

PR #804, opened 2024-09-20
fix(fluster-chromeos): Adjust platforms running fluster tests

nuclearcat:
Adjust list of platforms according to requirements.


nuclearcat:
Tested: https://staging.kernelci.org:9000/viewer?search=parent%3D66ed3efa1489f1a9fa101a32

PR #805, opened 2024-09-23
config: scheduler-chromeos: Enable graphics tasts and disable some tests

musamaanjum:
See commits for description


musamaanjum:
@nuclearcat @a-wai Please review and let's merge. We want to get results on production for these as well.

PR #816, opened 2024-10-07
fix(trigger.py): Ignore commit created by user

nuclearcat:
Users might trigger tip of tree commit before trigger, ignore them and execute our own test set.


nuclearcat:
Tested on staging

PR #820, opened 2024-10-09
config: chromeos: Disable AV1 10-bit Tast tests on cherry

laura-nao:
video.PlatformDecoding.v4l2_stateless_av1_10bit* Tast tests started passing on cherry from the R130-16008.0.0 version of ChromiumOS. Disable the test for now and add reminder to revisit this at the next upgrade.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -175,6 +175,10 @@ _anchors:
       <<: *tast-debian-params
       tests:
         - video.PlatformDecoding.v4l2_stateless_av1_*
+      excluded_tests:
+        # Started passing on cherry on R130-16008.0.0; to be re-evaluated on next CrOS upgrade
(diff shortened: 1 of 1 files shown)


a-wai (approved):

a-wai:
I'm not sure if the excluded_tests mechanism deals properly with globs though, so needs a job result to ensure that works as expected.

laura-nao:
> I'm not sure if the excluded_tests mechanism deals properly with globs though, so needs a job result to ensure that works as expected.

Thanks @a-wai - wildcard apparently didn't work (e.g. https://lava.collabora.dev/scheduler/job/16076771), so I listed the tests  explicitly instead. I'll check on more staging results to confirm the changes work as expected.

laura-nao:
Seems to be working now: https://lava.collabora.dev/scheduler/job/16094894

PR #822, opened 2024-10-09
feat(send_kcidb.py): Remove log_excerpt from console logs

nuclearcat:
log_excerpt is too large and fills up truncated docker and k8s logs. Better to remove it from log output.


nuclearcat:
This is temporary solution, but we need it asap in production to reduce size of logs. Merging, tested on staging.

PR #824, opened 2024-10-10
src/send_kcidb: handle incomplete test nodes

JenySadadia:
We map incomplete test node result to `ERROR` or `MISS` based on error code set. In some cases, we
have `null` as error code. Handle such situations by setting KCIDB test status to `None` as in this
situation we can't decide if the test actually could run or not.


nuclearcat:
tested on staging

PR #825, opened 2024-10-10
src/send_kcidb: add `kbuild` failure error code

JenySadadia:
Add `kbuild_internal_error` to missed test error codes as it means that kernel build failed and test couldn't run.


nuclearcat (approved):
LGTM

PR #832, opened 2024-10-11
config: jobs-chromeos: disable power.UtilCheck test

musamaanjum:
Initially I started investigating as it was failing only on Qualcomm targets. But as I investigated and made temporary fix in hardware_probe, I found out that this test fails on all the targets. Its just that its status gets marked as pass even though there are several warnings and missing directory structures for all other boards.

Let's disable this test as it only checks the hardware info which is closely tied with the chromeos. It doesn't run correctly on chromiumos.


nuclearcat (approved):
LGTM

PR #833, opened 2024-10-11
fix(docker/k8s): Fix image names

nuclearcat:
This commit depends on PR https://github.com/kernelci/kernelci-core/pull/2705

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -3,7 +3,7 @@ _anchors:
   kbuild-gcc-12-arm64-chromeos: &kbuild-gcc-12-arm64-chromeos-job
     template: kbuild.jinja2
     kind: kbuild
-    image: kernelci/staging-gcc-12:arm64-kselftest-kernelci
+    image: ghcr.io/kernelci/{image_prefix}gcc-12:arm64-kselftest-kernelci
     params: &kbuild-gcc-12-arm64-chromeos-params
       arch: arm64
       compiler: gcc-12
@@ -22,7 +22,7 @@ _anchors:
 
   kbuild-gcc-12-x86-chromeos: &kbuild-gcc-12-x86-chromeos-job
     <<: *kbuild-gcc-12-arm64-chromeos-job
-    image: kernelci/staging-gcc-12:x86-kselftest-kernelci
+    image: ghcr.io/kernelci/{image_prefix}gcc-12:x86-kselftest-kernelci
     params: &kbuild-gcc-12-x86-chromeos-params
       arch: x86_64
       compiler: gcc-12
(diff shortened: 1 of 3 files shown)


pawiecz (dismissed):
If there's a need to be more flexible with image registries, there are two more ways to handle it:

1. make requested registry a Job param (also setting e.g. `ghcr.io` as a default)
2. explicitly include registry in the image identifier

Hiding this information to create a [magic](https://en.wikipedia.org/wiki/Magic_number_(programming)) character `/` could become confusing to the config that's not an easy one to grasp in the first place.

pawiecz:
As I went through the PR this one depends on I see that there's even more implicit changes: non-prod instances will receive different set of images (unless there's a magic character in their configured name).

Do these changes address previous "hack with hardcoding"? Sure! But they also introduce a set of new ones buried deeper in the pipeline service logic instead of its configuration/job template.

nuclearcat:
We cannot set it in pipeline configuration, as it is shared by both prod and staging instance, so it means it should be in "instance-specific" configuration file - toml file.
My suggestion is to move these hardcoded values to toml configuration file (whoever can implement it).
/ is not magic character, it is indication that registry or username is set, and if it is set - we dont use default value of ghcr.io.

pawiecz:
> We cannot set it in pipeline configuration, as it is shared by both prod and staging instance, so it means it should be in "instance-specific" configuration file - toml file.

Sure we can - pipeline configuration should include pipeline-instance-specific values.

I agree that maintaining two files and keeping them in sync might become error prone. This case could be covered by keeping a prod config in repo and patch it with sed for staging environment. This way instance configuration stays in a static file, not service logic.

> My suggestion is to move these hardcoded values to toml configuration file (whoever can implement it).

This brings another layer of config. In this case more is not merrier. I realize toml is already utilized for secrets (which obviously cannot be kept publicly) but adding more options there makes config even more confusing - like configuring Maestro Tasks through two or more layers of templates in two different repositories (pipline:config/runtime and core:config/runtime)

> / is not magic character, it is indication that registry or username is set, and if it is set - we dont use default value of ghcr.io.

True, it's not magic by itself, core#2705 makes it so. You can have images built locally containing slashes which are not bound to specific user/group/registry but you won't be able to use them because service logic will process those as if they were an external resource.

KCI image build process with tags indicating available features has been confusing enough.

I don't have the capacity to prevent all the changes that I disagree with from being introduced in the repo, especially when they address crucial issues like pulling wrong container images (here). 

Let the tech debt grow - together with debugging/maintenance "interest rate"

nuclearcat:
> I don't have the capacity to prevent all the changes that I disagree with from being introduced in the repo, especially when they address crucial issues like pulling wrong container images (here).
Actually this is why feedback is here.

If you suggest magic is no-no, i think better i remove it, and leave just basic image name. And i believe you are right on this point.
Because this behavior need to be documented, and it is not easy to do, so better to keep things simple as possible. I will update PR.
toml is used not only as secrets, but i believe it will be "instance specific" configuration file. Probably everything instance specific will move there.

PR #851, opened 2024-10-23
Disable graphics tasts

musamaanjum:
(Please check individual commits for description)


a-wai (approved):
> (Please check individual commits for description)

This makes reviews more tedious, in the future please sum things up in the PR description; no need to repeat every information contained in the commits, but the description should at least contain a summary of changes and why they're needed. In this case I would expect something along the lines of:

> A few graphics Tast tests fail on AMD devices, due to (expected) differences in behaviour when running a mainline kernel. Disable those as they only bring noise for now.

Also, "Disable failing graphics Tast tests on AMD" would be a better PR title here.

PR #852, opened 2024-10-23
config: jobs-chromeos: Turn back camera, seecurity and video tast tests

musamaanjum:


musamaanjum:
@nuclearcat let's merge this as I need results for these tests from production.

Issue #857, opened 2024-10-29
nommu_k210_defconfig config not found


musamaanjum:
stable-rc I see errors like:

*** Can't find default configuration "arch/riscv/configs/nommu_k210_defconfig"!

and

make: *** No rule to make target 'Image'.  Stop.

Which looks like something we could fix in our side?

see:

https://staging.dashboard.kernelci.org:9000/tree/d91f47825b8b0b84ea95fc066f63d46d6c6d5507?tableFilter=%7B%22buildsTable%22%3A%22invalid%22%2C%22bootsTable%22%3A%22all%22%2C%22testsTable%22%3A%22all%22%7D&origin=maestro&currentTreeDetailsTab=treeDetails.builds&diffFilter=%7B%7D&treeInfo=%7B%22gitBranch%22%3A%22linux-4.19.y%22%2C%22gitUrl%22%3A%22https%3A%2F%2Fgit.kernel.org%2Fpub%2Fscm%2Flinux%2Fkernel%2Fgit%2Fstable%2Flinux-stable-rc.git%22%2C%22treeName%22%3A%22stable-rc%22%2C%22commitName%22%3A%22v4.19.322-322-gd91f47825b8b%22%2C%22headCommitHash%22%3A%22d91f47825b8b0b84ea95fc066f63d46d6c6d5507%22%7D

https://staging.dashboard.kernelci.org:9000/tree/7d820f6c1094f9df714ce1175e1640351f026a6a?tableFilter=%7B%22buildsTable%22%3A%22invalid%22%2C%22bootsTable%22%3A%22all%22%2C%22testsTable%22%3A%22all%22%7D&origin=maestro&currentTreeDetailsTab=treeDetails.builds&diffFilter=%7B%7D&treeInfo=%7B%22gitBranch%22%3A%22linux-5.4.y%22%2C%22gitUrl%22%3A%22https%3A%2F%2Fgit.kernel.org%2Fpub%2Fscm%2Flinux%2Fkernel%2Fgit%2Fstable%2Flinux-stable-rc.git%22%2C%22treeName%22%3A%22stable-rc%22%2C%22commitName%22%3A%22v5.4.284-425-g7d820f6c1094%22%2C%22headCommitHash%22%3A%227d820f6c1094f9df714ce1175e1640351f026a6a%22%7D

PR #858, opened 2024-10-29
config: pipeline: Add minimum kernel version for nommu_k210_defconfig

musamaanjum:
The nommu_k210_defconfig config isn't present in two oldest stable kernels. Hence we get build errors. Add minimum version of the kernel to be 5.10 in which this configuration is present. I've only investigated this configuration option on stable kernels as this config is only being built on stable kernels.

close: https://github.com/kernelci/kernelci-pipeline/issues/857


nuclearcat:
Good point, thank you

PR #859, opened 2024-10-29
fix(tarball/trigger): Change tip of branch logic

nuclearcat:
Right now by some reason tip_of_branch always false, even trigger points to latest commit in branch.
Probably function in kernelci-core should be used differently? Simplify it, we always expect trigger service to watch for tip of branch commit.

### DIFF
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -82,6 +82,7 @@ class Trigger(Service):
             'kind': 'checkout',
             'data': {
                 'kernel_revision': revision,
+                'tip_of_branch': True,
             },
             'timeout': checkout_timeout.isoformat(),
             'treeid': treeid,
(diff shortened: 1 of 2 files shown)


src/trigger.py
```diff
@@ -82,6 +82,7 @@ def _run_trigger(self, build_config, force, timeout, trees):
             'kind': 'checkout',
             'data': {
                 'kernel_revision': revision,
+                'tip_of_branch': True,
```
JenySadadia, line 85: This shouldn't be set statically. We should keep it dynamic.

nuclearcat: In trigger it is by definition tip of tree:
```
        head_commit = kernelci.build.get_branch_head(build_config)
```
so it is expected that trigger.py generates only branch head which is same meaning as tip of branch.

JenySadadia:
https://github.com/kernelci/kernelci-core/pull/2734 would fix the issue in core function.
Hence, we an close this PR without merging.

nuclearcat:
Let's  test it, then i will close.

PR #863, opened 2024-10-30
feat(example_api_events.py): Add example for new events endpoint

nuclearcat:
Provide example how to use new events endpoint.


JenySadadia:
We should also have a section for this in the docs.

nuclearcat:
Where would be the best place to put this?

JenySadadia:
How about https://docs.kernelci.org/maestro/api/api-details/#pubsub-and-cloudevent ?

nuclearcat:
Merging and will prepare documentation PR for kernelci-api

PR #864, opened 2024-10-31
fix(configs): Add allmodconfig to mainline builds

nuclearcat:
Add arm,arm64,x86_64,i386 allmodconfig builds.
As per Todd Kjos it is helpful to find most of build errors.


nuclearcat:
Seems good and catch regressions on some builds already

PR #867, opened 2024-10-31
platforms-chromeos: use ChromeOS R130 for ARM64 devices

a-wai:
ARM64 ChromeBooks are now using ChromiumOS R130, so we must point the platform definitions to the new Tast tarballs. x86 devices are still on R124 though, so they must keep using the current/old URL.


a-wai:
Merging as this has been approved and numerous devices are now on R130.

a-wai:
Oops, I meant to merge #866 instead, sorry... Hopefully this shouldn't cause trouble, but feel free to revert if it does!

PR #869, opened 2024-11-05
fix(send_kcidb): Do not crash if optional field is missing

nuclearcat:
Some fields might be not present, as they are optional in model. Also add this field in custom checkout, even it is not mandatory, it is better to have it for consistency.

### DIFF
--- a/src/lava_callback.py
+++ b/src/lava_callback.py
@@ -600,7 +600,8 @@ async def checkout(data: ManualCheckout, request: Request,
                 "tree": treename,
                 "branch": branch,
                 "commit": commit,
-                "url": treeurl
+                "url": treeurl,
+                "tip_of_branch": False
             }
         },
         "timeout": checkout_timeout.isoformat(),
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -141,11 +141,11 @@ class KCIDBBridge(Service):
             'git_repository_branch':
                 checkout_node['data']['kernel_revision']['branch'],
             'git_commit_tags':
-                checkout_node['data']['kernel_revision']['commit_tags'],
+                checkout_node['data']['kernel_revision'].get('commit_tags'),
             'git_commit_message':
-                checkout_node['data']['kernel_revision']['commit_message'],
+                checkout_node['data']['kernel_revision'].get('commit_message'),
             'git_repository_branch_tip':
-                checkout_node['data']['kernel_revision']['tip_of_branch'],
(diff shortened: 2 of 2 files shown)


src/send_kcidb.py
```diff
@@ -141,11 +141,11 @@ def _parse_checkout_node(self, origin, checkout_node):
...
                 checkout_node['data']['kernel_revision']['branch'],
             'git_commit_tags':
-                checkout_node['data']['kernel_revision']['commit_tags'],
+                checkout_node['data']['kernel_revision'].get('commit_tags'),
             'git_commit_message':
-                checkout_node['data']['kernel_revision']['commit_message'],
+                checkout_node['data']['kernel_revision'].get('commit_message'),
             'git_repository_branch_tip':
-                checkout_node['data']['kernel_revision']['tip_of_branch'],
+                checkout_node['data']['kernel_revision'].get('tip_of_branch', False),
```
JenySadadia, line 148: We can keep it `None` instead of `False` if the flag is not set.

src/lava_callback.py
```diff
@@ -600,7 +600,8 @@ async def checkout(data: ManualCheckout, request: Request,
                 "tree": treename,
                 "branch": branch,
                 "commit": commit,
-                "url": treeurl
+                "url": treeurl,
+                "tip_of_branch": False
```
JenySadadia, line 604: Custom checkout can also have a commit from the tip of the branch. Why is it set `False`?

JenySadadia, line 604: Custom checkout can also have a commit from the tip of the branch. Why is it set `False`?

nuclearcat: We cannot trust user to indicate if it is TOT, but to test it separately on our side will need additional logic to implement (or fix existing one). 
We might check it later, as it is not critical right now, custom checkout events should not be sent to kcidb anyway (yet).

src/send_kcidb.py
nuclearcat: Ok

src/lava_callback.py
JenySadadia: I don't think we rely much on the fact if the commit is at the tip of branch or not. We are just using that data to send to KCIDB.
I think we can set its value from user's input and set it to `None` if not provided.

nuclearcat: It worked fine, your code updated it to true, if it is tipoftree, so we can leave as is

nuclearcat:
Tested on staging

PR #870, opened 2024-11-05
config: pipeline: Update CONFIG_FRAME_WARN for arm

musamaanjum:
CONFIG_FRAME_WARN is set to 1024 by default for arm. The allowed value for this is 0 to 8192. Let's increase it to 2048 to avoid getting the warning when stack size is more than default 1024. As all warnings are being treated as errors, our allmodconfig build fails on arm.

Close: https://github.com/kernelci/kernelci-project/issues/482

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -301,6 +301,8 @@ jobs:
       defconfig:
         - defconfig
         - allmodconfig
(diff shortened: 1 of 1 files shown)


nuclearcat:
LGTM, staging works

PR #873, opened 2024-11-08
kube/aks/pipeline-kcidb: add `KCIDB_POSTGRESQL_PASSWORD` env

JenySadadia:
Add secret env variable for KCIDB postgresql password used to connect to KCIDB database to fetch existing issues.

### DIFF
--- a/kube/aks/pipeline-kcidb.yaml
+++ b/kube/aks/pipeline-kcidb.yaml
@@ -32,6 +32,11 @@ spec:
                   key: token
             - name: GOOGLE_APPLICATION_CREDENTIALS
               value: /secrets/kcidb-credentials.json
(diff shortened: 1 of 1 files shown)


nuclearcat (changes requested):
Maybe we can use toml variable? Anyway toml file keep secrets, because environment variable deployment will require also to create this key in deployment scenario, and we are trying to do opposite, reduce number of environment variables and move all possible to toml file.

JenySadadia:
Sure, I'll update it accordingly.

JenySadadia:
Dropping this PR as KCIDB postgresql password will be fetched from TOML file instead with the changes introduced in https://github.com/kernelci/kernelci-pipeline/pull/880.

Issue #878, opened 2024-11-08
Limit the tast tests to run only on stable and mainline kernels


musamaanjum:
cc: @nuclearcat @padovan

PR #879, opened 2024-11-08
config: pipeline: Run tast test only on mainline and stable-rc only

musamaanjum:
Let's reduce the number of trees on which we execute tast tests. The stable-rc and mainline are of interest to us only. This'll help us getting cleaner data and devices would be more free to execute other tests.

Close: https://github.com/kernelci/kernelci-pipeline/issues/878

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -58,6 +58,10 @@ _anchors:
     kind: job
     rules: *min-5_4-rules
     kcidb_test_suite: tast
+    rules:
+      tree:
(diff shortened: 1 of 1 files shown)


config/jobs-chromeos.yaml
```diff
@@ -58,6 +58,10 @@ _anchors:
     kind: job
     rules: *min-5_4-rules
```
a-wai, line 59: Ahem... Either replace or extend current rules, but don't add duplicate entries ;)

PR #880, opened 2024-11-11
src/send_kcidb: use kcidb postgresql password from TOML

JenySadadia:
In order to avoid using environment variables, use TOML file for KCIDB postgresql DB user password. Also, declare other existing args with `help` options in `cmd_run`.


nuclearcat (approved):
LGTM

PR #881, opened 2024-11-12
config: update fluster debian test mapping

JenySadadia:
Update fluster debian tests KCIDB mapping to use `fluster.<file-system>.v4l2.*` syntax to unify the pattern with chromeos fluster test mapping.


JenySadadia:
Tested OK with manual submission of job node to staging:
https://old-kcidb-grafana-415818378487.us-central1.run.app/d/test/test?orgId=1&var-datasource=playground&var-origin=All&var-build_architecture=All&var-build_config_name=All&var-id=maestro:66e928c7e7e294ee7a240893

PR #882, opened 2024-11-12
fix(k8s/postgres-proxy): Fix deployment of postgres-proxy

nuclearcat:
We need to define service, to be able to contact it by FQDN.


nuclearcat:
Tested on deployment

PR #885, opened 2024-11-13
src/logspec: fix version field for KCIDB issue

JenySadadia:
Fix version field name for existing KCIDB issue.


JenySadadia:
Merging this to fix issue submissions.

PR #886, opened 2024-11-13
KCIDB: Send `misc.measurement` for tast tests

JenySadadia:
Tast test `os-release` will contain chromeos version information in `node.data.misc.measurement` field.
Send the information to KCIDB.

### DIFF
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -393,6 +393,7 @@ the test: {sub_path}")
             compatible = self._platforms[platform].compatible
 
         runtime = test_node['data'].get('runtime')
+        misc = test_node['data'].get('misc')
         parsed_test_node = {
             'build_id': build_id,
             'id': f"{origin}:{test_node['id']}",
@@ -405,6 +406,7 @@ in {runtime}",
                 'compatible': compatible,
                 'misc': {
                     'platform': platform,
+                    'measurement': misc.get('measurement') if misc else None
                 }
             },
             'waived': False,
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -389,6 +389,7 @@ the test: {sub_path}")
             build_id = f"{origin}:{build_node['id']}"
 
         platform = test_node['data'].get('platform')
+        compatible = None
         if platform:
             compatible = self._platforms[platform].compatible
 


JenySadadia:
Hello @laura-nao 
Sorry, I missed that we don't send `setup` tests to KCIDB. So, `os-release` won't be submitted to it.
If you're still interested in getting `or-release` test on the dashboard, I can enable its submission.
I would like to know your specific requirements.

laura-nao:
> Hello @laura-nao Sorry, I missed that we don't send `setup` tests to KCIDB. So, `os-release` won't be submitted to it. If you're still interested in getting `or-release` test on the dashboard, I can enable its submission. I would like to know your specific requirements.

Thanks for the clarification @JenySadadia - the original request was part of a broader set of features for the Codec dashboard. Currently, the ChromeOS version isn’t reported anywhere else, and this information is helpful when determining whether certain Tast test failures or skipped tests are expected after enablement. For external users, the only alternative is manually checking the LAVA logs, which isn’t very convenient.

That said, if implementing this feature turns out to be too hacky or deviates significantly from how other tests are handled, I think we can manage without it. Since the information is already available in the test, it seemed reasonable to include it in the results sent to KCIDB. Let me know if you think the implementation would be too tricky, and we can explore other ways to report the version.

JenySadadia:
@laura-nao Thanks for explaining the requirements.

I can adjust in maestro to send the specific tests you are interested in and that won't be too tricky.
Currently, we have `os-release` and `tast-tarball` in `setup` test set for tast tests. 
Do you just require `os-release` test submissions to KCIDB or `tast-tarball` as well?

laura-nao:
> Do you just require `os-release` test submissions to KCIDB or `tast-tarball` as well?

`os-release` will be enough, thanks!

JenySadadia:
> > Do you just require `os-release` test submissions to KCIDB or `tast-tarball` as well?
> 
> `os-release` will be enough, thanks!

Okay. Thanks for confirming.

JenySadadia:
Tested OK on staging.
See the test node with `environment.misc.measurement` value:
https://kcidb.kernelci.org/d/test/test?var-datasource=edqupth5xi8e8b&var-origin=$__all&var-build_architecture=arm64&var-build_config_name=cros:%2F%2Fchromeos-6.6%2Farm64%2Fchromiumos-qualcomm.flavour.config&var-id=maestro:673a1e99dbf955cdda3d82c8&from=now-100y&to=now&timezone=browser&var-test_path=&var-issue_presence=$__all

laura-nao (approved):

Issue #888, opened 2024-11-14
Several trees have incorrect settings or unable to checkout


nuclearcat:
According the log, instead of commit id we have for some trees: False, which means trigger is unable to check them.
http://mon.kernelci.org:8080/k8s-api/logs/2024-11-14-trigger-55cf967d69-xswrh-kernelci-pipeline.log
We need to check if there is any mistake in configuration

PR #890, opened 2024-11-15
config: pipeline: Fix failures caused by unsuccessful checkouts

musamaanjum:
- Some branches have disappeared. Remove those branches.
- Some branches have been renamed.

There was for-kernelci branch present in peterz's tree which has disappeared. I'm replacing some other branch for now until we get a reply from him.

Close: https://github.com/kernelci/kernelci-pipeline/issues/888

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -113,7 +113,6 @@ _anchors:
     - 'linusw'
     - 'net-next'
     - 'next'
-    - 'peterz'
     - 'pm'
     - 'renesas'
     - 'robh'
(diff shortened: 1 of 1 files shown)


config/pipeline.yaml
```diff
@@ -3071,10 +3071,6 @@ build_configs:
     tree: amlogic
     branch: 'for-next'
 
-  amlogic_integ:
-    tree: amlogic
-    branch: 'integ'
```
musamaanjum, line 3076: Sent email asking the Neil about it. Let's wait before merging the PR.

config/pipeline.yaml
```diff
@@ -3345,7 +3333,7 @@ build_configs:
 
   peterz:
     tree: peterz
-    branch: 'kernelci'
+    branch: 'perf/core'
```
musamaanjum, line 3336: Sent email asking the Peter about it. Let's wait before merging the PR.

musamaanjum: Sent reminder

nuclearcat: I think remove tree and branches for now, if we get feedback we will add tree back

musamaanjum: Sure.

PR #892, opened 2024-11-18
config: jobs-chromeos: Remove LowPowerStateResidence

musamaanjum:
This tests expects that the disk activity would be less 50%. But it is hard to guarantee this even in normal circumstances. I've confirmed that even in sanely booted chromiumos, there is disk activity. I've tried discussing it with the owner of this test, but no conclusion there as well. I've discussed with Denis and he agrees and suggest that we should disable this test.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -411,7 +411,6 @@ _anchors:
         - platform.Mtpd
         - platform.TPMResponsive
         - storage.HealthInfo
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
LGTM

a-wai (commented):
LGTM, just watch out for typos in the commit title and message:
* LowPowerS***t***ateResidence
* g***u***arantee

PR #807, opened 2024-09-23
config: Add sashal/linus-next:linus-next tree

pawiecz:
Fixes: kernelci/kernelci-core#2686

PR #897, opened 2024-11-19
v4l2-decoder: Remove media tree monitoring, add stable-rc_6.6

denisyuji:
As per internal discussion (ref: https://issuetracker.google.com/issues/371638530 ), we will not monitor `media` branch for fluster/tast v4l2 decoder tests neither for ChromeOS or Debian rootfs. We will keep monitoring `next` branch for fluster tests for now and evaluate its value later. In addition, ChromeOS will stay in `stable-rc_6.6` and eventually `stable-rc_6.12`, so we will keep monitoring this branch.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -61,7 +61,8 @@ _anchors:
       <<: *min-5_4-rules
       tree:
         - mainline
-        - stable-rc
+        - collabora-chromeos-kernel
+        - stable
 
   tast-debian: &tast-debian-job
     template: 'generic.jinja2'
@@ -77,9 +78,8 @@ _anchors:
     rules:
       tree:
         - mainline
-        - next
         - collabora-chromeos-kernel
-        - media
+        - stable
 
   tast-basic: &tast-basic-job
     <<: *tast-job
@@ -464,7 +464,6 @@ _anchors:
         - mainline
         - next
         - collabora-chromeos-kernel
-        - media
 
   fluster-chromeos: &fluster-chromeos-job
     template: 'fluster-chromeos.jinja2'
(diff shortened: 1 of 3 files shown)


laura-nao:
@denisyuji from my understanding, the codec tests are going to target the 6.6 LTS (and the 6.12 LTS later). 

As far as I can see we're not tracking the 6.6 LTS tree/branch at the moment (that's different from stable-rc). In that case, I think we should start tracking the `linux-6.6.y` branch of https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git . That can be done by adding the URL to the `trees:` section of the `pipeline.yaml` file, and then adding a `stable_6.6` build configuration to track the `linux-6.6.y` branch (similar to what has been done for stable-rc).
I think that should be enough to start tracking the 6.6 LTS in KernelCI, right @nuclearcat ?

nuclearcat:
Yes, here is example: https://github.com/kernelci/kernelci-pipeline/pull/807

denisyuji:
@laura-nao, @nuclearcat , thank you for your inputs. Please, see the changes now and let me know if it looks good now.

nuclearcat (approved):
LGTM

laura-nao:
@nuclearcat did you see the changes I suggested above? With this patch as-is, _all_ Tast tests will run on mainline, stable and collabora-chromeos-kernel, not just the codec tests..which is not what we wanted. 
I believe we should keep Tast tests running on mainline and stable-rc by default and change the behavior for codec tests only.

nuclearcat:
@denisyuji can you fix that please?

denisyuji:
Thank you @laura-nao . That makes sense. I fixed it in a new PR: https://github.com/kernelci/kernelci-pipeline/pull/939

PR #898, opened 2024-11-19
src/send_kcidb: add new labs to LAVA labs mapping

JenySadadia:
Add recently enabled labs to LAVA lab mapping of lab name and URL to KCIDB bridge service. 
Also, add a warning print if lab URL mapping is not found for the specific lab.


nuclearcat (approved):
LGTM

PR #907, opened 2024-11-22
add(pipeline.yaml): Add allmodconfig to stable-rc builds

nuclearcat:
As suggested by Todd Kjos allmodconfig greatly help to find build errors and highly useful for stable-rc builds.


nuclearcat:
Tested on staging

PR #908, opened 2024-11-22
config: jobs-chromeos: Disable always failing tast tests on specific targets

musamaanjum:
Disable tests which have always been failing.
- The graphics stack for Mediatek is different on proprietary ChromeOS. Hence the test don't work on open-source ChromeOS. Disable all these test.
- Also disable all those tests which always fail because of missing dependencies. This includes camera clvk tests etc.


nuclearcat:
LGTM

PR #909, opened 2024-11-25
feat(pipeline.yaml): Add more k8s clusters for builds

nuclearcat:
As legacy builds disabled, we can use more build clusters.


nuclearcat:
Tested on staging, with new feature for better load balancing

PR #911, opened 2024-11-25
docker/kcidb: fix logspec to a specific version

helen-fornazier:
Some cases will require the kcidb bridge to interpret results from logspec before creating issues and incidents. Fix logspec to a specific commit so we can control when to upgrade it. This allow us to upgrade logspec in the same commit we upgrade kcidb bridge to keep them in sync.


JenySadadia (approved):

helen-fornazier:
@nuclearcat o/ please review and merge this if possible

PR #913, opened 2024-11-26
fix(trigger.py): Trigger should ignore incomplete nodes

nuclearcat:
As we have limited query language, count also incomplete nodes. If particular commit nodes are all incomplete - force checkout again, as checkout was never completed (which happen quite often).

### DIFF
--- a/src/trigger.py
+++ b/src/trigger.py
@@ -44,14 +44,26 @@ class Trigger(Service):
                 return
 
         head_commit = kernelci.build.get_branch_head(build_config)
-        node_count = self._api.node.count({
+        search_terms = {
             "kind": "checkout",
             "data.kernel_revision.commit": head_commit,
             "owner": self._current_user['username'],
             "submitter": "service:pipeline"
-        })
+        }
+        node_count = self._api.node.count(search_terms)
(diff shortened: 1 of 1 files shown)


src/trigger.py
```diff
@@ -50,8 +50,24 @@ def _run_trigger(self, build_config, force, timeout, trees):
             "owner": self._current_user['username'],
             "submitter": "service:pipeline"
         })
+        incomplete_node_count = self._api.node.count({
+            "kind": "checkout",
+            "data.kernel_revision.commit": head_commit,
+            "owner": self._current_user['username'],
+            "submitter": "service:pipeline",
```
JenySadadia, line 57: We can have a common `dict` with the below information and use it for calculating both `node_count` and `incomplete_node_count` (extend it with `result` field in this case).
```suggestion
       {
            "kind": "checkout",
            "data.kernel_revision.commit": head_commit,
            "owner": self._current_user['username'],
            "submitter": "service:pipeline",
```

nuclearcat: done

JenySadadia (approved):
LGTM.

Issue #910, opened 2024-11-25
Create LAVA labs mapping for viewer and kcidb bridge in config file


nuclearcat:
https://github.com/kernelci/kernelci-pipeline/pull/898#pullrequestreview-2458015874
We must try to avoid hardcoding such values in source code.

JenySadadia:
Yes, we need to have dynamic mapping at some point.
One way to achieve it is to read `runtimes` from `pipeline.yaml` and create a `dict` dynamically. All the runtimes can be fetched with `configs['runtimes']` in `KCIDBBridge` class just like we are fetching `jobs` and `platforms` atm.

PR #915, opened 2024-11-27
src/send_kcidb: get lava labs from config

JenySadadia:
Closes https://github.com/kernelci/kernelci-pipeline/issues/910

Fetch available LAVA runtimes dynamically instead of relying on a static lava labs mapping dictionary.


pawiecz (approved):
LGTM

PR #919, opened 2024-11-29
feat(send_kcidb.py): Set processed_by_kcidb_bridge flag after processing node

nuclearcat:
Requires: https://github.com/kernelci/kernelci-core/pull/2746 

After node is being processed, set flag to indicate we sent data to kcidb.

### DIFF
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -17,6 +17,7 @@ import io
 import gzip
 from urllib.parse import urljoin
 import requests
+import time
 
 import kernelci
 import kernelci.config
@@ -57,6 +58,7 @@ class KCIDBBridge(Service):
         self._jobs = configs['jobs']
         self._platforms = configs['platforms']
         self._lava_labs = {}
+        self._last_unprocessed_search = None
         for runtime_name, runtime_configs in configs['runtimes'].items():
             if isinstance(runtime_configs, RuntimeLAVA):
                 self._lava_labs[runtime_name] = runtime_configs.url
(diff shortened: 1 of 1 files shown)


src/send_kcidb.py
```diff
@@ -570,6 +573,14 @@ def _run(self, context):
                 }
             }
             self._send_revision(context['client'], revision)
+
+            # mark the node as processed, sent_kcidb field to True
+            # TBD: job nodes might have child nodes, mark them as processed as well
+            node['sent_kcidb'] = True
+            try:
+                self._api.node.update(node)
```
JenySadadia, line 581: I think the safe place to update the node is in `_send_revision` method after data validation is done with `kcidb.io.SCHEMA.is_valid(revision)`. As for some reason, if data is invalid, the service would still set the `kcidb_sent` flag to `True`.

src/send_kcidb.py
```diff
@@ -570,6 +573,14 @@ def _run(self, context):
                 }
             }
             self._send_revision(context['client'], revision)
+
+            # mark the node as processed, sent_kcidb field to True
+            # TBD: job nodes might have child nodes, mark them as processed as well
+            node['sent_kcidb'] = True
```
JenySadadia, line 579: I believe the field name is `kcidb_sent` from PR https://github.com/kernelci/kernelci-core/pull/2746/.

src/send_kcidb.py
nuclearcat: I might move there, but may data become valid later?
Because if i will not flag node as processed, it will keep trying to resend it (and fail, because data is invalid).
Maybe better we just report invalid data somewhere.

src/send_kcidb.py
JenySadadia: Sure, np. I'd be happy to help in case you find something unclear in the bridge service.

src/send_kcidb.py
JenySadadia: Data becomes invalid if incorrect type of data is stored in node or there is some parsing issue while generating KCIDB-compatible data dictionary in the bridge service.
That's why the same data won't be valid again without any changes to service or node model.
Maybe we can have an error log for invalid data. Also, we can have a `dict` field in `Node` model for KCIDB related processing information with 2 flags i.e. `kcidb_sent` and `processed_by_kcidb_bridge`, and use both the fields for a submission retry.

nuclearcat: Then maybe i will change name to processed_by_kcidb_bridge (but it feels a bit long), it will be more correct such way.

src/send_kcidb.py
```diff
@@ -486,12 +488,61 @@ def _get_test_data_recursively(self, node, origin, parsed_test_node, parsed_buil
...
         self.log.info("Listening for events... ")
         self.log.info("Press Ctrl-C to stop.")
 
         while True:
-            node, is_hierarchy = self._api_helper.receive_event_node(context['sub_id'])
+            is_hierarchy = False
+            node = self._find_unprocessed_node()
+
+            if not node:
+                node, is_hierarchy = self._api_helper.receive_event_node(context['sub_id'])
```
JenySadadia, line 545: I think we will miss PubSub events if this is conditional as this will only process real-time event data if no unprocessed nodes are found.

src/send_kcidb.py
```diff
@@ -486,12 +488,61 @@ def _get_test_data_recursively(self, node, origin, parsed_test_node, parsed_buil
...
                                                 parsed_build_node)
 
+    def _node_processed(self, node):
+        """
+        Mark the node as processed, sent_kcidb field to True
+        That means kcidb has received the node, and at least tried to process it
+        Data in node might be invalid, and not really sent to kcidb
+        This is workaround, until we improve event handling in kernelci-pipeline/api
+        """
+        node['processed_by_kcidb_bridge'] = True
```
JenySadadia, line 498: I though we were going to have 2 flags, i.e. one is for successful submission and another one is for processing.

src/send_kcidb.py
```diff
@@ -486,12 +488,61 @@ def _get_test_data_recursively(self, node, origin, parsed_test_node, parsed_buil
...
+            self.log.error(f"Failed to update node {node['id']}: {str(exc)}")
+
+    def _node_processed_recursively(self, node):
+        """
+        Mark the node and its child nodes as processed
+        """
+        child_nodes = self._api.node.find({'parent': node['id']})
+        if child_nodes:
+            for child in child_nodes:
+                self._node_processed_recursively(child)
```
JenySadadia, line 511: Instead of querying API, we should be getting node IDs from `revision` data as there might be some cases when not all the child nodes are submitted to KCIDB.
So, mark `sent_kcidb` to `True` only if the node was actually submitted.

src/send_kcidb.py
nuclearcat: As i recall we tested before together, active pubsub subscription will queue events and wont drop them.
If it doesnt(if i missed something), then it need to be redesigned, because even in current code, it means it might miss event, if other event submission going too slow.

src/send_kcidb.py
nuclearcat: for submission not yet, as it will be next step of handling errors, which is unclear if happening at all

src/send_kcidb.py
JenySadadia: Okay. This needs to be verified.

src/send_kcidb.py
nuclearcat: At the moment we mark just "processed by kcidb bridge", to workaround missing events issue.

JenySadadia: Okay. So, `sent_to_kcidb` would be a different flag for successful submission. Got it.

PR #920, opened 2024-11-29
src/send_kcidb: don't send timed-out nodes to KCIDB

JenySadadia:
Disable timed-out checkout, build, job, and test nodes submission to KCIDB for the time being.
This needs to be tweaked when maestro has job retry logic in place.


nuclearcat (approved):
LGTM, as temporary solution will resolve our problem and gradually i will try to improve it

PR #923, opened 2024-12-02
fix(send_kcidb.py): Submit incomplete test nodes

nuclearcat:
It seems due LAVA details incomplete nodes might contain useful data.


nuclearcat:
Tested on staging

PR #924, opened 2024-12-02
src/logspec_api: set code_culprit True

helen-fornazier:
After reviewing the current version of logspec we are using, all the errors parsed are of type code, so set culprit_code: True

### DIFF
--- a/src/kernelci_pipeline/logspec_api.py
+++ b/src/kernelci_pipeline/logspec_api.py
@@ -114,6 +114,13 @@ def new_issue(logspec_error, object_type):
         'comment': comment,
         'misc': {
             'logspec': error_copy
+        },
+        # Set culprit_code to True by default
+        # OBS: this needs to be reviewed on every logspec upgrade
+        'culprit': {
+            'code': True,
+            'harness': False,
+            'tool': False
         }
     }
     if 'build_valid' in object_types[object_type]:


helen-fornazier:
Successfully tested with a local instance of kernelci-pipeline

submitted to KCIDB playgroud, see https://grafana.kernelci.org/d/issue/issue?var-version=0&var-id=maestro:bba672c389cd8e575694e468d87a80b1b4074241&var-datasource=edqupth5xi8e8b

src/kernelci_pipeline/logspec_api.py
```diff
@@ -114,7 +114,10 @@ def new_issue(logspec_error, object_type):
         'comment': comment,
         'misc': {
             'logspec': error_copy
-        }
+        },
+        # Set culprit_code to True by default
+        # OBS: this needs to be reviewed on every logspec upgrade
+        'culprit': {'code': True}
```
spbnick, line 120: You also gotta set the rest of the members to false. Otherwise their status is considered unknown. That is here you're saying: "I know this is a kernel issue, but it also might be a test issue or a CI issue at the same time".

spbnick (changes requested):

src/kernelci_pipeline/logspec_api.py
helen-fornazier: Fixed, please check https://grafana.kernelci.org/d/issue/issue?var-version=0&var-id=maestro:bba672c389cd8e575694e468d87a80b1b4074241&var-datasource=edqupth5xi8e8b

spbnick (approved):

PR #927, opened 2024-12-03
upd(send_kcidb.py): Update reprocess time window to 4 days

nuclearcat:
We might have hiccups we notice too late, so increase time window to reprocess data.


nuclearcat:
Tested on staging

PR #932, opened 2024-12-04
fix(send_kcidb.py): Limit nodes search to 1

nuclearcat:
If there is too many nodes it will significantly impact performance.


nuclearcat:
Tested on staging

PR #933, opened 2024-12-04
fix(send_kcidb.py): Use faster findfast function

nuclearcat:
find() is paginated function using .count that might cause significant load on API server. Use more efficient findfast() call.


nuclearcat:
Tested on staging

Issue #925, opened 2024-12-03
LAVA Runtimes selection in Kubernetes deployment recipes for Production instance is out of sync


pawiecz:
Several LAVA Runtimes added to the Pipeline configuration were not added to the Kubernetes deployment recipes:

- 'lava-collabora-staging'
- 'lava-qualcomm'
- 'lava-cip'

pawiecz:
#937 had to be reverted due to the lack of `lab_type` filter set for the LAVA Runtime Scheduler

nuclearcat:
Closing as completed. Docker Compose and Kubernetes runtime selection were aligned in [c73fb11](https://github.com/kernelci/kernelci-pipeline/commit/c73fb11), then generalized to select configured runtimes by type in [65a61df](https://github.com/kernelci/kernelci-pipeline/commit/65a61df4e65331700e4ad29b12bafe4a4fa10acd).

PR #937, opened 2024-12-04
Drop Runtimes filter on LAVA Scheduler in Kubernetes deployment

pawiecz:
Scheduler is able to filter out Runtimes it will submit Jobs to with a "--runtimes" option. The filter currently in use for Kubernetes deployment got out-of-sync with the LAVA Runtimes enabled in the Pipeline config.

This patch drops this option to fall back to the default behaviour which is submitting Jobs to all the Runtimes available in the Pipeline config.

Fixes: #925


nuclearcat:
Can't test on staging, so let's hope it works on prod update :)

PR #938, opened 2024-12-04
fix(send_kcidb.py): Set flag by bulk operation

nuclearcat:
1)Implement file:// schema in logspec_api
2)Implement log file caching locally to reduce storage
call delays and reduce egress costs
3)Caching of excerpt logs to reduce storage retrievals
4)Caching of nodes to reduce API calls
5)Requests batching with fallback in number in case
submission size is too big.

This is emergency update, as we have performance issue,
so everything is crammed into single commit.


nuclearcat:
Tested on staging

PR #940, opened 2024-12-05
src/logspec_api: re-submit issues instead of querying

helen-fornazier:
For KCIDB it is better to re-submit the same object with the same id instead of querying to check if it exist already.

Remove the query and always submit a new issue.

Tested with local instance of kernelci-pipeline, submitting to KCIDB playground.

### DIFF
--- a/src/kernelci_pipeline/logspec_api.py
+++ b/src/kernelci_pipeline/logspec_api.py
@@ -177,15 +177,6 @@ def process_log(log_url, parser, start_state):
     return get_logspec_errors(parsed_data, parser)
 
 
-def get_issue_from_db(oo_client, signature):
-    """Get existing issue matching ID from DB"""
-    data = oo_client.query(kcidb.orm.query.Pattern.parse(f">issue[maestro:{signature}]#"))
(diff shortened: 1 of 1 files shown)


src/kernelci_pipeline/logspec_api.py
```diff
@@ -188,15 +179,10 @@ def generate_issues_and_incidents(result_id, log_url, object_type, oo_client):
...
-            if issue:
-                issue_id = issue.id
-                issue_version = issue.version_num
-            else:
-                issue = new_issue(error, object_type)
-                issues.append(issue)
-                issue_id = issue["id"]
-                issue_version = issue["version"]
+            issue = new_issue(error, object_type)
+            issues.append(issue)
```
spbnick, line 183: Could it happen that you would submit an array containing duplicate issues to KCIDB here? Not a correctness problem, but would be a waste of resources.

helen-fornazier: yes, good catch, I just added another commit to this PR that remove duplicated issues, please check

PR #943, opened 2024-12-05
fix(send_kcidb.py): Tune a bit workaround processing

nuclearcat:
We need to have chance to process events.
If last search returned less than max nodes, this means we got
all pending nodes, switch to events mode.


nuclearcat:
Tested on staging

PR #956, opened 2024-12-24
Various fixes 12 2024

nuclearcat:
1)pod memory allocation
2)scheduler fix due missing keys in node
3)trigger fix due tree returning tree during head retrieval


nuclearcat:
Tested on staging

PR #960, opened 2025-01-01
wip: Use just "qemu" as a platform name for x86 target

pawiecz:


nuclearcat:
I will put skip-staging as it conflict with enablement of aaeon

PR #962, opened 2025-01-08
config: pipeline: Limit target trees/branches for Chromebook tests

laura-nao:
Currently, most tests on Chromebooks run across all available tree/branch combinations. Restrict these tests to stable-rc and mainline trees to simplify monitoring while still identifying relevant regressions.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -491,6 +491,10 @@ _anchors:
       test_method: watchdog-reset
       bl_message: 'coreboot-'
       wdt_dev: 'watchdog0'
+    rules:
+      tree:
+        - mainline
+        - stable-rc
     kcidb_test_suite: kernelci_watchdog_reset
 
   wifi-basic: &wifi-basic-job
@@ -503,7 +507,7 @@ _anchors:
     rules:
       tree:
         - mainline
-        - next
+        - stable-rc
     kcidb_test_suite: kernelci_wifi_basic
 
 jobs:
@@ -512,6 +516,10 @@ jobs:
     template: baseline.jinja2
     kind: job
     kcidb_test_suite: boot
+    rules:
+      tree:
+        - mainline
+        - stable-rc
 
   baseline-arm64-qualcomm-chromebook: *baseline-job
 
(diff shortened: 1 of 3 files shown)


laura-nao:
@nuclearcat @pawiecz feel free to drop the `staging-skip` label once you've had a chance to review and confirm the changes look good. Thanks!

pawiecz (approved):
LGTM, let's drop `skip-staging` and give it a test run before it can be deployed in prod

laura-nao:
> LGTM, let's drop `skip-staging` and give it a test run before it can be deployed in prod

Thanks! I checked the results on staging - I see results on mainline/master, e.g.:
- https://staging.kernelci.org:9000/viewer?search=parent%3D67817c1f239cdda8d073822a
- https://staging.kernelci.org:9000/viewer?search=parent%3D6781418476323fe6ba19815e

And I don't see results on the kernelci tree anymore, e.g.:
- https://staging.kernelci.org:9000/viewer?search=parent%3D6784602eae836f7c89991daf

So I guess this worked as expected. Now that I saw the results on staging, I wonder if we should keep running the tests on the kernelci tree as well though? otherwise they will just be available on the mainline tree on staging. Any thought on this @pawiecz @nuclearcat ?

pawiecz:
> So I guess this worked as expected.

:+1:

> Now that I saw the results on staging, I wonder if we should keep running the tests on the kernelci tree as well though? otherwise they will just be available on the mainline tree on staging. Any thought on this @pawiecz @nuclearcat ?

Current default behavior on staging is to [run tests from `kernelci/*` trees only](https://github.com/kernelci/kernelci-pipeline/blob/9c2baf69ff05807253b94a57eca745d73900f10e/docker-compose.yaml#L147). Linked test runs on `mainline/master` were triggered by `@nuclearcat` manually.

If you'd like to make Chromebook tests passing one of the Maestro PRs merge criteria then `kernelci` tree should be added to the triggering rules.

If you're confident about these test suites' stability and don't plan on changing them soon - filtering only mainline + stable-rc triggers should be fine (but it will require manual trigger on staging, if needed).

nuclearcat:
Usually if something might affect non-kernelci tree tests i trigger that tree manually, so it is not necessary to add tests to kernelci tree.

laura-nao:
Ack, sounds good and should be ready to merge then. Thanks both for the feedback!

PR #963, opened 2025-01-10
upd(pipeline.yaml): new kselftest rootfs 

nuclearcat:
Test new rootfs with some packages enabled for netdev


nuclearcat:
tested on staging

PR #965, opened 2025-01-14
upd(pipeline.yaml): Enable kselftest-net collection

nuclearcat:
Now, as final step we try to enable kselftest-net collection. It might require future adjustments, as kernel options configuration.


nuclearcat:
Tested on staging: https://staging.kernelci.org:9000/viewer?search=parent%3D6786a260029aeabb05bfc54c

PR #966, opened 2025-01-14
fix(*.yaml): Add to event kind: kbuild

nuclearcat:
Some events have not specified kind, just kbuild-**somename**, as result when regression generated, with same name as kbuild event, it will trigger for example LAVA job with undesirable results.


nuclearcat:
Tested on staging

PR #967, opened 2025-01-15
Enable baseline test for Genio 700 and 1200 EVK boards

nfraprado:
This PR describes the Genio 700 and 1200 EVK boards and enables the baseline test for them.

Additionally, it adds the 'lab-setup' configuration fragment to the `kbuild-gcc-12-arm64-job` in order to avoid deferred probe timeout issues that make the boot fail on those boards. The fragment is added in the generic arm64 build since it should avoid similar issues on other platforms as well.

Genio 700 EVK boot on linux-next: https://lava.collabora.dev/scheduler/job/17371995
Genio 1200 EVK boot on linux-next: https://lava.collabora.dev/scheduler/job/17372467

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -2104,6 +2104,12 @@ platforms:
   # No job is being scheduled on this board as its infrastructure errors need to be fixed first.
   minnowboard-turbot-E3826: *x86_64-device
 
+  mt8390-genio-700-evk:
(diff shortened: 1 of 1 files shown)


nuclearcat:
Boards booted! https://staging.kernelci.org:9000/viewer?search=parent%3D67895303f96facda670a78d4

PR #972, opened 2025-01-22
src/{send_kcidb,logspec_api}: report unclean boot as PASS

helen-fornazier:
If the kernel booted, but maestro had set that to FAIL, change it to
PASS and create an issue with WARNING.


nuclearcat (approved):
Looks fine on staging, merging

PR #974, opened 2025-01-24
 feat(example_api_lab.py): Add example for labs how to submit result to Maestro

nuclearcat:
In some situations we have failures that are invible from logs. We need better logs.

### DIFF
--- a/tools/example_api_lab.py
+++ b/tools/example_api_lab.py
@@ -1,0 +1,180 @@
+#!/usr/bin/env python3
+"""
+KernelCI API example
+3rd party lab submission, when lab is not integrated with KernelCI pipeline
+
+"""
+
+import requests
+import json
+import sys
+
+# KernelCI API URL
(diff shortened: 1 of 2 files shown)


tools/example_api_lab.py
```diff
@@ -0,0 +1,179 @@
...
+        "artifacts": {},
+        "data": {},
+        "submitter": parent_node["submitter"],
+        "treeid": parent_node["treeid"],
+    }
+    job_node["data"]["kernel_revision"] = parent_node["data"]["kernel_revision"]
+    return create_node(job_node)
+
+
+def job_results(job_node_id, tests, artifacts):
```
padovan, line 86: maybe call this `submit_job_results()`. to make it clear?

padovan (commented):
It looks good!

tools/example_api_lab.py
nuclearcat: done

PR #975, opened 2025-01-24
Enable wifi-basic test on the Genio 700 EVK

nfraprado:
Depends on https://github.com/kernelci/kernelci-core/pull/2786

Example run: https://lava.collabora.dev/scheduler/job/17513959

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -1895,6 +1895,19 @@ jobs:
         - media
     kcidb_test_suite: h26forge.debian
 
+  wifi-basic:
+    template: generic.jinja2
+    kind: job
+    params:
+      test_method: wifi-basic
+      boot_commands: nfs
+      nfsroot: 'https://storage.kernelci.org/images/rootfs/debian/bookworm-wifi/20240313.0/{debarch}/'
+    rules:
+      tree:
+        - mainline
+        - stable-rc
+    kcidb_test_suite: kernelci_wifi_basic
+
 trees:
 
   amlogic:
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -2304,6 +2304,14 @@ scheduler:
       - rk3588-rock-5b
       - sun50i-h6-pine-h64
 
+  - job: wifi-basic
+    event:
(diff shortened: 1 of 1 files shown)


nfraprado:
There's already a `kbuild-gcc-12-arm64-chromebook` job in chromeos-jobs.yaml, so I've renamed the one introduced here to `kbuild-gcc-12-arm64-chromebook-main` to avoid the CI failures. Let me know if you want to fix the issue in a different way.

config/pipeline.yaml
```diff
@@ -926,6 +926,15 @@ jobs:
 
   kbuild-gcc-12-arm64: *kbuild-gcc-12-arm64-job
 
+  kbuild-gcc-12-arm64-chromebook-main:
```
pawiecz, line 929: It would make sense to me to have this one a plain "-chromebook" and the other one "-chromeos" (since it's in "*-chromeos.yaml" configs).

We can merge it as it is and do the rename as a next step (here, in jobs- and scheduler- files).

config/pipeline.yaml
```diff
@@ -1895,6 +1895,19 @@ jobs:
         - media
     kcidb_test_suite: h26forge.debian
 
+  wifi-basic:
```
pawiecz, line 1898: This job definition is not anchored anywhere in the main file - did you encounter any failures in processing scheduler entries with this job definition when it's read from "jobs-chromeos.yaml"?

nfraprado: I'm not set up to run the kci tools here, so I'm relying on the CI checks done in github.

My understanding is that the anchors are private to the yaml file. In jobs-chromeos.yaml the only `wifi-basic` entry is in the anchors, so it is not visible from the pipeline.yaml, hence there are no conflicts. I think that's what you were asking, but let me know otherwise.

config/pipeline.yaml
nfraprado: Yes, that sounds fine to me. I wanted to keep this PR contained to pipeline.yaml and so I preferred doing the name change here. But renaming the chromeos ones in a separate PR sounds good.

PR #978, opened 2025-01-24
Enable device error logs kselftest on Genio 700 and 1200 EVK boards

nfraprado:
Depends on https://github.com/kernelci/kernelci-core/pull/2786.

Genio 700: https://lava.collabora.dev/scheduler/job/17515212
Genio 1200: https://lava.collabora.dev/scheduler/job/17515224

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -3076,6 +3076,13 @@ scheduler:
     runtime: *lava-broonie-runtime
     platforms: *lava-broonie-arm64
 
+  - job: kselftest-device-error-logs-main
+    event: *kbuild-gcc-12-arm64-node-event
+    runtime: *lava-collabora-runtime
+    platforms:
+      - mt8390-genio-700-evk
+      - mt8395-genio-1200-evk
+
   - job: kselftest-dt
     event: *kbuild-gcc-12-arm64-node-event
     runtime: *lava-collabora-runtime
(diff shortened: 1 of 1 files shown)


nfraprado:
Job named `kselftest-device-error-logs-main` to avoid conflict with the `kselftest-device-error-logs` in chromeos-jobs.yaml. Let me know if that should be resolved otherwise.

nuclearcat:
Can you recheck networking setup please?
Here is LAVA job: https://lava.collabora.dev/scheduler/job/17539945
Kernel: https://staging.kernelci.org:9000/viewer?node_id=6798cc3820c4f0a20cc4eb81

nfraprado:
@nuclearcat There's a recent regression for which the fix is at https://lore.kernel.org/all/20250124191644.2309790-1-robh@kernel.org . Should hopefully land soon.

Besides that there's a hang on the Genio 700 EVK on latest nexts, which I'll tackle ASAP.

I'll ping back once these are resolved.

config/pipeline.yaml
```diff
@@ -1574,6 +1574,13 @@ jobs:
         - collabora-next:for-kernelci
     kcidb_test_suite: kselftest.devices-exist
 
+  kselftest-device-error-logs-main:
```
pawiecz, line 1577: [This also applies here](https://github.com/kernelci/kernelci-pipeline/pull/975#discussion_r1935330242) but it can be done as a next step.

I'd also like to see the diff of params applied to these job definitions (here and its ChromeOS counterpart) - are they that much different?

nfraprado:
@nuclearcat actually yes, both fixes were merged last week, so at least we should be able to see the tests running fine on linux-next.

I've rebased the PR to fix the conflict, so we should be good to go now.

PR #980, opened 2025-01-28
feat(send_kcidb.py): Add promisc flag to subscribe_filters

nuclearcat:
We need to be able to other users in API, to process events and send to kcidb.


nuclearcat:
Tested on staging

PR #981, opened 2025-01-29
fix(example_api_lab.py): Add mandatory field

nuclearcat:
We need to add field that indicates this record is not processed yet by kcidb bridge.


nuclearcat:
Tested manually

PR #983, opened 2025-01-29
upd(scheduler-chromeos.yaml): Disable debian decoder chromestack jobs

nuclearcat:
This jobs need optimization, they run over 7 hours and not completing.


pawiecz (approved):
LGTM

PR #988, opened 2025-01-31
Refactoring for kselftest handling in pipeline

broonie:
This is a minor refactoring of the base kselftest stuff in the pipeline,
in preparation for enabling more kselftests.

- config/pipeline.yaml: Factor generic kselftest rules into a separate block
- config/pipeline.yaml: Cover kselftest on -next by default

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -1537,20 +1537,30 @@ jobs:
       - 'stable-rc'
       - 'stable'
 
-  kselftest-cpufreq: &kselftest-job
+  # Named so it sorts before all the actual kselftests, we need to
+  # specify a kselftest suite for YAML validation.
+  kselftest-aaa: &kselftest-job
     template: generic.jinja2
     kind: job
     params: &kselftest-params
       test_method: kselftest
       boot_commands: nfs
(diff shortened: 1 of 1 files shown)


config/pipeline.yaml
```diff
@@ -1521,19 +1521,29 @@ jobs:
...
       test_method: kselftest
       boot_commands: nfs
       nfsroot: 'https://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20250117.0/{debarch}'
-      collections: cpufreq
       job_timeout: 10
     rules: &kselftest-rules
       tree:
         - mainline
         - stable-rc
+    kcidb_test_suite: kselftest.aaa
```
nuclearcat, line 1539: IMHO better to remove this default value to ensure values are explicitly set. This would allow CI/CD validation to catch incomplete entries rather than silently passing them with the default 'aaa' value to kcidb.

nuclearcat (changes requested):

nuclearcat:
Please rebase also

config/pipeline.yaml
broonie: We need to specify something since the YAML validation would blow up otherwise - my initial push didn't have it but it failed CI.

nuclearcat: Ok then just rebase and i will merge it and think how to solve this issue later.

PR #990, opened 2025-01-31
Enable DT selftests in my lab

broonie:
- config/pipeline.yaml: Define symbolic names for platform sets in my labs
- config/pipeline.yaml: Run DT selftests on the boards in my lab


nuclearcat (approved):
LGTM

PR #991, opened 2025-02-02
config: pipeline: add all stable branches

padovan:
We want to compare the stable-rc tests with the latest stable releases. So let's have the stable tests here too.

Also, I am sure people want to see stable results in KernelCI. :)

Next step is to match the same testing for both stable-rc and stable.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -3575,9 +3575,29 @@ build_configs:
     tree: soc
     branch: 'for-next'
 
+  stable_5.4:
+    tree: stable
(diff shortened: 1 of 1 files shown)


nuclearcat (commented):
Do we want to enable some(or same as stable) tests or just builds only for now?

nuclearcat:
I will merge, but in case we want to have tests too, you can add one more PR.

padovan:
> Do we want to enable some(or same as stable) tests or just builds only for now?

Yes. I need exact the same support as stable-rc.

PR #993, opened 2025-02-03
src/logspec_api: ignore errors of type linux.kernel.error_return_code

helen-fornazier:
do not generate issues for error_return_code since they are not
fatal, avoid noise.


nuclearcat (approved):
LGTM

PR #995, opened 2025-02-04
config: chromeos: Add ChromiumOS tree and Clang kbuild jobs

laura-nao:
Add the ChromiumOS kernel tree (branches 5.4 to 6.12) and define new Clang kbuild jobs for both ARM64 and x86_64 Chromebooks.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -1,4 +1,41 @@
 _anchors:
+  kbuild-clang-17-chromeos:
+    params: &kbuild-clang-17-chromeos-params
+      compiler: clang-17
+      defconfig: 'cros://chromeos-{krev}/{crosarch}/chromiumos-{flavour}.flavour.config'
+      flavour: '{crosarch}-generic'
+      kselftest: disable
+
+  kbuild-clang-17-arm64-chromeos: &kbuild-clang-17-arm64-chromeos-job
+    template: kbuild.jinja2
+    kind: kbuild
+    image: ghcr.io/kernelci/{image_prefix}clang-17:arm64-kselftest-kernelci
+    params: &kbuild-clang-17-arm64-chromeos-params
+      <<: *kbuild-clang-17-chromeos-params
+      arch: arm64
+      cross_compile: 'aarch64-linux-gnu-'
+      fragments:
+        - lab-setup
+        - arm64-chromebook
+        - CONFIG_MODULE_COMPRESS=n
+        - CONFIG_MODULE_COMPRESS_NONE=y
+    rules: &kbuild-clang-17-arm64-chromeos-rules
+      tree:
+      - chromiumos
+
+  kbuild-clang-17-x86-chromeos: &kbuild-clang-17-x86-chromeos-job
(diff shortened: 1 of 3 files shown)


laura-nao:
Please remove the `staging-skip` label if the changes look good enough for testing, I believe it requires a manual trigger for staging anyway.

config/jobs-chromeos.yaml
```diff
@@ -1,5 +1,43 @@
...
+      flavour: '{crosarch}-generic'
+      fragments:
+        - lab-setup
+        - x86-board
+        - CONFIG_MODULE_COMPRESS=n
+        - CONFIG_MODULE_COMPRESS_NONE=y
+      kselftest: disable
+    rules: &kbuild-clang-17-x86-chromeos-rules
+      tree:
+      - chromiumos
```
a-wai, line 39:
```suggestion
      <<: *kbuild-clang-17-arm64-chromeos-rules
```

config/jobs-chromeos.yaml
```diff
@@ -1,5 +1,43 @@
...
+      compiler: clang-17
+      cross_compile: 'aarch64-linux-gnu-'
+      defconfig: 'cros://chromeos-{krev}/{crosarch}/chromiumos-{flavour}.flavour.config'
+      flavour: '{crosarch}-generic'
+      fragments:
+        - lab-setup
+        - arm64-chromebook
+        - CONFIG_MODULE_COMPRESS=n
+        - CONFIG_MODULE_COMPRESS_NONE=y
+      kselftest: disable
```
a-wai, line 18: Those params (except `arch`, `cross_compile` and `fragments`) could be factored separately and re-used in the job definitions, adding only the arch-specific ones.

a-wai (commented):
LGTM, it could be a bit more concise but there's nothing wrong with the current implementation.

laura-nao:
> LGTM, it could be a bit more concise but there's nothing wrong with the current implementation.

Thanks for the feedback - I'll apply the fixes as suggested.

laura-nao:
> LGTM, I assume you tested the chromeos branch with Clang 17 :)

I ran just one manual LAVA job with a chromeos-6.12 build w/ Clang 17 (build was successful and boots) - see https://lava.collabora.dev/scheduler/job/17616230 for reference

I'll wait for the staging results to confirm if the other branches built successfully.

laura-nao:
@nuclearcat is there some way to manually trigger all related builds on staging once? should I process that myself with kci-dev?

laura-nao:
Refined branch filtering on the GCC builds as follows:
- MTK platforms: all except chromeos-5.4 and chromeos-6.6
- Qcom platforms: only chromeos-6.6
- AMD platforms: all except chromeos-5.15, chromeos-6.1, chromeos-6.6

Added also missing entries in scheduler file. Will check results on staging again and report back.

nuclearcat:
They are likely infrastructure issues, as staging restart often. Just retry one more time, but if it repeats on same jobs - i will investigate it step by step.

nuclearcat:
I think we will merge and investigate if any issues happen

PR #996, opened 2025-02-05
src/{send_kcidb,logspec_api}: add suport for kselftest

helen-fornazier:
In logspec_api, rename object_type to test_type, which are the supported types it can parse.

In send_kcidb, extract the test_type from the node.

Add kselftest to the list of types it supports.

### DIFF
--- a/src/kernelci_pipeline/logspec_api.py
+++ b/src/kernelci_pipeline/logspec_api.py
@@ -20,7 +20,7 @@ import logspec.main
 
 
 # Configuration tables per object type
-object_types = {
+test_types = {
     'build': {
         # logspec parser to use
         'parser': 'kbuild',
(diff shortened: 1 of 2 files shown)


pawiecz (approved):
Patch set looks good :+1: 

My personal preference would be to leave `object_type` as it is since it covers "build" Nodes as well but let's keep this PR as it is (reverting this rename is just a `sed` away if there are more non-test object types in the future)

padovan:
@helen-fornazier do you have an example of issue generated by this?

helen-fornazier:
> @helen-fornazier do you have an example of issue generated by this?

Yes: https://grafana.kernelci.org/d/issue/issue?var-datasource=edqupth5xi8e8b&var-id=maestro:346385016734add8169277c19da9a21b9d7fb1e2


Also, I just tested it and it seems to be working, I'm removing the Draft

PR #998, opened 2025-02-06
config/pipeline.yaml: Enable the arm64 selftests in my lab

broonie:
Running them on every single board in my lab is probably excessive but
we can deal with that if it gets to be an issue.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -1555,6 +1555,15 @@ jobs:
         - stable
     kcidb_test_suite: kselftest.aaa
 
+  kselftest-arm64:
+    <<: *kselftest-job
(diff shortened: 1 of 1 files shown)


pawiecz (approved):
LGTM

config/pipeline.yaml
```diff
@@ -1555,6 +1555,15 @@ jobs:
         - stable
     kcidb_test_suite: kselftest.aaa
 
+  kselftest-arm64:
+    <<: *kselftest-job
+    template: generic.jinja2
+    kind: job
```
pawiecz, line 1561:
```suggestion
```

These are covered by `<<: *kselftest-job` but having them also here doesn't hurt

PR #999, opened 2025-02-06
config/pipeline.yaml: Enable the KVM selftests on my arm64 boards

broonie:
We should really refine this to cover pKVM on at least some of the
systems that support this (currently just the i.MX8MP platforms), that
requires extra command line options and potentially some work on
reporting.


pawiecz (approved):
LGTM

PR #1000, opened 2025-02-06
config/pipeline.yaml: Enable timers selftests

broonie:
Cover the timers selftests on a subset of the boards in my lab with a
bit more capacity.


pawiecz (approved):
LGTM

PR #1001, opened 2025-02-06
upd(pipeline.yaml): Add aaptel tree for NVME-TCP offload

nuclearcat:
We need to test nvme-tcp offload feature, but until we integrate latest netdev trees, let's use this specific tree where latest version published.


pawiecz (approved):
LGTM

PR #1005, opened 2025-02-07
tools: improve api events example

padovan:
Add cmdline arg option for node kind and document the API parameters a bit.


nuclearcat (approved):
LGTM

PR #1015, opened 2025-02-13
config/pipeline.yaml: Build the kselftest fragment for more trees

broonie:
A lot of the kselftests require config options which are enabled in their
config fragments and incorporated into the kselftest build but currently we
only build that for the kselftest tree, meaning that we don't get the test
coverage for these. Enable build for the main upstream trees too so we can
run these tests.


nuclearcat (approved):
I think in new kbuild we didnt migrate this option, to take all options from kselftests, i have to fix that

PR #1016, opened 2025-02-14
Use dynamic branch triggers

pawiecz:
This patch series adds support for dynamically updated branch names in the Pipeline config. It's required for [NIPA system](https://github.com/linux-netdev/nipa/wiki/Netdev-CI-system) integration.

Please see also the details descriptions in specific patches.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -2081,6 +2081,9 @@ trees:
   net-next:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git"
 
+  netdev-testing:
+    url: "https://github.com/linux-netdev/testing.git"
+
   next:
     url: 'https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git'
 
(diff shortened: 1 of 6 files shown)


nuclearcat (approved):
LGTM!

padovan:
It looks like this does the job. I guess we try it in staging?

padovan (approved):

pawiecz:
Two patches added after initial review allow triggering custom `checkouts` using Pipeline API `/checkout` endpoint:

```
./tools/kci-maintainer --checkout -u https://github.com/linux-netdev/testing.git -b net-next-2025-02-14--15-00 --latest-commit -f kbuild-gcc-12-x86 -f baseline-x86
```

kci-dev uses the same endpoint so it should work fine as well

config/pipeline.yaml
```diff
@@ -3803,6 +3806,10 @@ build_configs:
     tree: net-next
     branch: 'main'
 
+  netdev-testing:
+    tree: netdev-testing
+    branch: 'https://netdev.bots.linux.dev/static/nipa/branches.json'
```
padovan, line 3811: The branches we are going to use comes from this file that Jakub just put together. They will only include code that  pass SW tests and will be generated every 8h usually.
 https://netdev.bots.linux.dev/static/nipa/branches-hw.json

nuclearcat: I will update this after merging

pawiecz: Fixed in https://github.com/kernelci/kernelci-pipeline/pull/1019

PR #1020, opened 2025-02-18
upd(kubernetes): Fix work directory

nuclearcat:
Due addition of base module we need to change work directory.


nuclearcat:
Tested on production (while fixing it).

PR #1022, opened 2025-02-18
send_kcidb: add 'lab' field to misc data submitted

padovan:
Today the Maestro lab is sent through the 'runtime' field, but that name is not that great for labs. Let's add a 'lab' field, but wait a bit to remove the 'runtime' field, so we can transition some KCIDB queries without any trouble.


nuclearcat (approved):
LGTM, looks simple enough and works on staging

PR #1031, opened 2025-02-20
config/pipeline.yaml: Enable LTP coverage on my BeagleBone Blacks

broonie:
I've got a lot of BeagleBone Black boards in my lab with plenty of spare
capacity, enable a lot of the LTP suites on them to give us more 32 bit
Arm coverage. There are some suites like the hugetlb one which don't make
sense on this architecture, and the board has relatively little memory so
the mm suite might also be an issue.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -3314,6 +3314,18 @@ scheduler:
     runtime:
       type: shell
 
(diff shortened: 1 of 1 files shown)


pawiecz (approved):
LGTM

PR #1032, opened 2025-02-20
config/pipeline.yaml: Enable the IPC selftests in my lab

broonie:
More mostly software only selftests so just enabled on one platform per
architecture.


pawiecz (approved):
LGTM

PR #1042, opened 2025-02-24
config/pipline.yaml: Enable mqueue selftests in my lab

broonie:
More software only tests, enabled with one board per architecture.


pawiecz (approved):
LGTM

PR #1045, opened 2025-02-24
docker-compose: remove obsolete 'version' attribute

padovan:
Fix this warning:
WARN[0000] kernelci-pipeline/docker-compose.yaml: the attribute version is obsolete, it will be ignored, please remove it to avoid potential confusion


nuclearcat:
It fails on staging because interpolation in parameters doesnt work in newer docker-compose files:
```
Stopping pipeline containers
ERROR: Invalid interpolation format for "monitor" option in service "services": "--settings=${KCI_SETTINGS:-/home/kernelci/config/kernelci.toml}"
```

padovan:
THis should be done as an Docker recipes overhaul.

Issue #1011, opened 2025-02-12
Annotate small variants of build config in the config name


padovan:
For example, kbuild-gcc-12-arm64-android-randomize is just defconfig + CONFIG_CPU_BIG_ENDIAN=y, but the config name still shows 'defconfig' only

maybe it should be 'defconfig+CONFIG_CPU_BIG_ENDIAN' ?

PR #1046, opened 2025-02-25
fix(send_kcidb.py): Send full config name to KCIDB

nuclearcat:
KCIDB need to be able to see full config name, not just defconfig.

Fixes: https://github.com/kernelci/kernelci-pipeline/issues/1011


padovan:
Looks good to me!

pawiecz (approved):
LGTM too

PR #1049, opened 2025-02-26
config/pipeline.yaml: Enable more LTP on my Beaglebone Blacks

broonie:
Add job stanzas for a lot of LTP suites we don't already cover and use the
spare capacity we have on Beaglebone Black in my lab to run a lot more of
LTP. Since the boards use NFS root, have only one CPU and limited memory
there are a bunch of suites that aren't appropriate but this is most of
them.

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -1865,6 +1865,18 @@ jobs:
       <<: *ltp-params
       tst_cmdfiles: "cap_bounds"
 
+  ltp-containers:
+    <<: *ltp-job
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "containers"
+
+  ltp-controllers:
+    <<: *ltp-job
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "controllers"
+
   ltp-cpuhotplug:
     <<: *ltp-job
     params:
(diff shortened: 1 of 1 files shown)


config/pipeline.yaml
```diff
@@ -1865,6 +1865,18 @@ jobs:
       <<: *ltp-params
       tst_cmdfiles: "cap_bounds"
 
+  ltp-containers:
+    <<: *ltp-job
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "containers"
```
pawiecz, line 1872: All the LTPs added by this patch will be submitted to KCIDB as "ltp" suite:
https://github.com/kernelci/kernelci-pipeline/blob/c4bb6cdc41e80b070c02db5bfe0ffa67c1d8dd7b/config/pipeline.yaml#L86

@broonie Do you plan to use KCIDB as a data source for processing the results? Should the LTP Jobs' parameters be extended with `kcidb_test_suite` like the kselftests?

broonie: Yeah, I was wondering about the decision to not use subsuites for LTP, I was just following what was there for the existing LTP entries. One reason for keeping things that way is that people make their own custom LTP test lists and run those so tests end up showing up in multiple groupings, or different groupings per CI system.

pawiecz: Got it - let's keep LTPs as they were for now and raise the topic of test suite naming in the KCIDB submissions at the next weekly call with @spbnick (I'll add it to the agenda)

pawiecz (approved):
LGTM

PR #914, opened 2024-11-26
Move trees URLs and legacy build configuration to a separate file

pawiecz:
These sections of Pipeline config should not be buried in already too long YAML document because they are the entry point for new kernel trees tested by Maestro.

PR #1054, opened 2025-02-27
pipeline: config: split config in multiple files

padovan:
The config grew to an unmanageable size, so let's split it
    into multiple files.
    
    Note: "config/pipeline.yaml" was mentioned in results_summary
    service, but that services doesn´t use configs at all.


doc/developer-documentation.md
```diff
@@ -125,11 +126,11 @@ build_configs:
     branch: <branch-name1>
 ```
 
-That's it! The tree is enabled now. All the jobs defined under `jobs` section of [config file](https://github.com/kernelci/kernelci-pipeline/blob/main/config/pipeline.yaml) would run on the specified branched for this tree.
+That's it! The tree is enabled now. All the jobs defined under `jobs` section of [config file](https://github.com/kernelci/kernelci-pipeline/blob/main/config/jobs.yaml) would run on the specified branched for this tree.
 
 ### Schedule the job
 
-We also need a `scheduler` entry for the newly added job to specify pre-conditions for scheduling, and defining runtime and platforms for job submissions.
+We also need a `scheduler` entry of [config/scheduler.yaml](https://github.com/kernelci/kernelci-pipeline/blob/main/config/scheduler.yaml) for the newly added job to specify pre-conditions for scheduling, and defining runtime and platforms for job submissions.
```
pawiecz, line 133:
```suggestion
We also need a `scheduler` entry in [config/scheduler.yaml](https://github.com/kernelci/kernelci-pipeline/blob/main/config/scheduler.yaml) for the newly added job to specify pre-conditions for scheduling, and defining runtime and platforms for job submissions.
```
:question:

pawiecz:
#914 has been adjusted and merged - sorry it requires another rebase of this PR.

Once the main file is finally split there should be less conflicts and less overall friction with the Pipeline config refactoring.

config/jobs.yaml
```diff
@@ -0,0 +1,2085 @@
...
+    - 'sashal-next'
+    - 'soc'
+    - 'stable'
+    - 'tegra'
+    - 'thermal'
+    - 'tip'
+    - 'ulfh'
+    - 'vireshk'
+
+api:
```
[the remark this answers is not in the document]
padovan: Oh, missed these files there.

doc/developer-documentation.md
padovan: ack

nuclearcat:
Tested on staging, thanks !

PR #1056, opened 2025-02-27
services: remove unused imports

padovan:
A bit of clean up, as part of a review of the kernelci. imports I was doing.


nuclearcat:
Thank you very much!

PR #1058, opened 2025-02-27
config: trees: Add aaptel for ci tree

padovan:


padovan:
To be merged after the config split change go in.

PR #1060, opened 2025-02-28
(docker-compose): Prepare staging for workflows

nuclearcat:
To prepare for workflows we need to ignore local sources, and use docker staging images built in github workflows.


nuclearcat:
staging... hmmm, tested on staging

PR #1065, opened 2025-03-03
config: Enable Raspberry Pi 3B+ in my lab

broonie:
Basic enablement for core features of the board which I've added to my
lab.


pawiecz (approved):
LGTM

PR #1068, opened 2025-03-03
(lava_callback.py): Add job-filter

nuclearcat:
Sometimes we want to retry test job, AND also
retry associated "child" jobs, such as nipa-update.


nuclearcat:
Tested on staging

PR #1071, opened 2025-03-04
config: jobs-chromeos: Fix rules override for Tast decoder tests

laura-nao:
The current rules override in the Tast decoder test job definition only considers the minimum kernel version, ignoring the existing tree-specific rules. Reintroduce the tree rules.


pawiecz (approved):
LGTM

PR #1075, opened 2025-03-07
config: Run only fluster-debian Jobs on media-committers tree

pawiecz:
This patch creates a separate KBuild Job for media-committers tree to exclude it from the base "kbuild-gcc-12-arm64-chromebook" kernel build one.

Duplicated base KBuild Job will run only for media-committers tree and will trigger only fluster-debian tests. Other tests could create noise when verifying media-committers changes.

Scheduler entries had to be duplicated to use the new triggering event.


nuclearcat:
Tested on staging

PR #1079, opened 2025-03-10
config/scheduler.yaml: Offload some arm64 jobs onto Raspberry Pi 3+

broonie:
Since the NFS issues have a fix on their way to -next offload some of the
longer running test jobs from other boards to the Raspberry Pi 3+s in my
lab to balance out capacity. The boards concerned are all four cour Cortex
A53s so there should be no meaningful coverage impact when everything is
running.


pawiecz (approved):
LGTM

PR #1085, opened 2025-03-17
config: Lower the frequency of chromiumos builds to 1/day

laura-nao:
Lower the frequency of chromiumos builds to 1/day, both for GCC-12 and Clang-17 builds.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -5,6 +5,7 @@ _anchors:
       defconfig: 'cros://chromeos-{krev}/{crosarch}/chromiumos-{flavour}.flavour.config'
       flavour: '{crosarch}-generic'
       kselftest: disable
+      frequency: 1d
 
   kbuild-clang-17-arm64-chromeos: &kbuild-clang-17-arm64-chromeos-job
     template: kbuild.jinja2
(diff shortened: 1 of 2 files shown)


laura-nao:
Hi @nuclearcat and @pawiecz - the changes in the PR are meant to lower the frequency of the `chromiumos` tree builds to 1/day. I couldn't find an easy way of defining different frequency for a kbuild job based on the tree, so I defined new kbuild jobs with the desired frequency targeting the `chromiumos` tree.

Does this sound right to you? is there any other more straightforward way of doing this?

If the changes look good to you, please feel free to remove the staging-skip label.

config/jobs-chromeos.yaml
```diff
@@ -56,6 +57,7 @@ _anchors:
     rules: &kbuild-gcc-12-arm64-chromeos-rules
       tree:
       - '!android'
+      - '!chromiumos'
```
pawiecz, line 60: Naming scheme for the set of key anchors in this config file might need an update in the future: now there's a ChromeOS job that is set to avoid ChromiumOS kernel tree - something to address later on.

laura-nao: I agree the current naming is a bit confusing (maybe `chromeoscfg` would be a better infix on kbuild jobs for non-ChromeOS kernels with the ChromeOS configs) . I can address this in a separate PR.

laura-nao:
Thanks @pawiecz ! I adjusted the config a bit based on the staging results. Will check again shortly and confirm when it's ready to go.

laura-nao:
I confirm staging results look ok to me, and the frequency setting seems to be working as expected. So, should be ready to be merged. Thanks!

PR #1095, opened 2025-03-27
Add brask, nissa, hana and cherry-r3 Chromebooks

laura-nao:
Add support for brask, nissa, hana and cherry-r3 Chromebooks. Enable baseline/baseline-nfs tests only for starters.


pawiecz:
Skipping from staging deployment to verify #1090

pawiecz (approved):
LGTM

PR #1096, opened 2025-03-27
config: Enable mm selftests on Avenger96

broonie:
The upstream issues which prevented kselftest-mm runs on arm have been
fixed so let's enable the tests, use Avenger96 since it's quite fast and
has a relatively large amount of RAM.  Standard two level page tables
don't support hugepages on 32 bit.

### DIFF
--- a/config/jobs.yaml
+++ b/config/jobs.yaml
@@ -1601,6 +1601,15 @@ jobs:
       collections: landlock
     kcidb_test_suite: kselftest.landlock
 
+  # No hugepages allocation (for architectures without it)
+  kselftest-mm:
+    <<: *kselftest-job
+    params:
+      <<: *kselftest-params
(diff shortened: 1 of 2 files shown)


config/jobs.yaml
```diff
@@ -1601,6 +1601,15 @@ jobs:
       collections: landlock
     kcidb_test_suite: kselftest.landlock
 
+  # No hugepages allocation (for architectures without it)
+  kselftest-mm:
+    <<: *kselftest-job
+    params:
+      <<: *kselftest-params
+      extra_kernel_args: "secretmem.enable kpti=off"
```
pawiecz, line 1609: kernelci/kernelci-core#2828 will introduce support for passing `extra_kernel_args`

pawiecz (approved):
LGTM

Blocked until kernelci/kernelci-core#2828 is merged

PR #1097, opened 2025-03-27
config: Enable the ublk selftests

broonie:
These are software only tests, run them on one board per architecture in
my lab.


pawiecz (approved):
LGTM

PR #1100, opened 2025-04-03
Add official PCI tree to run tests on the next branch

kwilczynski:
Add the official PCI tree such that tests can be run on the `next` branch for the following architectures:

- x86_64
- arm64
- arm

This aims to increase and improve test coverage as part of the PCI sub-system development efforts.

### DIFF
--- a/config/jobs.yaml
+++ b/config/jobs.yaml
@@ -102,6 +102,7 @@ _anchors:
     - 'tip'
     - 'ulfh'
     - 'vireshk'
+    - 'linux-pci'
 
 
 jobs:
--- a/config/trees.yaml
+++ b/config/trees.yaml
@@ -153,6 +153,9 @@ trees:
   hyperv:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/hyperv/linux.git"
 
+  linux-pci:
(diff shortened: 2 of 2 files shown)


kwilczynski:
@kernelci/hackers, let me know if anything else needs to be done here. Thank you!

padovan (approved):

config/jobs.yaml
```diff
@@ -102,6 +102,7 @@ _anchors:
     - 'tip'
     - 'ulfh'
     - 'vireshk'
+    - 'linux-pci'
```
pawiecz, line 105: I know _some_ Team Members who are strict about alphabetical order but I'm not one of them ;)

It will need sorting at some point anyway with "qcom" at the very top

pawiecz (approved):
LGTM

config/jobs.yaml
kwilczynski: @pawiecz, I noticed this too, but I wasn't sure whether this was something we cared about or not.

Happy to make mends here in the follow-up change. :+1:

kwilczynski: Closing the loop with the following:

- https://github.com/kernelci/kernelci-pipeline/pull/1109

PR #1101, opened 2025-04-03
Re-enable kontron's lab

hthiery:
The lab was not migrated to new pipeline, lets do it.

The callback token is still the same as used with the old pipeline configuration.


nuclearcat:
Hi, we are getting
`    "error_msg": "403 Client Error: Forbidden for url: https://lavalab.kontron.com/api/v0.2/jobs/?format=json&limit=256"`
Maybe token mismatch or some apache config issue? (ips are changed)

hthiery:
Hi,

when using the url  `https://lavalab.kontron.com/api/v0.2/jobs/?format=json&limit=256` in browser I see the results. Could you please check again?

Heiko

PR #1102, opened 2025-04-03
Fix and improve for local deployment

tollsimy:
This PR aims to fix and improve containers for local deployments by:
- remove obsolete Dockerfiles (move to `./kci docker`)
- merging all compose files in one (pipeline + lava-callback + pipeline-kcidb)
- refactoring compose files and removing obsolete entries
- add missing config path argument to services
- updating relative docs

This PR is tightly related to recent fixes and improvements I am proposing in `kernelci/kernelci-deploy`.


docker-compose.yaml
```diff
@@ -11,6 +11,9 @@ services:
   monitor: &base-service
     container_name: 'kernelci-pipeline-monitor'
     image: 'kernelci/staging-kernelci:pipeline'
+    build:
```
nuclearcat, line 14: This files are obsolete. We are building docker images using our own constructor, for example in: https://github.com/kernelci/kernelci-core/blob/main/.github/workflows/docker_images.yml

So for pipeline it will be:
./kci docker build kernelci pipeline
then for api     `./kci docker build kernelci pipeline`
and   'kernelci lava-callback'.
For staging you need a bit different flags to build images on non-default branch. So it is a bit non-trivial to add this in docker-compose.

nuclearcat (commented):
How to reproduce bug this patch (rename timeout) is fixing?

docker-compose.yaml
tollsimy: I am deploying locally using suggested scripts from `kernelci-deploy/localinstall/kci-deploy.py`.

tollsimy:
Rename timeout patch fixes:
```
cycle detected at path: services.timeout
cycle detected at path: services.timeout
```

Environment:
```
docker --version
Docker version 28.0.2, build 0442a73
docker-compose --version
Docker Compose version v2.30.2
```

docker-compose.yaml
nuclearcat: Yes, thats correct, look at: kernelci-deploy/localinstall/1-rebuild_all.sh
For example:
```
echo Build docker images: api
./kci docker $args kernelci api --version="$api_rev"
failonerror
```
Same way you can add missing pipeline and lava callback

tollsimy: Got it, should we mentions somewhere that the Dockerfiles are deprecated or just remove them? This is quite confusing honestly.

tollsimy:
What about something like this? I remove obsolete Dockerfiles. I will open another PR in deployment to add missing container builds.

tollsimy:
Added build of missing pipeline images here: https://github.com/kernelci/kernelci-deploy/pull/223

nuclearcat:
I will need to check if renaming will cause any issues with existing deployment scripts, it might take a bit of time, but likely will do this week.

tollsimy:
Edit: apparently the `cycle detected at path` is a docker-compose bug related to yaml anchors. Does not happen if using the new `docker compose` command. Time to remove depreated `docker-compose`.
I will change and push new changes.

nuclearcat:
Probably this PR causes breakage at staging script
```
ERROR: Version mismatch: file ./docker-compose.yaml specifies version 1 but extension file ./docker-compose-kcidb.yaml uses version 3.0
```

nuclearcat:
I think you need to rebase your patch

tollsimy:
Done, this should be work, thanks @nuclearcat !

tollsimy:
In order not to break staging this would require to change `staging.kernelci.org` in `kernelci-deploy` from:

```
cmd_api_pipeline() {
    compose_files="\
-f docker-compose.yaml \
-f docker-compose-kcidb.yaml \
-f docker-compose-lava.yaml \
-f docker-compose-production.yaml \
"
```
to
```
cmd_api_pipeline() {
    compose_files="\
-f docker-compose.yaml \
-f docker-compose-production.yaml \
"
```

PR #1103, opened 2025-04-03
Add architectures filter to limit builds to relevant platforms only

kwilczynski:
KernelCI will schedule builds for each currently supported architecture for each tree. This is also the default behaviour if no specific architecture filters are provided. Such builds can include architectures that are of no interest to the owner of a given tree, or the code that the tree carries would not be intended to run on some architectures.

For example, for an ARM64 development tree, testing on an x86_64 or RISC-V platform is potentially of no use to the developers, even if errors were reported. However, running these extra builds impacts the available resources (CPU, memory, I/O, etc.) and the time it takes to allocate a build machine to run builds and tests on.

Thus, limit each tree to the most suitable architectures (on a base-effort basis). This can include specific target architectures (such as RISC-V for the "riscv" tree), or a set of standard architectures, such as x86_64, ARM and ARM64, on which the tree will build to reduce resource usage across runners and limit the amount of results that are not that useful.

While at it, arrange the list of trees and their respective configuration in a sorted order, and add entries for stable 6.13 and 6.14 trees, as these kernel versions have been released a while ago.

Related:

- https://github.com/kernelci/kernelci-pipeline/pull/1086

### DIFF
--- a/config/trees.yaml
+++ b/config/trees.yaml
@@ -1,5 +1,10 @@
 trees:
 
+  _base:
+    url: 'https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git'
+
+  # NOTE: Please keep the trees list below in sorted order.
+
   aaptel:
     url: "https://github.com/aaptel/linux.git"
 
@@ -54,6 +59,9 @@ trees:
   efi:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/efi/efi.git"
 
+  hyperv:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/hyperv/linux.git"
+
   kernelci:
     url: "https://github.com/kernelci/linux.git"
 
@@ -75,14 +83,17 @@ trees:
   linusw:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/linusw/linux-gpio.git"
 
+  linux-pci:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/pci/pci.git"
+
   mainline:
     url: 'https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git'
 
-  media:
-    url: 'https://git.linuxtv.org/media.git'
-
   media-committers:
     url: 'https://gitlab.freedesktop.org/linux-media/media-committers.git'
+
+  media:
+    url: 'https://git.linuxtv.org/media.git'
 
   mediatek:
     url: 'https://git.kernel.org/pub/scm/linux/kernel/git/mediatek/linux.git'
@@ -126,14 +137,14 @@ trees:
   soc:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/soc/soc.git"
 
-  stable:
-    url: 'https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git'
-
   stable-rc:
     url: 'https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git'
 
   stable-rt:
     url: 'https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-stable-rt.git'
+
+  stable:
+    url: 'https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git'
 
   tegra:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/tegra/linux.git"
@@ -149,12 +160,6 @@ trees:
 
   vireshk:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/vireshk/linux.git"
-
-  hyperv:
-    url: "https://git.kernel.org/pub/scm/linux/kernel/git/hyperv/linux.git"
-
-  linux-pci:
-    url: "https://git.kernel.org/pub/scm/linux/kernel/git/pci/pci.git"
 
 # -----------------------------------------------------------------------------
 # Legacy configuration data (still used by trigger service)
@@ -191,144 +196,192 @@ build_variants:
 
 build_configs:
 
+  # Includes all supported architectures.
+  _base: &base
+    tree: _base
+    branch: 'master'
+
+  # Includes most common and/or widely deployed architectures.
+  # The 32-bit x86 architecture is considered legacy nowadays,
+  # and as such, not included here.
+  _base-standard: &base-standard
+    <<: *base
+    architectures:
+      - x86_64
+      - arm64
+      - arm
+
+  # Includes ARM-specific architectures only.
+  _base-arm: &base-arm
+    <<: *base
+    architectures:
+      - arm64
+      - arm
+
+  # NOTE: Please keep the trees list below in sorted order.
+
   aaptel:
+    <<: *base-standard
     tree: aaptel
     branch: 'nvme-rx-offload-ci'
 
   amlogic:
+    <<: *base-arm
     tree: amlogic
     branch: 'for-next'
 
-  android_mainline:
+  android_mainline: &android
+    <<: *base-arm
     tree: android
     branch: 'android-mainline'
 
   android_mainline_tracking:
-    tree: android
+    <<: *android
     branch: 'android-mainline-tracking'
 
   android11-5.4:
-    tree: android
+    <<: *android
     branch: 'android11-5.4'
 
   android12-5.4:
-    tree: android
+    <<: *android
     branch: 'android12-5.4'
 
   android12-5.4-lts:
-    tree: android
+    <<: *android
     branch: 'android12-5.4-lts'
 
   android12-5.10:
-    tree: android
+    <<: *android
     branch: 'android12-5.10'
 
   android12-5.10-lts:
-    tree: android
+    <<: *android
     branch: 'android12-5.10-lts'
 
   android13-5.10:
-    tree: android
+    <<: *android
     branch: 'android13-5.10'
 
   android13-5.10-lts:
-    tree: android
+    <<: *android
     branch: 'android13-5.10-lts'
 
   android13-5.15:
-    tree: android
+    <<: *android
     branch: 'android13-5.15'
 
   android13-5.15-lts:
-    tree: android
+    <<: *android
     branch: 'android13-5.15-lts'
 
   android14-5.15:
-    tree: android
+    <<: *android
     branch: 'android14-5.15'
 
   android14-5.15-lts:
-    tree: android
+    <<: *android
     branch: 'android14-5.15-lts'
 
   android14-6.1:
-    tree: android
+    <<: *android
     branch: 'android14-6.1'
 
   android14-6.1-lts:
-    tree: android
+    <<: *android
     branch: 'android14-6.1-lts'
 
   android15-6.6:
-    tree: android
+    <<: *android
     branch: 'android15-6.6'
 
   android15-6.6-lts:
-    tree: android
+    <<: *android
     branch: 'android15-6.6-lts'
 
   android16-6.12:
-    tree: android
+    <<: *android
     branch: 'android16-6.12'
 
   ardb:
+    <<: *base-standard
     tree: ardb
     branch: 'for-kernelci'
 
   arm64:
     tree: arm64
     branch: 'for-kernelci'
+    architectures:
+      - arm64
 
   arnd:
+    <<: *base-standard
     tree: arnd
     branch: 'to-build'
 
+  broonie: &broonie
+    <<: *base-standard
+
   broonie-misc:
+    <<: *broonie
     tree: broonie-misc
     branch: 'for-kernelci'
 
   broonie-regmap:
+    <<: *broonie
     tree: broonie-regmap
     branch: 'for-next'
 
   broonie-regmap-fixes:
+    <<: *broonie
     tree: broonie-regmap
     branch: 'for-linus'
 
   broonie-regulator:
+    <<: *broonie
     tree: broonie-regulator
     branch: 'for-next'
 
   broonie-regulator-fixes:
+    <<: *broonie
     tree: broonie-regulator
     branch: 'for-linus'
 
   broonie-sound:
+    <<: *broonie
     tree: broonie-sound
     branch: 'for-next'
 
   broonie-sound-fixes:
+    <<: *broonie
     tree: broonie-sound
     branch: 'for-linus'
 
   broonie-spi:
+    <<: *broonie
     tree: broonie-spi
     branch: 'for-next'
 
   broonie-spi-fixes:
+    <<: *broonie
     tree: broonie-spi
     branch: 'for-linus'
 
-  chrome-platform:
+  chrome-platform: &chrome-platform
+    <<: *base-standard
     tree: chrome-platform
     branch: 'for-next'
 
   chrome-platform-firmware:
-    tree: chrome-platform
+    <<: *chrome-platform
     branch: 'for-firmware-next'
 
-  chromiumos-5.4: &chromiumos
+  chromiumos: &chromiumos
+    <<: *base-standard
     tree: chromiumos
+
+  chromiumos-5.4:
+    <<: *chromiumos
     branch: 'chromeos-5.4'
 
   chromiumos-5.10:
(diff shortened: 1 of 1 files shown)


nuclearcat:
Sorry pressed button by mistake, reversed back to draft.

kwilczynski:
@kernelci/hackers, when you have a moment, please help with reviews here. I want to make sure that the right architectures are selected for each tree, and whether we can improve things further. Thank you!

config/trees.yaml
```diff
@@ -188,144 +188,188 @@ build_variants:
 
 build_configs:
 
+  mainline: &mainline
+    tree: mainline
+    branch: 'master'
+
+  mainline-standard: &mainline-standard
```
padovan, line 195: I think we are missing an option for *all* supported architectures that should be used in:
* mainline
* next
* stable and stable-rc
* sashal/linus-next

padovan (commented):
It looks good to me. Left one comment. After we solve it, I think we are good to go.

config/trees.yaml
kwilczynski: @padovan, if we do not add the architectures property filter, a given tree would run jobs on every architecture we support. This is the default behaviour.

kwilczynski: Done.

kwilczynski: I wonder if the "net-next" tree should also run on every architecture. Same for "netdev-testing" tree. But, perhaps only the former would benefit from running on every platform we cover.

kwilczynski: I will move "net-next" to run on every architecture. Seems only prudent.

padovan: Keep netdev-testing out of it for now. We are running some specific tests on it for the time being.

kwilczynski: Done.

config/trees.yaml
```diff
@@ -351,232 +407,263 @@ build_configs:
...
     tree: next
-    branch: 'master'
 
   next_pending-fixes:
-    tree: next
+    <<: *next
     branch: 'pending-fixes'
 
   omap:
+    <<: *base-standard
```
pawiecz, line 600: Please correct me if I'm wrong - does x86_64 build make sense for this tree? Shouldn't it be `base-arm`?

config/trees.yaml
```diff
@@ -351,232 +407,263 @@ build_configs:
...
-  stable_6.6:
-    tree: stable
-    branch: 'linux-6.6.y'
-
-  stable_6.12:
-    tree: stable
-    branch: 'linux-6.12.y'
-
-  stable-rc_5.4: &stable-rc
+  stable-rc: &stable-rc
```
pawiecz, line 661: If I understand correctly this entry was added to create a nice and clean base anchor for all build configurations of `stable-rc` tree even though previously there was no entry for `stable-rc:master`.

I agree this is a good approach for trees with dormant `master` branch (like `linux-cip` or `kernelci` above) but I'm not sure if this also applies to (semi-)active master branches.

Would it make sense to create a base anchor on a first build configuration entry using given tree? Let me know what you think (maybe this additional cost turns out negligible over time)

config/trees.yaml
kwilczynski: Good catch. Makes sense.

kwilczynski: Done.

config/trees.yaml
kwilczynski: Done.

kwilczynski:
The difference between old trees.yaml and version with the changes introduced here:

<details>
<summary>patch.diff</summary>

```diff
--- trees.yaml	2025-04-10 17:48:04.315858091 +0900
+++ trees.yaml	2025-04-10 17:47:51.533091532 +0900
@@ -1706,163 +1706,381 @@
     }
   },
   "build_configs": {
+    "_base": {
+      "branch": "master",
+      "tree": "_base"
+    },
+    "_base-arm": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
+      "branch": "master",
+      "tree": "_base"
+    },
+    "_base-standard": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
+      "branch": "master",
+      "tree": "_base"
+    },
     "aaptel": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "nvme-rx-offload-ci",
       "tree": "aaptel"
     },
     "amlogic": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "amlogic"
     },
     "android11-5.4": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android11-5.4",
       "tree": "android"
     },
     "android12-5.10": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android12-5.10",
       "tree": "android"
     },
     "android12-5.10-lts": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android12-5.10-lts",
       "tree": "android"
     },
     "android12-5.4": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android12-5.4",
       "tree": "android"
     },
     "android12-5.4-lts": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android12-5.4-lts",
       "tree": "android"
     },
     "android13-5.10": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android13-5.10",
       "tree": "android"
     },
     "android13-5.10-lts": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android13-5.10-lts",
       "tree": "android"
     },
     "android13-5.15": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android13-5.15",
       "tree": "android"
     },
     "android13-5.15-lts": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android13-5.15-lts",
       "tree": "android"
     },
     "android14-5.15": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android14-5.15",
       "tree": "android"
     },
     "android14-5.15-lts": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android14-5.15-lts",
       "tree": "android"
     },
     "android14-6.1": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android14-6.1",
       "tree": "android"
     },
     "android14-6.1-lts": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android14-6.1-lts",
       "tree": "android"
     },
     "android15-6.6": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android15-6.6",
       "tree": "android"
     },
     "android15-6.6-lts": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android15-6.6-lts",
       "tree": "android"
     },
     "android16-6.12": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android16-6.12",
       "tree": "android"
     },
     "android_mainline": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android-mainline",
       "tree": "android"
     },
     "android_mainline_tracking": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "android-mainline-tracking",
       "tree": "android"
     },
     "ardb": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-kernelci",
       "tree": "ardb"
     },
     "arm64": {
+      "architectures": [
+        "arm64"
+      ],
       "branch": "for-kernelci",
       "tree": "arm64"
     },
     "arnd": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "to-build",
       "tree": "arnd"
     },
+    "broonie": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
+      "branch": "master",
+      "tree": "_base"
+    },
     "broonie-misc": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-kernelci",
       "tree": "broonie-misc"
     },
     "broonie-regmap": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "broonie-regmap"
     },
     "broonie-regmap-fixes": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-linus",
       "tree": "broonie-regmap"
     },
     "broonie-regulator": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "broonie-regulator"
     },
     "broonie-regulator-fixes": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-linus",
       "tree": "broonie-regulator"
     },
     "broonie-sound": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "broonie-sound"
     },
     "broonie-sound-fixes": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-linus",
       "tree": "broonie-sound"
     },
     "broonie-spi": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "broonie-spi"
     },
     "broonie-spi-fixes": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-linus",
       "tree": "broonie-spi"
     },
     "chrome-platform": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "chrome-platform"
     },
     "chrome-platform-firmware": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-firmware-next",
       "tree": "chrome-platform"
     },
+    "chromiumos": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
+      "branch": "master",
+      "tree": "chromiumos"
+    },
     "chromiumos-5.10": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "chromeos-5.10",
       "tree": "chromiumos"
     },
     "chromiumos-5.15": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "chromeos-5.15",
       "tree": "chromiumos"
     },
     "chromiumos-5.4": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "chromeos-5.4",
       "tree": "chromiumos"
     },
     "chromiumos-6.1": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "chromeos-6.1",
       "tree": "chromiumos"
     },
     "chromiumos-6.12": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "chromeos-6.12",
       "tree": "chromiumos"
     },
     "chromiumos-6.6": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "chromeos-6.6",
       "tree": "chromiumos"
     },
@@ -1907,46 +2125,94 @@
       "tree": "cip"
     },
     "clk": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "clk-next",
       "tree": "clk"
     },
     "collabora-chromeos-kernel_for-kernelci": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-kernelci",
       "tree": "collabora-chromeos-kernel"
     },
     "collabora-next_for-kernelci": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-kernelci",
       "tree": "collabora-next"
     },
     "efi": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "next",
       "tree": "efi"
     },
     "hyperv-fixes": {
+      "architectures": [
+        "x86_64"
+      ],
       "branch": "hyperv-fixes",
       "tree": "hyperv"
     },
     "hyperv-next": {
+      "architectures": [
+        "x86_64"
+      ],
       "branch": "hyperv-next",
       "tree": "hyperv"
     },
     "kernelci_staging-mainline": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "staging-mainline",
       "tree": "kernelci"
     },
     "kernelci_staging-next": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "staging-next",
       "tree": "kernelci"
     },
     "kernelci_staging-stable": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "staging-stable",
       "tree": "kernelci"
     },
     "khilman": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "to-build",
       "tree": "khilman"
     },
     "krzysztof": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "krzysztof"
     },
@@ -1959,22 +2225,52 @@
       "tree": "kselftest"
     },
     "lee_backlight": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-backlight-next",
       "tree": "lee-backlight"
     },
     "lee_mfd": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-mfd-next",
       "tree": "lee-mfd"
     },
+    "linusw": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
+      "branch": "master",
+      "tree": "linusw"
+    },
     "linusw_devel": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "devel",
       "tree": "linusw"
     },
     "linusw_fixes": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "fixes",
       "tree": "linusw"
     },
     "linusw_for-next": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "linusw"
     },
@@ -1992,22 +2288,46 @@
       "tree": "mainline"
     },
     "media-committers_fixes": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "fixes",
       "tree": "media-committers"
     },
     "media-committers_next": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "next",
       "tree": "media-committers"
     },
     "media_fixes": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "fixes",
       "tree": "media"
     },
     "media_next": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "next",
       "tree": "media"
     },
     "mediatek_for_next": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "mediatek"
     },
@@ -2028,26 +2348,52 @@
       "tree": "next"
     },
     "omap": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "omap"
     },
     "padovan": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-kernelci",
       "tree": "padovan"
     },
     "pm": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "testing",
       "tree": "pm"
     },
     "qcom": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "qcom"
     },
     "renesas": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "master",
       "tree": "renesas"
     },
     "renesas_next": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "next",
       "tree": "renesas"
     },
@@ -2066,10 +2412,20 @@
       "tree": "riscv"
     },
     "robh": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-kernelci",
       "tree": "robh"
     },
     "rppt": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-kernelci",
       "tree": "rppt"
     },
@@ -2078,10 +2434,18 @@
       "tree": "sashal-next"
     },
     "soc_fixes": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "arm/fixes",
       "tree": "soc"
     },
     "soc_for-next": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "soc"
     },
@@ -2117,6 +2481,10 @@
       "branch": "linux-6.13.y",
       "tree": "stable-rc"
     },
+    "stable-rc_6.14": {
+      "branch": "linux-6.14.y",
+      "tree": "stable-rc"
+    },
     "stable-rc_6.6": {
       "branch": "linux-6.6.y",
       "tree": "stable-rc"
@@ -2189,27 +2557,59 @@
       "branch": "linux-6.12.y",
       "tree": "stable"
     },
+    "stable_6.13": {
+      "branch": "linux-6.13.y",
+      "tree": "stable"
+    },
+    "stable_6.14": {
+      "branch": "linux-6.14.y",
+      "tree": "stable"
+    },
     "stable_6.6": {
       "branch": "linux-6.6.y",
       "tree": "stable"
     },
     "tegra": {
+      "architectures": [
+        "arm64",
+        "arm"
+      ],
       "branch": "for-next",
       "tree": "tegra"
     },
     "thermal": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "testing",
       "tree": "thermal"
     },
     "tip": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "master",
       "tree": "tip"
     },
     "ulfh": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "next",
       "tree": "ulfh"
     },
     "vireshk": {
+      "architectures": [
+        "x86_64",
+        "arm64",
+        "arm"
+      ],
       "branch": "for-kernelci",
       "tree": "vireshk"
     }
@@ -19232,6 +19632,9 @@
     }
   },
   "trees": {
+    "_base": {
+      "url": "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git"
+    },
     "aaptel": {
       "url": "https://github.com/aaptel/linux.git"
     },
```

</details>

config/trees.yaml
pawiecz: > > Would it make sense to create a base anchor on a first build configuration entry using given tree? Let me know what you think (maybe this additional cost turns out negligible over time)
> 
> You mean the cost of parsing this file?

Not at all - I thought of the additional builds triggered by those new "base anchors". I agree readability is paramount in this case of large YAMLs and might take priority over some extra builds.

Thanks for adjusting your changes and sorry for writing confusing comment, I should have proof-read it before posting :sweat_smile:

kwilczynski: > > You mean the cost of parsing this file?
> 
> Not at all - I thought of the additional builds triggered by those new "base anchors". I agree readability is paramount in this case of large YAMLs and might take priority over some extra builds.

@pawiecz, ahh, this makes more sense now.

I wasn't aware of this, or rather, I assumed (and it seems incorrectly) that these trees are simply configuration elements, and each has to be explicitly referenced in a job somewhere for it to be used.

But if these are picked verbatim, I must revisit adding the `_base` trees to facilitate reuse.

> Thanks for adjusting your changes and sorry for writing confusing comment, I should have proof-read it before posting 😅

No problem at all!

PR #1105, opened 2025-04-04
(send_kcidb.py) Add qualcomm to allowed users

nuclearcat:
As Qualcomm use Maestro to submit results, we need to allow kcidb bridge to process their records.


nuclearcat:
Works fine on staging, patch is trivial, should not break anything

PR #1107, opened 2025-04-06
Fix validation for trees within build_configs

kwilczynski:
Commit c00b656 ("(tests/validator) Improve build_configs verification") added validation of the `tree` and `branch` attributes to the `build_configs` property.

However, if a referenced tree is missing, the validation code will terminate with an error as the variable used is not a dictionary but a string type, and as such, the `.get()` method cannot be called.

Thus, fix which variables are referenced and the order of validation for the `tree` and `branch` attributes to validate their presence within a given tree object first.

While at it, update the accompanying code comment.

Related:

- https://github.com/kernelci/kernelci-pipeline/pull/1104

### DIFF
--- a/tests/validate_yaml.py
+++ b/tests/validate_yaml.py
@@ -100,24 +100,24 @@ def validate_unused_jobs(data):
 
 def validate_build_configs(data):
     '''
-    Each entry in build_configs have a tree parameter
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
LGTM

PR #1109, opened 2025-04-08
Sort list of trees alphabetically for build-only jobs

kwilczynski:
Sort the list of trees alphabetically for build-only jobs, and add a note about preserving the sort order.

Related:

- https://github.com/kernelci/kernelci-pipeline/pull/1100

### DIFF
--- a/config/jobs.yaml
+++ b/config/jobs.yaml
@@ -75,9 +75,10 @@ _anchors:
 
 ### Frequently used rules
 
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
LGTM

kwilczynski:
> I wonder if we can enforce sorted order in tests/validator

@nuclearcat, a nice thought. We definitely can. :smile:

PR #1111, opened 2025-04-10
Move entries for anchors correctly into the _anchors section

kwilczynski:
Commit cfe037a ("Add architectures filter to various trees") added architectures filters to a variety of trees and, while doing so, also introduced a utility tree called `_base`, which is then used to provide configuration anchors to facilitate reuse.

However, this utility tree will also trigger jobs to be run, which is not desired, as these extra anchors are only used to reduce configuration duplication.

Thus, move the `base`, `base-standard`, and `base-arm` anchors into the `_anchros` section, which will only be used to assemble the configuration and, as such, will prevent any extra jobs from being run.

### DIFF
--- a/config/trees.yaml
+++ b/config/trees.yaml
@@ -1,7 +1,28 @@
-trees:
+_anchors:
 
-  _base:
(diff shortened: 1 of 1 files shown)


pawiecz (approved):
LGTM

PR #1115, opened 2025-04-12
Add python requirements to the root

nuclearcat:
We have now requirements buried in the docker directory, which is planned to be deleted. Let's move it to root directory. We still need them for kernelci-core recipes.


kwilczynski (approved):
Looks good!

PR #1117, opened 2025-04-15
Remove deprecated library kernelci.db

nuclearcat:


nuclearcat:
Fixes restarting services

PR #1132, opened 2025-04-25
Extract basic .kernelci.yml file template

pawiecz:


pawiecz:
This PR produces Maestro Pipeline configuration equivalent to what has been used before applying it with following changes:
- `base*` anchors are no longer used and were removed
- base `broonie` configuration is promoted to an anchor instead of being a `build_config` following the `mainline:master` tree

It can be verified with:
```
tests/validate_yaml.py -o new
git checkout HEAD~2
tests/validate_yaml.py -o old
$difftool old new
```

PR #1134, opened 2025-04-25
(jobs.yaml) Add back as example cip:// config fragment

nuclearcat:
This is initial commit to return CIP specific fragments, and also will serve as verification that everything works right for this fragments.


kwilczynski (approved):
Looks good!

PR #1131, opened 2025-04-24
Add kbuild job with coverage enabled and run tests using this kernel

a-wai:
As we're enabling code coverage using GCOV in KernelCI, we must ensure we have a kernel with the corresponding options enabled using the `coverage` fragment, and also run some tests using this kernel.

The LTP tests we run on x86 Chromebooks are good candidates for helping during developement.

PR #1140, opened 2025-05-04
Feat: send input files ref to kcidb, add input files to maestro "artifacts" field

tollsimy:
This PR complements what added in https://github.com/kernelci/kcidb-io/pull/99 and https://github.com/kernelci/kcidb/pull/635, allowing to send reference to input files to kcidb as well as input files to "artifacts" field in maestro.

PR #1141, opened 2025-05-06
src: lava_callback: handle artifacts uploaded from test jobs

a-wai:
Test jobs might need to upload artifacts directly from LAVA (this is needed e.g. for GCOV data when checking for code coverage), but we currently have no way to add those artifacts to the corresponding node.

Add such a feature in a rather generic way, so jobs can provide both the artifact name and its URL through a specially-formatted result entry:

  `artifact-upload:<artifact name>:<artifact URL>`

Therefore, all test results starting with `artifact-upload:` will be processed and stored in the corresponding node's `artifacts` dict, with `<artifact name>` being the entry key and `<artifact URL>` being its value.

Depends on #1159 & https://github.com/kernelci/kernelci-core/pull/2878

### DIFF
--- a/src/lava_callback.py
+++ b/src/lava_callback.py
@@ -233,6 +233,16 @@ def async_job_submit(api_helper, node_id, job_callback):
         job_node['error_msg'] = None
     if device_id:
         job_node['data']['device'] = device_id
+    # add artifacts uploaded from the running LAVA job
+    upload_result = results.pop('upload', {})
+    for (name, state) in upload_result.items():
+        if name.startswith("artifact-upload:") and state == 'pass':
+            artifact = name.split(':', 2)
+            if len(artifact) != 3:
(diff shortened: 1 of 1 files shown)


tollsimy:
Hi @a-wai could it be solved by https://github.com/kernelci/kernelci-pipeline/pull/1140 ?
We added a way to introduce input artifacts in https://github.com/kernelci/kcidb-io/pull/99, but I am not sure it will fix your issue.

tollsimy:
@a-wai Got it, that makes sense, thanks!

a-wai:
@nuclearcat only manually tested for now, we should have more results soon now that I've (hopefully) fixed #1131

a-wai:
Coverage data successfully retrieved on the latest staging run, see [this node](https://staging.kernelci.org:9000/viewer?node_id=682dbd14427748cdff14ac9b) for an example.

PR #1144, opened 2025-05-08
Add kbuild-gcc-12-arm64-chromebook-media-committers to scheduler entries

nuclearcat:
To execute kbuild, it needs to be added to scheduler entries.


kwilczynski (approved):
Good catch! Thank you!

PR #1150, opened 2025-05-12
(send_kcidb.py) Fix logspec trigger by status

nuclearcat:
Due deprecation of field 'valid' in kcidb schema, we lost trigger condition when 'valid' == false was triggering logspec. We can trigger same as on tests, by status == FAIL, which was equivalent to `valid` retrieved by status map.


nuclearcat:
Tested on staging

PR #1159, opened 2025-05-21
src: lava_callback: improve logging

a-wai:
As this service is actually just a FastAPI app, only the use of the uvicorn logger can allow us to print messages.

Take this opportunity to replace undifferenciated `print`s with proper logs w/ severity and set log level to `debug`` by default.


a-wai:
Merging as it works as expected, see logs for `kernelci-pipeline-lava-callback` on https://staging.kernelci.org:9088

PR #1160, opened 2025-05-21
config: jobs: use new LTP rootfs

a-wai:
We now have GCOV-enabled LTP test runs, which require a new rootfs in order to upload artifacts directly from the DUT in LAVA.

### DIFF
--- a/config/jobs.yaml
+++ b/config/jobs.yaml
@@ -65,7 +65,7 @@ _anchors:
     kind: job
     params: &ltp-params
       boot_commands: nfs
-      nfsroot: 'https://storage.kernelci.org/images/rootfs/debian/bookworm-ltp/20240313.0/{debarch}'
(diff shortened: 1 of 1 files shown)


a-wai:
Draft for now as said rootfs is only available on staging and only for amd64.

To be merged only once https://github.com/kernelci/kernelci-core/pull/2878 is in and rootfs is generated and deployed in production for both amd64 & arm64.

a-wai:
Successfully tested in https://lava.collabora.dev/scheduler/job/18721559 (using the staging-deployed rootfs), will update this PR once https://github.com/kernelci/kernelci-core/pull/2878 and the corresponding rootfs' have been built and deployed to production.

a-wai:
New rootfs deployed in production @ https://storage.kernelci.org/images/rootfs/debian/bookworm-ltp/20250523.0/

Manually tested in [this LAVA job](https://lava.collabora.dev/scheduler/job/18747143)

PR #1161, opened 2025-05-21
Remove regression tracker from K8S, its deprecated

nuclearcat:


tollsimy:
@nuclearcat Changes in `tools/jwt_generator.py` are generating not working kci-dev tokens on local deployment, probably in staging too.

nuclearcat:
I will fix that, just noticed, edited it by mistake

PR #1165, opened 2025-05-23
config: scheduler*: change job trigger for available kbuilds

a-wai:
Until now, test jobs were triggered when corresponding `kbuild` nodes
were changed to `result: pass`. This made sense as those nodes would
immediately and always transition to state `done` upon completion.

As they're now going through the `available` state when successful, we
can use this event instead for triggering test jobs.

Make this change throughout all scheduler configs, improving
`scheduler-chromeos` through the use of a corresponding anchor in the
process.

Depends on https://github.com/kernelci/kernelci-core/pull/2882

### DIFF
--- a/config/scheduler-chromeos.yaml
+++ b/config/scheduler-chromeos.yaml
@@ -44,6 +44,11 @@ _anchors:
     runtime:
       name: k8s-all
 
+  kbuild-event: &kbuild-event
+    channel: node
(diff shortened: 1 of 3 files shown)


a-wai:
No, as kbuilds will transition to `available` only if the build passes, otherwise it'll go directly to `done` (same as checkout nodes, actually)

PR #1174, opened 2025-05-29
State hotfix and linter addition

nuclearcat:
Right now, for example kbuild might appear with result: done, in states: available, closing, done
This might trigger jobs that rely only on result - 3 times.
We need to fix that and add verification in linter.


nuclearcat:
Tested on staging

PR #1176, opened 2025-06-02
config: scheduler-chromeos: Add intel platforms for fault injection testing

crazoes:
Currently, fault injection test is only running on one platform, add specific platforms for the test to run on them.


nuclearcat (approved):
LGTM

PR #1177, opened 2025-06-03
doc: update to reflect latest changes in state transitions

a-wai:
We no longer transition successful `kbuild` nodes directly to `done` state, but go through `available` first. This state triggers test jobs' creation, while we can schedule post-processing jobs when the node finally reaches the `done` state` (once every test job is complete, or the timeout is reached).

Ensure the documentation (especially the examples) reflects the current design.


nuclearcat:
If job failed, i will go also "available" state?

PR #1179, opened 2025-06-04
config: Enable the fchmodat2 selftests

broonie:
These are again small, software only tests.


pawiecz (approved):
LGTM

PR #1184, opened 2025-06-06
config: Split 32 and 64 bit Arm jobs in the Baylibre lab

broonie:
The configuration has a single baseline job for the 32 and 64 bit Arm
boards in the Baylibre lab, tied to the 32 bit Arm builds.  This
obviously doesn't work for the 64 bit boards, resulting in inconclusive
boots as the 64 bit boards don't understand zImage.  Split into a
separate baseline job like the other labs.


broonie:
@montjoie @khilman - I think I got the list of boards right here.

khilman (approved):
LGTM

PR #1187, opened 2025-06-06
config: Add breakpoints selftest

broonie:
This does have some hardware dependencies, the number and kind of
physical breakpoints can vary.


nuclearcat (approved):
LGTM

PR #1192, opened 2025-06-10
(kbuild.jinja2) Get return code properly

nuclearcat:
In Python, os.system() returns the exit status in a platform-specific format. On Unix-like systems, it returns the exit status shifted left by 8 bits. Here's what's happening:
- My shell script exits with code 1 (build failure)
- os.system() returns 256 (1 << 8)


nuclearcat:
Tested on staging

PR #1198, opened 2025-06-16
config: trees: Drop nonexistent Build configuration for linux-pci tree

pawiecz:
Branch "for-kernelci" does not exist for this tree.

### DIFF
--- a/config/trees/linux-pci.yaml
+++ b/config/trees/linux-pci.yaml
@@ -10,7 +10,3 @@ build_configs:
   linux-pci_for-linus:
     <<: *linux-pci
     branch: 'for-linus'
-
-  linux-pci_for-kernelci:
-    <<: *linux-pci
-    branch: 'for-kernelci'


nuclearcat (approved):
LGTM

kwilczynski:
I believe this change needs to be reverted.

> Branch "for-kernelci" does not exist for this tree.

@pawiecz, this is as expected? We, the PCI maintainers, add and remove branches as we see fit, and we add "for-kernelci" on demand, once there is something to be added to it. This is especially true during the merge window, where we update both the "next" and "for-kernelci" branches.

When I was more involved in the project just recently, I discussed the need for branches to be there in length with @nuclearcat, who also allegedly fixed the code that was failing where the branch wasn't there. Or, was that not really fixed correctly?

Moreover, why hasn't the owner of this branch, or the original author of this change, been notified about it?

Don't you think that it would be prudent to let the users know you are going to remove the configuration they rely on?

Please revert this change and restore the "for-kernelci" branch support for the PCI tree.

Thank you!

PR #1199, opened 2025-06-16
docker-compose-lava: Comply with the new Staging deployment process

pawiecz:
Docker image for Service "lava_callback" had previously been built on-demand during Staging instance deployment. This has been changed in the new Staging deployment workflow.

This patch uses prebuilt Docker image for "lava_callback" Service. Docker image build context will be dropped in the future if it's deemed unnecessary from development point of view as well.


nuclearcat:
Thanks for this important patch!

PR #1203, opened 2025-06-17
config: Update to LTP 20250530 release

broonie:
Update to the 20250617.0 rootfs build, which switches to (and ensures
that we stick to) the LTP 20250530 release tag.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -120,7 +120,7 @@ _anchors:
     kind: job
     params: &ltp-cros-kernel-params
       boot_commands: nfs
-      nfsroot: 'https://storage.kernelci.org/images/rootfs/debian/bookworm-ltp/20250523.0/{debarch}'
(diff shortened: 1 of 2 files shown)


pawiecz (approved):
LGTM

@laura-nao @a-wai Please let me know when you get a chance if this change is ok from ChromeOS point of view as well

a-wai:
No objection from me.

a-wai (approved):

broonie:
There appears to be some issue with the standard LTP runner in this release when used in conjunction with our rootfs: https://lava.sirena.org.uk/scheduler/job/1481791 - since the standard runner is deprecated and we want to switch to Kirk anyway it's probably more productive to do that rather than diagnose and fix the issue. I've created https://github.com/kernelci/kernelci-core/pull/2907 for that, I'll rebuild the rootfs images and update this PR to reference them once that's got into staging. The two PRs will need to be merged together, we need both the LAVA template and the new rootfs images.

nuclearcat:
Merging this and https://github.com/kernelci/kernelci-core/pull/2907 , let me know if i missed anything

PR #1206, opened 2025-06-18
Switch Trigger Service main loop data structure

pawiecz:
Build configurations list will not be known up front when external configuration files are enabled in Maestro. Dropping `build_configs` filter will allow simplifying Trigger Service logic taking into account only parameters that are still relevant.


JenySadadia (approved):
LGTM.

PR #1217, opened 2025-06-24
config: Enable more tests on the Orion O6

broonie:
- config: Add Radxa Orion O6
- config: Enable more tests on the Orion O6


pawiecz:
LGTM

PR #1221, opened 2025-06-24
config: Move some jobs off Avenger96

broonie:
Especially when the stable releases come out the Avenger96 can build up
quite a backlog, move a lot of the smaller single threaded jobs over to
Beaglebone Black which has more capacity in my lab.


pawiecz (approved):
LGTM

PR #1225, opened 2025-06-27
tests/validate_yaml: Implement walker to emulate checkout and forecast builds

nuclearcat:


tests/validate_yaml.py
```diff
@@ -194,16 +198,148 @@ def dumper(o_filename, merged_data):
...
+            if VERBOSE:
+                print(
+                    f"Validating scheduler entry for job: {job_name} with scheduler rules"
+                )
+            if not validate_rules(node, scheduler_rules):
+                if VERBOSE:
+                    print(f"Skipping job: {job_name} due to scheduler rules")
+                continue
+            checkout["kbuilds"].append(job_name)
+        checkout["kbuilds_identical"] = compare_builds(merged_data, checkout["kbuilds"])
```
aliceinwire, line 323: compare_builds() takes 1 positional argument but 2 were given

tests/validate_yaml.py
```diff
@@ -194,16 +198,148 @@ def dumper(o_filename, merged_data):
...
-    print("Options:")
-    print("-d, --dir       Directory to validate yaml files (default: config)")
-    print("-o, --output    Output file to dump merged yaml file without anchors")
-    print("-h, --help      Show this help message and exit")
-    sys.exit(1)
+def validate_rules(node, rules):
+    """
+    Validate rules for a given node
+    """
+    import kernelci.api.helper
```
aliceinwire, line 205: https://peps.python.org/pep-0008/
```
Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.”
```

aliceinwire (changes requested):

tests/validate_yaml.py
nuclearcat: It is intentional, avoiding the cost of loading a heavy dependency that is used in just one rare code path.
Because kernelci.api.helper have a lot of dependencies, and this particular function of forecasting for now not planned to be used in CI, it is more manual tool to "forecast" builds. 
If i put in top, it will require to rework current workflow and include kernelci-core as dependency (and slow down whole process, making it also more fragile).

aliceinwire:
> > would be nice to have kernelci-core as sub-module instead of searching from which repository are you getting kernelci.api.helpers
> 
> I will try to implement lazy loader function that will give more info on failure.

As kernelci-core kernelci import is used extensively throughout kernelci-pipeline (https://github.com/search?q=repo%3Akernelci%2Fkernelci-pipeline%20import%20kernelci.api&type=code )
instead of lazy import it should be a required import

nuclearcat:
Well, for me making dependencies for this validator also no-go, this will require to rework workflows, make them slower, and worst part i cannot use this tool as standalone config validator. 
I will port some of changes from this PR, and will try to implement separate tool for forecasting builds.

PR #1234, opened 2025-07-06
config: Enable AM62 Mallow in my lab

broonie:
Toradex have been kind enough to send me an AM62 Mallow system, enable
some basic tests on it in my lab.


pawiecz (approved):
LGTM

PR #1236, opened 2025-07-08
config: Restore baseline-nfs kind of tests

nuclearcat:
During migration we lost one kind of tests, baseline-nfs, which are different from baseline by following:
1)Uses debos, instead of buildroot
2)Have networking enabled and used, for nfs rootfs


nuclearcat:
It didnt worked yet (still not nfs boot), will try to make it work next week.

PR #1240, opened 2025-07-09
config/scheduler.yaml: Enable device error log test in my lab

broonie:
Run the device error logs test on all the boards in my lab.  It would
probably be interesting to consider adding this to the end of other test
jobs as boilerplate in case the other tests trigger any issues.

While we're at it sort the existing entry for these tests in the
Collabora lab.


nuclearcat:
Can you rebase please

aliceinwire (approved):

PR #1244, opened 2025-07-09
config: rework all coverage-enabled jobs

a-wai:
Our only planned user for code coverage analysis is the ChromeOS kernel team, so this PR fully reworks the coverage-enabled builds and test jobs.

To sum things up, the current, generic x86 kernel build is replaced by 4 distinct builds:
* 2 for Intel & AMD ChromeBooks (x86)
* 2 for Qualcomm & Mediatek ChromeBooks (arm64)

Those builds use the following config fragments on top of the usual ones:
* `coverage` for enabling GCOV support in the kernel
* `kselftest` so we can run the corresponding tests on those kernels (will be enabled as a follow-up PR)
* `crypto` so those kernels can also execute the `ltp-crypto` job

All LTP jobs currently executed on the ChromeOS kernel are now also scheduled for those new builds, replacing the previous LTP scheduler entries (those were using the generic x86 build which is phased out in this PR).

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -126,8 +126,6 @@ _anchors:
       extra_kernel_args: "lsm=capability,landlock,yama,loadpin,safesetid,selinux,bpf"
     kcidb_test_suite: ltp
     rules: &ltp-cros-kernel-rules
-      fragments:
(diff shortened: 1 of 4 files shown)


a-wai:
Created as draft for now as it requires more testing

PR #1245, opened 2025-07-09
config: use new rootfs for kselftest jobs

a-wai:
This rootfs contains `curl`, which is required for uploading coverage data from LAVA jobs.


a-wai:
Updated with links to the (currently building) new rootfs from the production instance.

nuclearcat:
congrats!

PR #1253, opened 2025-07-11
scheduler: implement /health endpoint for k8s and docker

nuclearcat:
Sometimes process might get stuck, and this compromise whole idea of scheduler, as events will get lost and etc.
We can implement http server thread that will watch for variable that is updated every "loop". If variable not updated anymore, loop is not working anymore.

Depends-on: https://github.com/kernelci/kernelci-core/pull/2923

### DIFF
--- a/src/scheduler.py
+++ b/src/scheduler.py
@@ -15,6 +15,8 @@ import requests
 import re
 import datetime
 import time
+import threading
+from http.server import HTTPServer, BaseHTTPRequestHandler
 
 import kernelci
 import kernelci.config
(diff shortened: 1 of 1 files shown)


kube/aks/scheduler-k8s.yaml
```diff
@@ -68,22 +92,6 @@ spec:
             - name: tmpsecrets
               mountPath: /home/kernelci/.azure
               subPath: k8s-credentials/.azure
-      initContainers:
```
pawiecz, line 71: Please correct me if I'm wrong: this change is just a cleanup to group both `initContainers` together, not a functional one?

nuclearcat: Yes, i just spotted mistake when i was editing file. Not best approach to mix in PR, but at least did as separate commit. I was afraid yaml failure might lead to other failures, e.g. to read health section.

PR #1254, opened 2025-07-15
src/timeout: fix logic marking nodes as "incomplete"

JenySadadia:
Do not mark timed-out node as "incomplete" if the result is set to "fail" or "skip".

Fixes: fe42dd5 ("src: timeout: make state transitions more predictable")

### DIFF
--- a/src/timeout.py
+++ b/src/timeout.py
@@ -75,7 +75,8 @@ class TimeoutService(Service):
             if mode == 'TIMEOUT':
                 if 'data' not in node_update:
                     node_update['data'] = {}
-                if node.get('result') != 'pass':
+                if not node.get('result'):
+                    # Set result as "incomplete" only if any value is not set yet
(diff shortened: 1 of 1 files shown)


JenySadadia:
I was investigating baseline jobs with infra issue and came to a scenario where its `setup` node and child nodes of `setup` were marked as `fail` initially, but then the `timeout` service changed them to "incomplete".

Job node: https://api.kernelci.org/viewer?node_id=686dabcc34612746bbb7c727
Job link: https://lava.ciplatform.org/scheduler/job/1289216

Here are the logs stating `setup` was in `fail` state:
```
INFO:     Callback data uploaded to https://files.kernelci.org/baseline-x86-cip-686dabcc34612746bbb7c727/lava_callback.json.gz
DEBUG: Setup failed for 686dabcc34612746bbb7c727. Transitting job node result: incomplete -> fail
```

I fixed this in the timeout service.

PR #1262, opened 2025-07-24
config: chromeos: enable kselftest for coverage-enabled builds

a-wai:
For now, we only run a (very) limited set of kselftests. However, enabling more of those would allow for greater code coverage. This patch creates kselftest jobs suited to running on the ChromiumOS kernel (in particular, ensuring we don't enable the ChromiumOS downstream LSM, which doesn't work with a Debian userspace).

Those tests are triggered by coverage-enabled kbuilds only, so we can eventually lower the frequency they run.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -115,6 +115,28 @@ _anchors:
       tree:
         - chromiumos
 
+  kselftest: &kselftest-job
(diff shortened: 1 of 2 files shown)


JenySadadia:
Replaced with https://github.com/kernelci/kernelci-pipeline/pull/1323

PR #1264, opened 2025-07-29
Revert "Fix re-subscription for pubsub events"

nuclearcat:
Reverts kernelci/kernelci-pipeline#1258

### DIFF
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -67,9 +67,6 @@ class KCIDBBridge(Service):
             if isinstance(runtime_configs, RuntimeLAVA):
                 self._lava_labs[runtime_name] = runtime_configs.url
         self._current_user = self._api.user.whoami()
-        self._filters = {
-            'state': ('done', 'available'),
-        }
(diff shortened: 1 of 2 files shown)


JenySadadia:
The issue on the staging was due to scheduler service changes for enabling job retry.
On the other hand, we need to remove re-subscription logic from all the services as when `self._api_helper.receive_event_node` fails, there must be some issue with extracting node data from the event rather than an issue with the subscription.
Re-subscribing should only be performed if `self._api_helper.subscribe_filters` or `self._api_helper.subscribe` fails.

nuclearcat:
> The issue on the staging was due to scheduler service changes for enabling job retry. On the other hand, we need to remove re-subscription logic from all the services as when `self._api_helper.receive_event_node` fails, there must be some issue with extracting node data from the event rather than an issue with the subscription. Re-subscribing should only be performed if `self._api_helper.subscribe_filters` or `self._api_helper.subscribe` fails.

thanks i will close pr

PR #1270, opened 2025-08-01
tools/extract_compatible: Fix extract_compatible path for linux sources

aliceinwire:
This patch fixes the directory of the linux sources files, for when the extract_compatible script is called from a different directory than tools/

### DIFF
--- a/tools/extract_compatible.py
+++ b/tools/extract_compatible.py
@@ -16,13 +16,16 @@ import subprocess
 import yaml
 import re
 
+script_path = os.getcwd()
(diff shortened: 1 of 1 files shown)


tools/extract_compatible.py
```diff
@@ -16,21 +16,25 @@
 import yaml
 import re
 
+script_path = os.getcwd()
+linux_path = os.path.join(script_path + "/linux")
+print(linux_path)
```
aliceinwire, line 21: forget to remove, will remove on the next update.

pawiecz (approved):
LGTM

PR #1273, opened 2025-08-07
config: scheduler-chromeos: enable Tast tests for gcov-enabled builds

a-wai:
Draft for now as this PR only contains sample entries for debugging


a-wai:
Successfully [tested on staging](https://lava.collabora.dev/scheduler/job/19588612). Setting `staging-skip` though as it depends on https://github.com/kernelci/kernelci-core/pull/2940 (which has `staging-skip` for now).

Also keeping as draft as it only contains a single enabled test.

Issue #1276, opened 2025-08-12
Address deprecation warnings


JenySadadia:
Update pipeline services to address deprecation warnings as below:
```
/home/kernelci/pipeline/./src/tarball.py:177: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
  'holdoff': str(datetime.utcnow() + timedelta(minutes=10))
```
```
/home/kernelci/pipeline/src/timeout.py:119: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
```
```
/home/kernelci/pipeline/./src/trigger.py:105: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
```

PR #1277, opened 2025-08-13
Address deprecation warnings for `datetime.datetime.utcnow()`

JenySadadia:
Closes https://github.com/kernelci/kernelci-pipeline/issues/1276

Update pipeline services to use `datetime.datetime.now(datetime.UTC)` in place of deprecated method i.e., `datetime.datetime.utcnow()`.


JenySadadia:
Tested OK on staging.

PR #1288, opened 2025-08-21
src: send_kcidb: process coverage-report nodes

a-wai:
Coverage report nodes are meant to provide global coverage percentages for a while `kbuild`, aggregating the results from individual test jobs. As such, they shouldn't be sent to KCIDB as nodes, but rather parsed in order to submit an updated build node which includes the functions/lines coverage percentages as misc fields.

Add a new processing function for such nodes, which are of kind `process` and are named `coverage-report-<arch>`.

### DIFF
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -567,6 +567,7 @@ in {runtime}",
         return parsed_test_node, dummy_build
 
     def _parse_coverage_node(self, origin, node):
+        self.log.debug(f"Parsing coverage node: {node['id']}")
         # Coverage report nodes are not sent to the DB themselves, but rather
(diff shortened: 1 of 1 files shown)


JenySadadia:
The KCIDB bridge is sending coverage data but KCIDB isn't able to store it since KCIDB doesn't update field data upon re-submission.
In this case, `misc` is updated for an already existing build node.
Alternate solution for the time being: https://github.com/kernelci/kci-dev/pull/214

JenySadadia:
Once https://github.com/kernelci/kcidb/issues/654#issuecomment-3356578298 is implemented, we will be able to see data on KCIDB.

PR #1297, opened 2025-09-03
Retry failed boot jobs

JenySadadia:


src/job_retry.py
```diff
@@ -52,6 +52,11 @@ def _run(self, sub_id):
                 self.log.error(f"Error receiving event: {e}")
                 continue
 
+            # Only retry failed job if it's a boot job
```
aliceinwire, line 55: yes, this make sense

PR #1298, opened 2025-09-03
validate_yaml: Catch missing keyword runtime on jobs

aliceinwire:


nuclearcat (approved):
Looks great, thank you very much!

PR #1300, opened 2025-09-04
Do not timeout coverage process since it's taking long

JenySadadia:


JenySadadia:
Issue is indentified. Timeout is not an issue.

PR #1302, opened 2025-09-08
config: chromeos: pass probability value

vigneshraman:
Pass a random probability value between 1 and 100 for the fault-injection-cros-kernel job. This ensures each job run uses a different probability value.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -464,6 +464,7 @@ _anchors:
       test_method: fault-injection
       test_name: suspend-resume
       test_command: 'rtcwake -m mem -s 15'
+      probability: "{{ range(1, 101)|random }}"
       boot_commands: nfs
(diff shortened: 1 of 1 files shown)


laura-nao (approved):
If we end up needing this on all fault injection tests, it might be worth changing the default in the test job template itself: https://github.com/kernelci/kernelci-core/blob/main/config/runtime/tests/fault-injection.jinja2#L26

But for now LGTM, thanks!

vigneshraman:
> If we end up needing this on all fault injection tests, it might be worth changing the default in the test job template itself: https://github.com/kernelci/kernelci-core/blob/main/config/runtime/tests/fault-injection.jinja2#L26
> 
> But for now LGTM, thanks!

Ack. I will modify fault-injection.jinja2 and update the other PR https://github.com/kernelci/kernelci-core/pull/2964

vigneshraman:
Closing in favor of https://github.com/kernelci/kernelci-core/pull/2964

PR #1303, opened 2025-09-08
platforms: Add support for SiFive HiFive Premier P550

pawiecz:
This PR currently depends on #1267 - it uses changes introduced by that PR. It can be made an independent PR if merging its current parent has to be postponed ~~until the root cause of [the error](https://github.com/kernelci/kernelci-pipeline/pull/1267#issuecomment-3266961592) is found~~.

### DIFF
--- a/config/platforms.yaml
+++ b/config/platforms.yaml
@@ -98,6 +98,12 @@ platforms:
 
   docker:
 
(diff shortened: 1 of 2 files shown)


nuclearcat:
Seems PR need rebasing

PR #1311, opened 2025-09-16
config: add support for ramdisk based kselftest pipeline

quic-ylal:
Some platform does not support nfs boot which prevents these platforms to enable kselftest pipeline.
This change adding support for ramdisk-based kselftest pipeline by adding a new job definition.


nuclearcat:
Wonderful job! Thank you very much!

PR #1314, opened 2025-09-22
Send final retry of failed boot jobs

JenySadadia:
Update KCIDB bridge to only send final retry i.e. job with `retry_counter=3` for failed boot jobs.


JenySadadia:
Tested OK on staging.

PR #987, opened 2025-01-31
Enable more devices from the Pengutronix lab

hnez:
This PR is based on https://github.com/kernelci/kernelci-pipeline/pull/986 and contains some of the same commits.
It is split-off since getting all the devices working may require some back-and-forth.

This adds all devices from the [Pengutronix LAVA Lab](https://lava.pengutronix.de/scheduler/alldevices/active) that currently pass their health check jobs.

The lab contains mostly older i.MX devices that we want to prevent from bit-rotting in the mainline kernel, but also some newer devices like i.MX8MM, i.MX8MP or the STM32MP1 based LXA TAC.

Closes: https://github.com/kernelci/kernelci-core/pull/2542 (Adding stm32mp157c-lxa-tac-gen1 support to the legacy infrastructure).

PR #1316, opened 2025-09-24
Add devices currently in lab at pengutronix

gumulka:
See https://lava.pengutronix.de/scheduler/alldevices/online for an overview

Close to #987, but updated to the new structure with splitted yaml files and without the one non-working board. Will add that later.

Closes #987

### DIFF
--- a/config/platforms.yaml
+++ b/config/platforms.yaml
@@ -104,6 +104,13 @@ platforms:
     dtb: dtbs/eswin/eic7700-hifive-premier-p550.dtb
     compatible: ['sifive,hifive-premier-p550', 'eswin,eic7700']
 
+  imx23-olinuxino:
+    <<: *arm-device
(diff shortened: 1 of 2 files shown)


nuclearcat:
Hi! If you have new hardware platform not defined in [platforms.yaml](https://github.com/kernelci/kernelci-pipeline/blob/main/config/platforms.yaml), you need to add it there, this is what linter complaining about.
For compatibles option we have a tool that can assist with that: [extract_compatible.py](https://github.com/kernelci/kernelci-pipeline/blob/main/tools/extract_compatible.py)

gumulka:
Yeah, it was defined, but there was an indentation error.

JenySadadia:
I approved it. Thanks.

nuclearcat:
Let's merge it? Seems all ok

PR #1318, opened 2025-09-25
config: Update to wifi 20250925 release

vigneshraman:
Update to the 20250925.0 rootfs build, which includes fixes to load iwlmvm before iwlwifi. This ensures the wlan interface is created for intel boards.


nuclearcat (approved):
LGTM

PR #1323, opened 2025-10-02
config: chromeos: enable kselftest for coverage-enabled builds

JenySadadia:
Replaced https://github.com/kernelci/kernelci-pipeline/pull/1262 with this one since it needed rebase.

For now, we only run a (very) limited set of kselftests. However, enabling more of those would allow for greater code coverage. This patch creates kselftest jobs suited to running on the ChromiumOS kernel (in particular, ensuring we don't enable the ChromiumOS downstream LSM, which doesn't work with a Debian userspace).

Those tests are triggered by coverage-enabled kbuilds only, so we can eventually lower the frequency they run.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -115,6 +115,28 @@ _anchors:
       tree:
         - chromiumos
 
+  kselftest: &kselftest-job
+    template: generic.jinja2
+    kind: job
+    params: &kselftest-params
+      test_method: kselftest
+      boot_commands: nfs
+      nfsroot: 'https://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20250724.0/{debarch}'
+      job_timeout: 10
+    kcidb_test_suite: kselftest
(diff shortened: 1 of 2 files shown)


JenySadadia:
Deployed this on staging. 
All the jobs ended up with `submit_error`.
See https://staging.kernelci.org:9000/viewer?node_id=68df6c044ce74261cefc6cc5  and  https://staging.kernelci.org:9000/viewer?node_id=68df6bee4ce74261cefc6cbc

a-wai:
@JenySadadia IMHO the errors are a consequence of https://github.com/kernelci/kernelci-core/commit/df4d46d93eafcd109cd14cc86fe124d2a0a5b652 : Lava doesn't recognize `hatch` as a device type, but it does recognize `asus-C436FA-Flip-hatch`, for instance.

@nuclearcat what motivated this change and has it been thoroughly tested?

a-wai:
@JenySadadia Those errors should be fixed with #1329

PR #1338, opened 2025-10-16
src: send_kcidb: store `process` nodes as regular test nodes

a-wai:
Coverage report nodes are a new kind of node which should be sent to KCIDB as well. As we're mostly interested in providing an easy way to retrieve those nodes' artifacts, process them as regular test nodes.


a-wai:
Ok, so what would be needed, aside from adding a `kcidb_test_suite: coverage` to job definitions?

PR #1349, opened 2025-10-24
cron tools: change config variables location

JenySadadia:
Change config variables location for cron tool scripts such as `upload_log.py` and `email_sender.py`.
 
- Use `config/pipeline.yaml` to get storage configs
- Use `kernelci.storage` API to use storage upload API
- Use pipeline TOML config file i.e., `kernelci.toml` to define cron configs


JenySadadia:
Moved `email_password` config variable to TOML file.

PR #1365, opened 2025-11-18
Add support for CIP 4.4 and qemu-arm64

patersonc:
This PR technically adds two separate topics, but they depend on each other hence putting them together in one PR.

1) Enable builds using CIP configs for v4.4
2) Add support for qemu-arm64 (for all, not just CIP)
3) Enable boot testing on qemu/qemu-arm/qemu-arm64 for CIP trees

Note that none of this is tested yet. Hopefully kernelci-staging can help here.


bhcopeland:
LGTM

aliceinwire (approved):

PR #1368, opened 2025-11-20
CIP: Add RT enabled builds and tests

patersonc:
This PR does the following:

1. Commit 1: Adds a build for each CIP kernel/arch that enables the preempt_rt configuration fragment
2. Commit 2: Switches our default testing to only use the CIP reference platforms
3. Commit 3: Creates a cyclicdeadline test for each build from 1.
4. Commit 4: Creates a cyclictest test for each build from 1.

All done with heavy (hopefully correct) usage of anchors.

Thanks!

### DIFF
--- a/config/scheduler-cip.yaml
+++ b/config/scheduler-cip.yaml
@@ -48,6 +48,9 @@ scheduler:
     platforms: *cip-arm64-platforms
 
   - <<: *job-gcc-14-arm64-612-cip-rt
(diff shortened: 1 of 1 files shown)


patersonc:
Note that this PR is currently based on top of https://github.com/kernelci/kernelci-pipeline/pull/1382 and shouldn't be merged as-is.

aliceinwire (approved):

PR #1376, opened 2025-11-25
config: jobs-*: replace KernelCI containers with TuxMake containers

alok0366:
Replace the KernelCI build container images with the corresponding TuxMake images. This aligns the build environment with the toolchains and configurations provided by TuxMake and ensures consistent, reproducible kernel builds across architectures.

Key improvements over the previous KernelCI containers:
- Reproducible, uniform build environments across all architectures
- Lower maintenance burden with upstream-maintained images
- First-class support for cross-compilers, QEMU, and common kernel toolchains
- Stable, versioned interface optimized for CI workflows
- Better integration with modern CI systems and cloud-native pipelines

No functional changes to the build steps; only the container source is updated. Only dtbs_check affected.

### DIFF
--- a/config/jobs-chromeos.yaml
+++ b/config/jobs-chromeos.yaml
@@ -10,7 +10,7 @@ _anchors:
   kbuild-clang-21-arm64-chromeos: &kbuild-clang-21-arm64-chromeos-job
     template: kbuild.jinja2
     kind: kbuild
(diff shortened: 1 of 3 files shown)


nuclearcat:
Most likely best way will be to build kernelci containers based on this images, similar as we do with debos and existing compiling tools:
https://github.com/kernelci/kernelci-core/blob/main/config/docker/base/debos.jinja2
https://github.com/kernelci/kernelci-core/tree/main/config/docker

PR #1389, opened 2025-12-03
config/platforms: rename serial to console for qemu-arm and arm64

bhcopeland:


bhcopeland:
Confirmed working: https://lava.ciplatform.org/scheduler/job/1342337

PR #1396, opened 2025-12-11
Add --runtime-type parameter

nuclearcat:
To not add each lab into list and update k8s and docker-compose, we introduce new parameter - runtime-type.


nuclearcat:
Tested on staging

PR #1397, opened 2025-12-11
Add orion-o6 to pull-labs

bhcopeland:
The orion-o6 uses an r8169 NIC which isn't included in the generic KernelCI initrd. For NFS rootfs boots, the kernel needs to handle DHCP directly via IP_PNP rather than relying on the initrd.

Add netdev fragment with built-in r8169 driver and nfs-root-boot fragment enabling IP_PNP. The kbuild-gcc-14-arm64-nfsboot job combines these for LTP/smoketest jobs that mount rootfs over NFS.

Boot jobs use the standard kernel since they only need the ramdisk.


bhcopeland:
WIP

PR #1407, opened 2026-01-14
scheduler: add tree-based priority for LAVA jobs

bhcopeland:
Add tree priority lookup using string values (high/medium/low) instead
of numeric priorities. Tree priorities are read from build configs and
passed to LAVA runtime via node data.

Priority assignments:
- high: mainline, stable, stable-rc
- medium: next, stable-rt, kselftest
- low: android, cip, kernelci (default)

Human submissions always get highest priority for bisection/debugging.

Signed-off-by: Ben Copeland <ben.copeland@linaro.org>

### DIFF
--- a/config/jobs.yaml
+++ b/config/jobs.yaml
@@ -11,6 +11,7 @@ _anchors:
   baseline-job: &baseline-job
     template: baseline.jinja2
     kind: job
+    priority: medium
     kcidb_test_suite: boot
     params:
       ramdisk: 'http://storage.kernelci.org/images/rootfs/buildroot/buildroot-baseline/20230703.0/{brarch}/rootfs.cpio.gz'
@@ -18,6 +19,7 @@ _anchors:
   baseline-nfs-job: &baseline-nfs-job
     template: baseline-nfs.jinja2
     kind: job
+    priority: medium
     kcidb_test_suite: boot.nfs
     params:
       boot_commands: nfs
@@ -26,6 +28,7 @@ _anchors:
   kbuild-job: &kbuild-job
     template: kbuild.jinja2
     kind: kbuild
+    priority: high
     rules:
       tree:
       - '!android'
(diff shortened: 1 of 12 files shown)


patersonc:
Another query,

Would we be better to avoid using the "high" priority in the LAVA jobs?

In doing so labs won't have a way to inject jobs with a higher priority over incoming KernelCI jobs as "high" is the maximum priority level (in my understanding).

Maybe we could keep with "low, medium, high" in the kernelci yaml files, but translate that to something like "20, 50, 80" for the LAVA job definitions?

bhcopeland:
> Hi @bhcopeland
> 
> Thanks for the PR.
> 
> There are a few instances of `priority_min` and `priority_max` in `config/pipeline.yaml`. Do these need updating somehow? Or do these serve a different purpose?
> 
> Chris

No worries!

This is correct, this allows each lab to define proirty range. The approach scales between 0-100, to fit within each lab's range. IMO, they don't need updating; they serve different purposes.

> Another query,
> 
> Would we be better to avoid using the "high" priority in the LAVA jobs?
> 
> In doing so labs won't have a way to inject jobs with a higher priority over incoming KernelCI jobs as "high" is the maximum priority level (in my understanding).
> 
> Maybe we could keep with "low, medium, high" in the kernelci yaml files, but translate that to something like "20, 50, 80" for the LAVA job definitions?

You're right. If we use the full range (90-100), labs can't inject their own urgent jobs above KernelCI jobs.

My current values

```
  PRIORITY_HIGHEST = 90  # human submissions
  PRIORITY_HIGH = 75 # mainline, stable
  PRIORITY_MEDIUM = 50 # next, kselftest
  PRIORITY_LOW = 25 # android, cip
```

Suggested:

```
  PRIORITY_HIGHEST = 80  # human submissions (was 90)
  PRIORITY_HIGH = 60     # mainline, stable
  PRIORITY_MEDIUM = 40   # next, kselftest
  PRIORITY_LOW = 20      # android, cip

```
This will 81-100 for labs to use for their own urgent jobs.

bhcopeland:
https://lava.ciplatform.org/scheduler/job/1379868  I did a temp push with high priority to make sure it is set, and it worked fine.

PR #1410, opened 2026-01-16
tools: add script and CI workflow to sync stable-rc branches

bhcopeland:
Add sync_stable_branches.py which fetches kernel.org/releases.json and updates config/trees/stable-rc.yaml with current stable and longterm branches, removing any that have reached EOL.

Add GitHub Actions workflow that runs weekly to check for updates and automatically creates a PR when changes are needed.


nuclearcat:
We will have to test it IRL, as staging pipeline can't test workflows (permissions). Merging.

PR #1431, opened 2026-02-25
config: use qemu-x86_64 platform name instead of bare qemu

bhcopeland:
Replace bare 'qemu' platform references with 'qemu-x86_64' for consistency with the arch-prefixed naming convention used by other qemu platforms (qemu-arm, qemu-arm64, qemu-riscv64). Rename the platform alias in platforms.yaml from qemu-x86 to qemu-x86_64 to match the actual architecture.


bhcopeland:
Hello @patersonc 

Great spot, yes you are right, I have updated all those files too, they should also be included. Thanks!

PR #1438, opened 2026-03-03
rootfs: Update rootfs to trixie

nuclearcat:
We need to migrate to new trixie images


minash-ti:
@nuclearcat have trixie images been confirmed to work in test? Asking because I tried using them last week (https://storage.kernelci.org/images/rootfs/debian/trixie/) and saw booting issues. I haven't had a chance to triage this, so might be a misconfiguration on my side, but wanted to confirm it works on your side.

PR #1471, opened 2026-04-23
config: rpm aws builds

bhcopeland:
Add support for building kernel RPM packages targeting AWS EC2 instances:

- New aws-ec2 config fragment with EC2-critical drivers built-in (ENA, NVMe, XFS, virtio) so kernels boot without an initramfs
- New kbuild-gcc-14-x86-aws-ec2 job using the fragment and tuxmake's binrpm-pkg target to produce installable RPMs
- Scheduler entry to trigger builds on new checkouts
- Template update to pass extra_targets through to KBuild

Requires the matching extra_targets support in kernelci-core.

### DIFF
--- a/config/fragments.yaml
+++ b/config/fragments.yaml
@@ -903,3 +903,17 @@ fragments:
 
   x86_kvm_guest:
     path: "kernel/configs/kvm_guest.config"
+
+  aws-ec2:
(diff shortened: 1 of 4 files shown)


nuclearcat:
Tested on staging, seems at least not breaking anything. Thank you for all efforts!

PR #1485, opened 2026-05-15
Enable e850-96 to pull labs

roxell:


bhcopeland:
Let me test this on staging first that it comes up in events

PR #1487, opened 2026-05-15
config: restrict AWS/RPM builds to kernels >= 6.6

bhcopeland:
Add min_version 6.6 to both kbuild-gcc-14-{x86,arm64}-aws-ec2 so the jobs are skipped on stable/stable-rc branches since current it'll error out.


bhcopeland:
This will fix: https://dashboard.kernelci.org/issues?is=rpm

PR #1488, opened 2026-05-19
logspec: skip kbuild packaging-target failures

bhcopeland:
Kbuild packaging targets (binrpm-pkg, bindeb-pkg, etc.) fail with RPM/DEB tooling errors like "error: Bad exit status from /var/tmp/rpm-tmp.XXX (%install) in binrpm-pkg" — packaging infra failures, not kernel build errors. Logspec currently classifies these as kbuild.other and the random rpm-tmp filename in the summary causes many distinct issues to be filed for the same root cause, flooding the dashboard with infra noise.

Skip issue/incident generation for any logspec error whose target is in a known packaging-target set, alongside the existing linux.kernel.error_return_code skip.

### DIFF
--- a/src/kernelci_pipeline/logspec_api.py
+++ b/src/kernelci_pipeline/logspec_api.py
@@ -41,6 +41,33 @@ test_types = {
 }
 
 
+# Kbuild packaging targets: failures here are packaging-tool infra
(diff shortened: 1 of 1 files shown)


nuclearcat:
Wonderful, thank you @bhcopeland !

PR #1512, opened 2026-06-13
ci: validate KernelCI dashboard daily

### LINKED ISSUE #1125: Validate maestro build results consistency
In this task we want to improve build results consistency. For example: we have reports from users that sometimes Maestro reports less total builds to KCIDB for some checkouts. But we don’t have any way for the user to know what happened. 

To improve the build results reliability it is possible that we will have to debug and fix issues in maestro. This task will unfold in a few different subtasks. Here's a list of items to get started:

- [ ] Write a simple verifier that pulls data from the Dashboard API and compare builds results with the Maestro API. It should be a debug tool (maybe in kci-dev) to help us find inconsistencies in the data. 
- [ ] Such a tool should also compare the builds results of a given checkout with the previous (or another recent checkout) and look for mismatch in the number of total builds in each.
- [ ] Report Infrastructure builds as ERROR to KCIDB
- [ ] Figure out why some builds are disappearing and fix maestro accordingly
- [ ] Evaluate reporting builds that never executed as MISS to make sure we always report the same amount of results (we may not fix all the bug found through this effort immediately)

The main goal of this task is to make sure we are not losing build results anywhere and statuses (PASS, FAIL, ERROR or MISS) are reported with as much consistency as possible. It is not the goal of this task to improve the efficiency or resilience of our build process. Eg we don' t care at this point if a build fails with an infrastructure error, but we care if Maestro reports it wrongly to KCIDB or fails to report it at all.

nuclearcat:
Add a daily GitHub Actions workflow that validates recent Maestro builds, boots, and build history against dashboard records
It will run `kci-dev maestro validate` with `--json --fail-on-mismatch` so missing or inconsistent records fail visibly

## Notes

Temporary installs `kci-dev` from `nuclearcat/kci-dev@maestro-validate-json` while the upstream kci-dev JSON/fail-on-mismatch support is pending in kernelci/kci-dev#275.
Once that support is released, the workflow install step should switch back to the released package.

Fixes #1125


nuclearcat:
merging as its blocking staging

PR #1540, opened 2026-07-08
Durable pubsub subscribers

nuclearcat:


src/base.py
```diff
@@ -21,7 +21,7 @@ class Service:
     """Common class for pipeline services"""
 
     def __init__(self, configs, args, name):
-        self._name = name
+        self._name = getattr(args, "name", None) or name
```
bhcopeland, line 26: Do we want a unique name per instance (so multiple pods will work)?

nuclearcat: It is already easy to set:
```
kernelci-pipeline/docker-compose.yaml
  scheduler: &scheduler
...
    command:
...
      - '--name=scheduler'
```
is docker-compose for example, but that for different role schedulers (lava/shell/etc).

For multiple replicas/pods of the same service, they are generally not supported in current scheduler design (old or after this patch) and might need better approach, as they will likely schedule duplicate jobs/etc. We can design that easily if we reach need in that.

bhcopeland: Gothca, so looking at the current setup, it defaults to RollingUpdate, and there is no readinessProbe, before it kills the old one. So there is a small window where two pods are running the same --name, so the same sub_id. I guess that overlap could now drop or replay events, which didn't matter before with the fire and forget approach.

Do you think it makes sense to set Recreate strategy here? Then when old pods get torn down before a new one gets created, there will be a bit of downtime, but the durable subs should cover that gap, as anything during the gap should just get replayed on reconnect.

I'm not totally sure how the api handles two connections with the same sub_id (do they compete? or is it last one wins?). But, I feel Recreate here with this change feels a safe move either way.

nuclearcat: I'll add strategy: { type: Recreate } at least to the k8s deployments that consume durable subscriptions: schedulers, tarball, and pipeline-kcidb.

bhcopeland: LGTM, just add a DCO tag to the commit :)

PR #1541, opened 2026-07-09
validate: audit Maestro scheduling against the pipeline forecast

nuclearcat:
kci config forecast (kernelci-core) predicts which kbuilds each tree/branch checkout should produce and which test jobs each kbuild should trigger, but nothing verified that Maestro actually scheduled them.  Add a daily audit, following the same pattern as the dashboard validation (workflow + supplementary script in tests/):

  * tests/validate_forecast.py fetches checkout nodes old enough for their whole job tree to have settled (24h, over a 24h window), rebuilds the forecast graph with each checkout's real kernel revision, and compares expected vs actual kbuild children, then expected vs actual test jobs per (job, platform) for every kbuild that passed.  A missing child is only a finding when its parent succeeded: a failed checkout produces no kbuilds and a failed kbuild triggers no test jobs, so those absences are expected.

    Nodes that exist but are not forecast are classified before being flagged: externally-owned jobs (e.g. owner=qualcomm scheduling on runtimes outside our config) and runner-created children (the per-suite kselftest build nodes kbuild.py attaches) are counted as informational instead of reported as drift.

  * .github/workflows/validate-forecast.yml runs it daily, rewinding config/ to its git state at the start of the validated window so jobs added to the config since the checkout ran are not falsely reported as missing.  kernelci-core is checked out from the branch carrying the forecast graph support until it lands upstream.

A trial run against production caught a real gap noticed by @bhcopeland, missing pull lab jobs, while the same job ran on the same platforms two days earlier.

### DIFF
--- a/.github/workflows/validate-forecast.yml
+++ b/.github/workflows/validate-forecast.yml
@@ -37,6 +37,12 @@ jobs:
           # Full history so the config can be rewound to the window start.
           fetch-depth: 0
 
+      - name: Set up Python
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: '3.13'
+          cache: 'pip'
+
       - name: Rewind config to the start of the validated window
         # Checkouts up to (age + window) hours old are validated; using
         # today's config would flag jobs added since then as MISSING.
@@ -48,6 +54,7 @@ jobs:
           ref=$(git rev-list -1 --before="${hours} hours ago" HEAD)
           if [ -n "$ref" ]; then
             echo "Using config from $ref ($(git log -1 --format='%ci %s' "$ref"))"
+            rm -rf config
             git checkout "$ref" -- config
           else
             echo "No commit older than ${hours}h; keeping HEAD config"
@@ -56,17 +63,10 @@ jobs:
       - name: Check out kernelci-core (forecast tool)
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
-          # TODO: switch to kernelci/kernelci-core@main once the forecast
-          # graph support (evaluate_forecast_entry) lands upstream.
-          repository: nuclearcat/kernelci-core
-          ref: codex/forecast-missing-job-debug
+          # Pinned until the forecast graph API is part of a tagged release.
+          repository: kernelci/kernelci-core
+          ref: d2efb339f31f684dfd144366ddafdab2f7dea6c4
           path: kernelci-core
-
-      - name: Set up Python
-        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
-        with:
-          python-version: '3.13'
-          cache: 'pip'
 
       - name: Install dependencies
         # Only what `import kernelci.cli.config` pulls in, plus the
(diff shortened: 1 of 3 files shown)


tests/validate_forecast.py
```diff
@@ -0,0 +1,674 @@
...
+            )
+            if kind == "job"
+        }
+        actual_jobs = jobs_by_parent.get(kbuild.get("id"), [])
+        if not node_settled(kbuild):
+            report["unsettled_kbuilds"].append(describe_node(kbuild))
+            continue
+        if not node_ok(kbuild):
+            # A failed kbuild legitimately triggers nothing; only count it.
+            report["skipped_failed_kbuilds"].append(describe_node(kbuild))
```
bhcopeland, line 445: So we will surpress jobs that won't run, but will incomplete builds also be dropped? Might be worth a seperate counter for the other uses cases than just their children.

tests/validate_forecast.py
```diff
@@ -0,0 +1,674 @@
...
+        report,
+        pairs,
+        expected_root_jobs,
+        root_jobs,
+        source_name="checkout",
+        parent_id=checkout["id"],
+        parent_ok=node_ok(checkout),
+    )
+
+    for kbuild in kbuilds:
```
bhcopeland, line 431: I think this will miss 3 levels down, as it will check only its children groups. 

I.e.

```
  # Execute nipa-update on the blktests-ddp-x86 node
  # complete with the blktests-ddp-x86 event
  - job: nipa-update
    event:
      <<: *job-event
      name: blktests-ddp-x86
    runtime:
      type: shell
```

.github/workflows/validate-forecast.yml
```diff
@@ -0,0 +1,93 @@
...
+
+      - name: Install dependencies
+        # Only what `import kernelci.cli.config` pulls in, plus the
+        # validator's own deps; the heavy kernelci-core extras (azure,
+        # kubernetes, docker) are not needed for the forecast code.
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install click requests cloudevents toml pyyaml jinja2
+
+      - name: Validate scheduled jobs against forecast
```
bhcopeland, line 79: We should probably have a check here so MISSING_KBUILD does spam. I think for example gcc-15 changes, will cause this to be quite noisey until we get prod  up-to-date. And in this setting its merged, but not yet deployed. We probably want to catch this differently, rather than noticing its a scheduler bug or something.

tests/validate_forecast.py
```diff
@@ -0,0 +1,674 @@
...
+    """Read-only Maestro API client with pagination"""
+
+    def __init__(self, url, timeout=60, page_size=1000):
+        self.base = url.rstrip("/")
+        self.timeout = timeout
+        self.page_size = page_size
+        self.session = requests.Session()
+        adapter = requests.adapters.HTTPAdapter(
+            max_retries=requests.adapters.Retry(
+                total=3, backoff_factor=2, status_forcelist=[500, 502, 504]
```
bhcopeland: We should probably add the other status codes here, 503, and 429.

.github/workflows/validate-forecast.yml
```diff
@@ -0,0 +1,93 @@
...
+        # today's config would flag jobs added since then as MISSING.
+        env:
+          KCI_AGE_HOURS: ${{ github.event.inputs.age_hours || '24' }}
+          KCI_WINDOW_HOURS: ${{ github.event.inputs.window_hours || '24' }}
+        run: |
+          hours=$(python3 -c "print(int(float('${KCI_AGE_HOURS}') + float('${KCI_WINDOW_HOURS}')))")
+          ref=$(git rev-list -1 --before="${hours} hours ago" HEAD)
+          if [ -n "$ref" ]; then
+            echo "Using config from $ref ($(git log -1 --format='%ci %s' "$ref"))"
+            git checkout "$ref" -- config
```
bhcopeland, line 58: We probably want to add rm -rf config before this. Also up in line 3 we have python3 commands going on, but setup python is down on like 65. I would put setup python at the top before these steps. (or we use some sort of awk command here.

.github/workflows/validate-forecast.yml
```diff
@@ -0,0 +1,93 @@
...
+            echo "No commit older than ${hours}h; keeping HEAD config"
+          fi
+
+      - name: Check out kernelci-core (forecast tool)
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          # TODO: switch to kernelci/kernelci-core@main once the forecast
+          # graph support (evaluate_forecast_entry) lands upstream.
+          repository: nuclearcat/kernelci-core
+          ref: codex/forecast-missing-job-debug
```
bhcopeland: I think we should pin this to a commit sha, until it lands upstream, rathar than a outside repo

bhcopeland:
Thanks for putting this together @nuclearcat this is a good approach to catch this going forward. I wonder if we should add something around the  if: failure() step (like create github issue, or some sort of of discord message)?

tests/validate_forecast.py
nuclearcat: nipa is a bit weird job btw, it must be deprecated as seems netdev changed things since long time

nuclearcat:
> Thanks for putting this together @nuclearcat this is a good approach to catch this going forward. I wonder if we should add something around the if: failure() step (like create github issue, or some sort of of discord message)?

I am planning that after we make sure there is no unnecessary noise (as we already have too much from -status latency warnings, and it must be reduced).

PR #1542, opened 2026-07-09
platforms: use appended DTB for sm8350-hdk fastboot

nuclearcat:
Required for Collabora lab device


bhcopeland:
Do you want to add these bits to LAVA upstream too?

nuclearcat:
I will sync with Collabora guys, maybe already someone planned such efforts. If not - i will try to upstream

PR #1545, opened 2026-07-13
Make dragonboards run over dockerized fastboot

nuclearcat:


config/platforms.yaml
```diff
@@ -106,13 +106,17 @@ platforms:
     mach: qcom
     dtb: dtbs/qcom/apq8016-sbc.dtb
     compatible: ['qcom,apq8016-sbc', 'qcom,apq8016']
+    params:
+      fastboot_docker_image: registry.gitlab.collabora.com/lava/health-check-docker:main
```
bhcopeland, line 110: Is there a none :main tag? I.e. version locking it? Just a wee bit safer. Also, looking at image its on bullseye, not an issue but bullseye is EOL

nuclearcat: I didn't design that specific setup, so I adopted the current configuration to align with the recommendation to use health jobs container.

bhcopeland: Ack not a blocker from my side just a FYI, probably mainly the main is the bit I'd worry about (since it can easily change)

PR #1546, opened 2026-07-13
scheduler: Improve scheduler reliability

nuclearcat:
We update watchdog only after full event processing cycle, which might cause issues if event takes too long time.


nuclearcat:
Tested on staging

PR #1548, opened 2026-07-14
send_kcidb: don't report execution backends as labs

### LINKED ISSUE #1547: send_kcidb: don't report execution backends (shell/k8s) as labs in KCIDB misc fields
The KCIDB bridge reports the Maestro runtime name as the lab for every build and test. For LAVA jobs the runtime is a real lab (`lava-collabora`, `lava-broonie`, ...), but for builds and virtualised tests it is an execution backend (`shell`, `k8s-gke-eu-west1-...`), which is not a lab.

  So in `send_kcidb.py`:

  1. At init, record which runtime names have a lab-like `lab_type` (`lava`, `pull_labs`).
  2. For build nodes, only set `misc.lab` when `data.runtime` is one of those lab runtimes;
     omit it otherwise. `misc.runtime` can keep the raw executor name since it no longer
     claims to be a lab.
  3. For test nodes, do the same, and emit an explicit `misc.lab` instead of overloading
     `misc.runtime`, so builds and tests use one consistently named field.

A build or test with no lab then arrives lab-less in KCIDB, and consumers can apply their own fallback (the dashboard already coalesces a missing lab to the origin as a "virtual lab"), without needing to pattern-match executor names.

bhcopeland:
The bridge copies the Maestro runtime name into misc.lab (builds) and misc.runtime (tests) for every node. For LAVA jobs that is a real lab, but for builds and virtualised tests it is an execution backend such as shell or a k8s cluster name, forcing consumers like the dashboard to pattern-match and strip those values at ingest time.

Record at init which configured runtimes are actual labs (LAVA and pull labs) and only report those in misc.lab, for tests as well as builds. Nodes executed by shell, Kubernetes or Docker runtimes are submitted without a lab so consumers can apply their own fallback. misc.runtime keeps the raw executor name.

Closes: https://github.com/kernelci/kernelci-pipeline/issues/1547

### DIFF
--- a/src/send_kcidb.py
+++ b/src/send_kcidb.py
@@ -24,7 +24,7 @@ import kcidb
 import kernelci
 import kernelci.config
 import requests
-from kernelci.config.runtime import RuntimeLAVA
+from kernelci.config.runtime import RuntimeLAVA, RuntimePullLabs
 from kernelci.legacy.cli import Args, Command, parse_opts
 
 from base import SERVICE_PIPELINE, Service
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
Thanks for catching this

PR #1551, opened 2026-07-15
config: runtime: ltp-pull-labs: carry the kirk path to tuxrun

bhcopeland:
The base pull_labs template renders tuxrun_parameters into the job definition's top-level parameters.


bhcopeland:
https://github.com/kernelci/kernelci-core/pull/3148

PR #1552, opened 2026-07-15
Qcom kernelci lab update boot method , add new platform and kselftests pipeline

yogeshlal:
migrate eight boards from fastboot to EFI
add hamoa-evk (x1e80100) and purwa-evk (x1e001x0) EFI platforms
Enable new test pipelines


nuclearcat (approved):
Thank you Yogesh, looks great and seems will resolve issue with validator.

PR #1557, opened 2026-07-20
config: refine kcidebug builds

nuclearcat:
This is an initial step toward providing boot coverage with extended kernel debugging enabled.

Split common diagnostics from architecture-specific KASAN settings, use lower-overhead software tag-based KASAN on arm64, and correct the x86 job to build x86_64. Limit these debug builds to kernelci, stable, mainline, and next trees.


nuclearcat:
tested on staging

PR #1561, opened 2026-07-22
tools: Add example LAVA pull lab script

nuclearcat:
Labs running LAVA behind a firewall cannot use the push-style LAVA integration as the pipeline has no URL to reach them. When a lava runtime is configured without a url (e.g. lava-testpull), the pipeline stores the rendered LAVA job definition in external storage and publishes its URL as the job_definition node artifact instead.

Add tools/example_pull_lab_lava.py which polls the events API for such jobs, filtered by a mandatory lab name and optional platforms and device types, and submits the definitions verbatim to a local LAVA instance via its REST API. Results are reported by LAVA itself through the notify callback embedded in the definition, so the lab only needs outbound connectivity.

Also document the LAVA pull lab flow in doc/connecting-pull-lab.md.

Assisted-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/tools/example_pull_lab_lava.py
+++ b/tools/example_pull_lab_lava.py
@@ -1,0 +1,459 @@
+#!/usr/bin/env python3
+"""
+Poll KernelCI API for LAVA pull-lab jobs and run them on a local LAVA
+instance.
+
+This is a companion to example_pull_lab.py for labs that run LAVA behind
+a firewall.  Such labs cannot use the regular push-style LAVA integration
+because the KernelCI pipeline has no URL to reach them.  Instead, a LAVA
+runtime is configured in the pipeline *without* a `url` (see the
+`lava-testpull` entry in config/pipeline.yaml): the pipeline then renders
+the complete LAVA job definition, stores it in external storage and
+publishes its URL as the `job_definition` artifact of the job node.
+
+This script pulls those jobs:
+
+1. Poll the KernelCI events API for job nodes with a `job_definition`
+   artifact pointing at a LAVA YAML definition.
+2. Download the definition and submit it verbatim to the local LAVA
+   instance via its REST API.
+
+No result handling is needed here: the definition already contains a
+`notify` callback pointing back at the KernelCI pipeline, so LAVA posts
+its standard callback (outbound HTTP) when the job finishes, just like
+push-style LAVA labs do.  The callback token name in the definition
+(e.g. `kernelci-lab-testpull`) must exist as a "remote token" in the
+LAVA profile of the submitting user, holding the secret value shared
+with the KernelCI pipeline admins.
+
+Submitted jobs and the last seen event timestamp are persisted in an
+append-only JSONL journal (see --journal-file), so restarting the
+script neither resubmits jobs nor replays old events.  The journal
+tolerates interrupted writes and is compacted automatically.
+
+Examples:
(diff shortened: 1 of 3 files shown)


tools/example_pull_lab_lava.py
```diff
@@ -0,0 +1,336 @@
...
+            time.sleep(POLL_PERIOD)
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Run KernelCI LAVA pull-lab jobs on a local LAVA instance."
+    )
+    parser.add_argument(
+        "--api-url",
+        default="https://staging.kernelci.org:9000/latest",
```
patersonc, line 358: Do you intend to change this default to the production server before merging?
Or did you want to merge it first, test and update it later?

tools/example_pull_lab_lava.py
```diff
@@ -0,0 +1,336 @@
...
+        f"platform: {platform}, runtime: {runtime})"
+    )
+    processed.add(node_id)
+
+    definition = retrieve_job_definition(job_definition_url)
+    process_definition(definition, args)
+
+
+def poll_loop(args):
+    global timestamp
```
patersonc: What happens if a user runs the script once, quits out and later runs the script again?

`pollevents()` uses this timestamp as a "from" marker when looking for new jobs.

Do we want to initialise `timestamp` here to be the time that the script is run? Or store the "latest" used timestamp to a file?

tools/example_pull_lab_lava.py
```diff
@@ -0,0 +1,336 @@
...
+    if runtime != args.lab_name:
+        return
+    if args.platforms and platform not in args.platforms:
+        return
+
+    print(
+        f"Processing job {node_id} (name: {node.get('name')}, "
+        f"platform: {platform}, runtime: {runtime})"
+    )
+    processed.add(node_id)
```
patersonc: I guess `processed` will just keep getting larger and larger forever? (until the script is quit and re-run)

tools/example_pull_lab_lava.py
```diff
@@ -0,0 +1,336 @@
...
+
+    definition = retrieve_job_definition(job_definition_url)
+    process_definition(definition, args)
+
+
+def poll_loop(args):
+    global timestamp
+
+    processed = set()
```
patersonc: Should we be saving the "processed" jobs to a file for subsequent script runs to build upon?

tools/example_pull_lab_lava.py
nuclearcat: It is "safe default", so anybody who want to experiment with it, at beginning better hit staging than making problems at production.

tools/example_pull_lab_lava.py
nuclearcat: I will redesign it with persistence, for timestamp

tools/example_pull_lab_lava.py
nuclearcat: Good point, i will adjust it and add expiration and persistence for the processed nodes.

tools/example_pull_lab_lava.py
nuclearcat: Yes, will update it

patersonc:
FYI I've created https://github.com/kernelci/kernelci-pipeline/pull/1562 to use to test this PR with

PR #1562, opened 2026-07-22
config: Add lava-renesas pull lab

patersonc:
Add a new lava-renesas runtime using the PULL_LABS-compatible LAVA runtime (lab_type: lava with no url), so the pipeline renders and stores baseline-nfs job definitions for the lab to pull via tools/example_pull_lab_lava.py instead of pushing them directly.

Start with baseline-nfs boot testing on RZ/G2M and RZ/G2L

### DIFF
--- a/config/pipeline.yaml
+++ b/config/pipeline.yaml
@@ -164,5 +164,11 @@ runtimes:
       callback:
         token: kernelci-lab-testpull
 
+  lava-renesas:
+    lab_type: lava
+    notify:
+      callback:
+        token: kernelci-lab-renesas
+
   shell:
     lab_type: shell
--- a/config/platforms.yaml
+++ b/config/platforms.yaml
@@ -586,6 +586,12 @@ platforms:
 
   fvp-aemva: *fvp-device
 
+  r8a774a1-hihope-rzg2m-ex:
+    <<: *arm64-device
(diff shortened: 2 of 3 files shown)


patersonc:
Apart from the LAVA callback notifications, this PR is all working now, ready for review - thanks!

config/pipeline.yaml
```diff
@@ -164,5 +164,13 @@ runtimes:
       callback:
         token: kernelci-lab-testpull
 
+  lava-renesas:
+    lab_type: lava
```
patersonc: Hi @nuclearcat 

Were you saying that I should make this `pull_labs` rather than `lava`?

nuclearcat: No, i was confused first. We have also "pull_lava", which is same type, just no url defined to submit job (such lab pulls them by themself). I will generate token now.

patersonc:
Callbacks are now working. Tests jobs passing:

r9a07g044l2-smarc-rzg2l: https://staging.kernelci.org:9000/viewer?node_id=6ab59c37f751e4a5434ced0e
r8a774a1-hihope-rzg2m-ex: https://staging.kernelci.org:9000/viewer?node_id=6ab59c36f751e4a5434ced0b

PR #1563, opened 2026-07-24
Restore patchset function and add patch upload functionality

nuclearcat:
This series restore patchset, patch bugs existed before, adapt to new changes (durable events, etc), add functionality to upload patches to storage.

### DIFF
--- a/src/patchset.py
+++ b/src/patchset.py
@@ -1,0 +1,334 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# Author: Nikolay Yurin <yurinnick@meta.com>
+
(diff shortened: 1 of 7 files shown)


src/patchset.py
```diff
@@ -0,0 +1,502 @@
...
+        if len(patch_data) > max_size:
+            raise PatchValidationError(
+                f"Patch {patch_name} exceeds maximum size of {max_size} bytes"
+            )
+        if b"\x00" in patch_data:
+            raise PatchValidationError(
+                f"Patch {patch_name} contains binary data"
+            )
+
+        lines = patch_data.split(b"\n")
```
bhcopeland: lines = [ln.rstrip(b"\r") for ln in patch_data.split(b"\n")]

I'd normalise this to catch these use cases

nuclearcat: Fixed

PR #1565, opened 2026-07-24
RFC triage-agent - automated RCA for LAVA job failures

yogeshlal:
[`triage-agent`](https://github.com/yogeshlal/triage-agent) is a small FastAPI service that triage LAVA job. POST /{job_id} submits a run, GET /triage/{run_id} polls for the report. It fetches the LAVA log, drives the skills/kernel-triage skill through a LiteLLM tool-calling loop (Anthropic / OpenAI / Gemini / Ollama  any LiteLLM provider), and writes a Markdown RCA with a filled-in applicability-gate table and a root-cause hypothesis.

It plugs into kernelci-pipeline via the new TriageNotifier: when a LAVA job comes back fail or incomplete, lava_callback fires TriageNotifier.maybe_trigger, which POSTs to this service and polls until the report is ready. Optional GitHub issue and email delivery live on the pipeline side.

### DIFF
--- a/config/jobs.yaml
+++ b/config/jobs.yaml
@@ -182,6 +182,7 @@ jobs:
   baseline-arm64-pengutronix: *baseline-job
   baseline-arm64-foundriesio: *baseline-job
   baseline-arm64-qualcomm: *baseline-job
+  baseline-arm64-preempt_rt-qualcomm: *baseline-job
   baseline-riscv-broonie: *baseline-job
   baseline-riscv-clabbe: *baseline-job
   baseline-riscv-collabora: *baseline-job
@@ -1908,12 +1909,26 @@ jobs:
         - collabora-next:for-kernelci
     kcidb_test_suite: kselftest.devices-exist
 
+  kselftest-devices-exist-ramdisk:
+    <<: *kselftest-job-ramdisk
+    params:
+      <<: *kselftest-params-ramdisk
+      collections: devices-exist
+    kcidb_test_suite: kselftest.devices-exist
+
   kselftest-device-error-logs-main:
     <<: *kselftest-job
     params:
       <<: *kselftest-params
       collections: devices/error_logs
     kcidb_test_suite: kselftest.device_error_logs
+
+  kselftest-device-error-logs-ramdisk:
+    <<: *kselftest-job-ramdisk
+    params:
+      <<: *kselftest-params-ramdisk
+      collections: device-error-logs
+    kcidb_test_suite: kselftest.device-error-logs
 
   kselftest-devices-probe:
     <<: *kselftest-job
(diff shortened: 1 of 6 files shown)


bhcopeland:
Really like the direction as an RFC, my concern is implementation rather than the idea.

kci-dev already provides this layer: the KernelCIClient library (#[277](https://github.com/kernelci/kci-dev/pull/277)) and the MCP server built on it and they're the sanctioned way to talk to KernelCI, fetching nodes/tests/logs and driving actions like retry/checkout (although the latter does need some more testing). This PR adds a second path beside it, the TriageNotifier glue in lava_callback plus the agent doing its own KernelCI/LAVA access, and I'd rather we didn't maintain two systems for the same thing.

I'd push on is having it consume KernelCIClient (and the MCP tools) for its data and any follow-up actions, instead of reimplementing fetch/auth and wiring a bespoke notifier into the pipeline. Both the trigger-on-failure and the act-on-it steps fit the MCP model well, so it should integrate more cleanly there.

The MCP can drive fetching nodes/tests/logs and driving actions like retry/checkout (although the latter I am testing, so it is experimental).

yogeshlal:
Thanks @bhcopeland  for review, my vision was something like 

submit job -> job executed (LAVA) -> Test failed -> lava_callback  ->triage-agent -> share RCA

And for that lava_callback was right point where I can link https://github.com/yogeshlal/triage-agent to triage the failure job and generate RCA for failed job and using that create issue/bug.

bhcopeland:
> Thanks @bhcopeland for review, my vision was something like
> 
> submit job -> job executed (LAVA) -> Test failed -> lava_callback ->triage-agent -> share RCA
> 
> And for that lava_callback was right point where I can link https://github.com/yogeshlal/triage-agent to triage the failure job and generate RCA for failed job and using that create issue/bug.

Understood, that LGTM (i.e. thats the right place for that but I am not questioning that bit).

My concern is for the one layer down. How the agent reaches KCI. 
 
- Issue matching: We have that in kci-dev list_issues/get_issue. I'd probably use this system here.
- Context: pulling the node/test/build around the failure via KernelCIClient, so the RCA has the KernelCI-side picture and not just the LAVA log.
- Follow-up actions: filing/linking the bug, or retrying the job, are things the kci-dev/MCP action tools already cover.

So keep lava_callback as the trigger and the LAVA-log fetch as-is. I just want the KernelCI interaction and the act-on-it side to go through kci-dev so we're not standing up a second integration path beside it. Bonus is the MCP setup then gets the same capability for free, which is where I think this gets genuinely powerful.

yogeshlal:
@bhcopeland got your perspective , let me revisit the trigger point  but I think we are clear on having triage agent like https://github.com/yogeshlal/triage-agent for log analysis and linking to know issues ?

bhcopeland:
@yogeshlal I put together a PR: https://github.com/kernelci/kci-dev/pull/306 in regards to what we were speaking about last week in meeting. 

The bit thats probably missing needs discussing is issue matching. We have get_test_issues/get_build_issues/list_issues, but it might not be fully what you are after in the triage model. You need to have a play with it.

yogeshlal:
@bhcopeland , let me try it

PR #1587, opened 2026-08-26
config: pull-labs: add a UnixBench benchmark job for aws-ec2

bhcopeland:
Add the job definition and jinja2 template for running UnixBench on the pull-labs aws-ec2 runtime.

The scheduler entry is included but commented out. The runner cannot honour an A/B benchmark yet: a job definition carries a single kernel and the benchmark needs two to compare against each other. With no scheduler entry the job is never dispatched, so this lands the config shape for review without changing what runs today.


bhcopeland:
@nuclearcat rebased. Can you give a once over so we can merge.

@nmanthey any blockers to not merge this?

nmanthey (approved):

PR #1593, opened 2026-09-03
validate_yaml: Diagnostic improvements

broonie:
A couple of fairly trivial usability improvements for validate_yaml.py.

- **validate_yaml: Handle and print exceptions**
- **validate_yaml: Also handle ValueError from recursive_merge()**

### DIFF
--- a/tests/validate_yaml.py
+++ b/tests/validate_yaml.py
@@ -394,11 +394,15 @@ def main():
         "-o", "--output", type=str, help="Output file to dump yaml files"
     )
     args = parser.parse_args()
-    merged_data = merge_files(args.dir)
-    if args.output:
-        dumper(args.output, merged_data)
+    try:
+        merged_data = merge_files(args.dir)
+        if args.output:
+            dumper(args.output, merged_data)
 
-    validate_yaml(merged_data)
+        validate_yaml(merged_data)
+    except (yaml.YAMLError, ValueError, OSError) as exc:
+        print(f"Validation failed: {exc}")
+        sys.exit(1)
 
 
 if __name__ == "__main__":
--- a/tests/validate_yaml.py
+++ b/tests/validate_yaml.py
@@ -338,7 +338,7 @@ def merge_files(dir="config"):
                 data = yaml.safe_load(stream)
                 validate_duplicate_jobs(data, file)
                 merged_data = recursive_merge(merged_data, data)
-            except yaml.YAMLError as exc:
+            except (yaml.YAMLError, ValueError) as exc:
                 print(f"Error in {file}: {exc}")
                 sys.exit(1)
     return merged_data


tests/validate_yaml.py
```diff
@@ -393,12 +393,16 @@ def main():
     parser.add_argument(
         "-o", "--output", type=str, help="Output file to dump yaml files"
     )
-    args = parser.parse_args()
-    merged_data = merge_files(args.dir)
-    if args.output:
-        dumper(args.output, merged_data)
-
-    validate_yaml(merged_data)
+    try:
```
bhcopeland, line 397: Nice, this is a good improvement, YAML validation is painful. LGTM. A few small things.

tests/validate_yaml.py
```diff
@@ -393,12 +393,16 @@ def main():
...
+        args = parser.parse_args()
```
bhcopeland: args = parser...() (should be above the try block).. Args will raise a SystemExit anyway, and we probably don't want it in the try block

tests/validate_yaml.py
```diff
@@ -393,12 +393,16 @@ def main():
...
+        merged_data = merge_files(args.dir)
+        if args.output:
+            dumper(args.output, merged_data)
+
```
bhcopeland, line 401: Can we drop the new line?

tests/validate_yaml.py
```diff
@@ -393,12 +393,16 @@ def main():
...
+        validate_yaml(merged_data)
+    except Exception as exc:
```
bhcopeland: I think we want to catch yaml.YAMLError here instead of Exception. As Exception will probably swallow up tracebacks for bugs.

broonie: We're raising other exception types (see the second commit about ValueError).

tests/validate_yaml.py
broonie: That was in the original code that's being refactored?

tests/validate_yaml.py
bhcopeland: Thanks for the reindent fix. Indeed, when testing locally I hit https://github.com/kernelci/kernelci-pipeline/blob/f73aea0a245e31310ef5d011c288e8dbf5b5c626/tests/validate_yaml.py#L341-L343.

Since this already calls sys.exit(1), the ValueError never reaches the handler at the bottom (SystemExit isn't an Exception). So we either need to fix that up, or change this to YAMLError. I'd lean towards the latter. The good thing about Exception here is it will catch accidental stuff, but I feel what it will do in the process is swallow up bugs.

tests/validate_yaml.py
bhcopeland: Yes you're right. Ignore that comment.

tests/validate_yaml.py
broonie: If you chmod -r the YAML files you can also get something like "[Errno 13] Permission denied: 'configs/jobs.yaml'" printed, that's a lot nicer than the backtrace that would otherwise result. The things we raise in the code itself are YAMLError and ValueError exceptions but we're also calling into other code which reports errors as exceptions which has similar potential issues with generating hard to read splats and not formatting the error messages well.

Given that we log every exception and report a failure exit code I'm not really concerned about errors getting lost, the worst case is that it might be hard to figure out where an exception came from and you'd need to do something like delete the handler to figure it out, but that's also true of the errors we raise ourself if we do a bad job of describing the error when we raise within the validation code.

fboudra: I'll tend to agree with @bhcopeland, we should narrow the exception handling in main(). Something like:
```diff
--- a/tests/validate_yaml.py
+++ b/tests/validate_yaml.py
@@ -400,8 +400,8 @@ def main():
             dumper(args.output, merged_data)
 
         validate_yaml(merged_data)
-    except Exception as exc:
-        print(f"Validation failed: {exc}")
+    except (yaml.YAMLError, ValueError, OSError) as exc:
+        print(f"Error: {exc}")
         sys.exit(1)
 
```

Catch only the exceptions the merge, dump and validation paths can actually raise (YAMLError, ValueError, OSError) instead of Exception so genuine code bugs are reported with a full traceback rather than being masked as a validation failure. Also, rename the message to 'Error' as the handler also covers
dumper failures.

The exception types actually reachable through merge_files/dumper/validate_yaml are enumerable: YAMLError, ValueError, OSError. If a new type ever surfaces in practice, adding it to the tuple is a one-token fix triggered by a real incident.

broonie: On the one hand it's a one token change, on the other hand I've much more often had to add a global exception handler to a Python program to make an error message intelligible than I've found a backtrace constructive. If you're actively developing the program they can help, but as a user it's too much information.

PR #1600, opened 2026-09-13
config: add LTP and kselftest jobs for sasha-lab

sashalevin:
sasha-lab currently runs a baseline boot only. Add LTP and kselftest jobs on the same three platforms, so the hardware reports something more than "it booted".


config/jobs-pull-labs.yaml
```diff
@@ -267,3 +267,143 @@ jobs:
...
         - next
+
+# Test suites for sasha-lab. Separate entries rather than reuse of the
+# ltp-*-pull-labs / kselftest-*-pull-labs jobs because this lab only tests
+# mainline and next, and those jobs carry no tree rule.
+#
+# The lab boots the machine's own installed rootfs, so it runs each suite in a
+# chroot of the same KernelCI rootfs the NFS-booting labs use. `nfsroot` is
+# therefore still required: the template turns it into the `rootfs` artifact,
+# which is the tarball the chroot is built from.
```
bhcopeland: Hey Sasha, thanks for the PR. Could we trim the comments to those explaining non-obvious requirements or constraints? The explanation of why nfsroot is needed seems useful, but some of the surrounding commentary could be shortened.

config/jobs-pull-labs.yaml
```diff
@@ -267,3 +267,143 @@ jobs:
       tree:
         - mainline
         - next
+
+# Test suites for sasha-lab. Separate entries rather than reuse of the
+# ltp-*-pull-labs / kselftest-*-pull-labs jobs because this lab only tests
+# mainline and next, and those jobs carry no tree rule.
```
bhcopeland: This isn't quite true; the full LTP jobs already inherit a tree restriction allowing mainline, next and stable from ltp-job-pull-labs. LTP smoke and the existing kselftest jobs have no tree restriction, so this comment is too broad. We can reuse the shared jobs and exclude stable through your runtime rules

config/jobs-pull-labs.yaml
```diff
@@ -267,3 +267,143 @@ jobs:
...
+  ltp-smoke-sasha-lab:
+    template: ltp-pull-labs.jinja2
+    kind: job
+    priority: low
+    kcidb_test_suite: ltp
+    params:
+      <<: *ltp-params-pull-labs
+      tst_cmdfiles: "smoketest"
+      job_timeout: 30
+    rules:
```
bhcopeland: Could we put this restriction once on the existing sasha-lab runtime in config/pipeline-pull-labs.yaml?

```
   sasha-lab:
     lab_type: pull_labs
     poll_interval: 60
     timeout: 7200
     rules:
       tree:
         - mainline
         - next
  
```
Runtime rules are enforced alongside job rules, so this would keep Sasha restricted to mainline and next without repeating the restriction on every job.

config/jobs-pull-labs.yaml
```diff
@@ -267,3 +267,143 @@ jobs:
...
+      tree:
+        - mainline
+        - next
+
+  kselftest-cgroup-sasha-lab:
```
bhcopeland: Could we name these kselftest-cgroup-pull-labs, etc., so other pull labs can reuse the definitions by adding their own scheduler entries? Sasha’s tree restriction would live on its runtime, and its platform selection would stay in the scheduler. That should avoid duplicating each collection’s definition for every lab.

config/jobs-pull-labs.yaml
```diff
@@ -267,3 +267,143 @@ jobs:
...
+    rules:
+      tree:
+        - mainline
+        - next
+
+# Server-only suites. These run on the large machines and deliberately not on
+# the Jetson: until several jobs can share one boot, every job there costs a
+# full ~50 min reflash, and the four suites above already fill that budget.
+
+  ltp-syscalls-sasha-lab:
```
bhcopeland: Same with this: with the mainline/next restriction on the sasha-lab runtime we can reuse `ltp-syscalls-pull-labs` and `ltp-smoke-pull-labs` and drop both. `kselftest-arm64-sasha-lab` too, since `kselftest-arm64-pull-labs` already exists, so if you want to change timeouts etc I think we should really only set it once.

If the default timeouts don't work for you, I say we adjust later, but keep them where they are for now. FWIW we have been using these defaults in tuxlava for a while: [LTP](https://github.com/kernelci/tuxlava/blob/291aac19926972e1470ab8b1d15f54b486f5ac69/tuxlava/tests/ltp.py) and [kselftest](https://github.com/kernelci/tuxlava/blob/291aac19926972e1470ab8b1d15f54b486f5ac69/tuxlava/tests/kselftest.py)

config/jobs-pull-labs.yaml
sashalevin: Thanks for the review!!

Ack

config/jobs-pull-labs.yaml
sashalevin: Ack

config/jobs-pull-labs.yaml
sashalevin: Ack

